PR #60, opened 2019-10-31
LevelDb integration

oaleksandrov-ip:
1. Integrate storage from ErgoPlatform as ErgoPlatformStorageAdapter
2. Performance test for storage creation


sdk/pom.xml
```diff
@@ -17,6 +17,12 @@
 			<name>Scala-Tools Maven2 Repository</name>
 			<url>http://scala-tools.org/repo-releases</url>
 		</repository>
+
+		<repository>
+			<id>spring-plugin-repo</id>
```
i-Alex: Do we need to have this dependency?

sdk/pom.xml
```diff
@@ -136,6 +142,33 @@
...
+		</dependency>
+
+		<!-- https://mvnrepository.com/artifact/org.ethereum/leveldbjni-all -->
+		<dependency>
+			<groupId>org.ethereum</groupId>
+			<artifactId>leveldbjni-all</artifactId>
+			<version>1.18.3</version>
+		</dependency>
+
+		<!--dependency>
```
i-Alex: can be removed

sdk/src/main/scala/com/horizen/storage/ErgoPlatformStorageAdapter.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
+
+
+/*
+*  @TODO to discuss
+*    1. Why we use ByteArrayWrapper instead of Array[Byte]?
+*    2. We need iterator over the storage
+* */
+class ErgoPlatformStorageAdapter(pathToDB: String, keepVersions: Int) extends Storage{
```
i-Alex: Why not to replace "ErgoPlatform..." to "VersionedLevelDB..." or something like that

sdk/src/main/scala/org/ergoplatform/db/ChangeSet.scala
```diff
@@ -0,0 +1,68 @@
+package org.ergoplatform.db
```
i-Alex: change to our package structure?

sdk/src/test/scala/com/horizen/storage/performance/StoragePerformanceTest.scala
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+  val runner = new StoragePerformanceRunner(performanceConfig)
+
+  def printHeader(): Unit = {
+    TimeMeasure.header.foreach(column => print(s"$column\t"))
+    println()
+  }
+
+  @Test
```
i-Alex: This test consumes a lot of time and need to be executed only if we want to compare different storage implementations. Maybe we can comment `@Test` to avoid running it each time we run full test suit.

sdk/src/main/scala/com/horizen/storage/ErgoPlatformStorageAdapter.scala
```diff
@@ -0,0 +1,72 @@
...
+import javafx.util.{Pair => JPair}
+import org.ergoplatform.db.LDBFactory.factory
+import org.ergoplatform.db.VersionedLDBKVStore
+import org.iq80.leveldb.Options
+
+import scala.collection.JavaConverters._
+import scala.compat.java8.OptionConverters._
+
+
+/*
```
i-Alex: LevelDB storage related interfaces are based on top of Array[Byte] instead of ByteArrayWrapper in IODB Store.
If we are going to remove IODB totally with Level DB implementation, we can think about reviewing of `Storage` interface:
1) Do we need to keep `ByteArrayWrapper` usage here?
2) review `Storage` methods in general.

sdk/src/main/scala/com/horizen/storage/leveldb/VersionedLevelDbAdapterToStorageAdapter.scala
```diff
@@ -18,7 +18,7 @@ import scala.compat.java8.OptionConverters._
 *    1. Why we use ByteArrayWrapper instead of Array[Byte]?
 *    2. We need iterator over the storage
 * */
-class ErgoPlatformStorageAdapter(pathToDB: String, keepVersions: Int) extends Storage{
+class VersionedLevelDbAdapterToStorageAdapter(pathToDB: String, keepVersions: Int) extends Storage{
```
i-Alex: Why not just VersionedLevelDbStorageAdapter?

sdk/pom.xml
```diff
@@ -143,6 +143,14 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- https://mvnrepository.com/artifact/javax.xml.bind/jaxb-api -->
```
i-Alex: Note: should be automatically removed after merging with changes of https://github.com/ZencashOfficial/Sidechains-SDK/pull/68

PR #67, opened 2019-11-08
MCPublicKeyHash, WithdrawalRequestBox initial implementation.

SergeyMakukha:
Changes in RegularTransaction to support WithdrawalRequestBoxes.


sdk/src/main/java/com/horizen/transaction/RegularTransactionCreator.java
```diff
@@ -14,26 +15,37 @@
...
+
+        long toAmount = 0;
         for(Pair<PublicKey25519Proposition, Long> pair : to) {
             if (pair.getValue() < 0)
                 throw new IllegalArgumentException("Output values must be >= 0.");
-            to_amount += pair.getValue();
+            toAmount += pair.getValue();
+        }
+
+        long withdrawalRequestsAmount = 0;
```
i-Alex: why not to use the single aggregator `toAmount`? In general box regular and withdrawal outputs have the same meaning in context of coins

sdk/src/main/scala/com/horizen/api/http/SidechainTransactionApiRoute.scala
```diff
@@ -288,12 +301,14 @@ case class SidechainTransactionApiRoute(override val settings: RESTApiSettings,
   def sendCoinsToAddress: Route = (post & path("sendCoinsToAddress")) {
```
i-Alex: As I understood, it is a method for simple sending coins inside sidechain.
Let's leave it as is (without withdrawals, only regular box outputs) and create another method `withdrawcoins` with withdrawals only.

sdk/src/main/scala/com/horizen/api/http/SidechainTransactionApiRoute.scala
```diff
@@ -199,6 +199,12 @@ case class SidechainTransactionApiRoute(override val settings: RESTApiSettings,
               new lang.Long(element.value))
           ).toIndexedSeq
 
+          val withdrawalRequests: IndexedSeq[Pair[MCPublicKeyHash, lang.Long]] = body.withdrawalRequests.map(element =>
```
i-Alex: please, update also swagger yaml file with the recent API changes.

sdk/src/test/java/com/horizen/box/WithdrawalRequestBoxSerializerTest.java
```diff
@@ -0,0 +1,75 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class WithdrawalRequestBoxSerializerTest
+{
+    WithdrawalRequestBox box;
+
+    @Before
+    public void setUp() {
+        Tuple2<byte[], byte[]> keyPair = Curve25519.createKeyPair("12345".getBytes());
```
i-Alex: can be removed

sdk/src/test/java/com/horizen/box/WithdrawalRequestBoxSerializerTest.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        */
+    }
+
+    @Test
+    public void WithdrawalRequestBoxSerializerTest_SerializationTest() {
+        BoxSerializer<WithdrawalRequestBox> serializer = box.serializer();
+        byte[] bytes = serializer.toBytes(box);
+
+        WithdrawalRequestBox box2 = serializer.parseBytesTry(bytes).get();
```
i-Alex: `serializer.parseBytesTry(bytes).get()` -> `serializer.parseBytes(bytes)`

i-Alex: same in other places

sdk/src/test/java/com/horizen/box/WithdrawalRequestBoxTest.java
```diff
@@ -0,0 +1,72 @@
...
+public class WithdrawalRequestBoxTest
+{
+    MCPublicKeyHash proposition;
+    long nonce;
+    long value;
+
+    @Before
+    public void setUp() {
+        byte[] anotherSeed = "testseed".getBytes();
+        Tuple2<byte[], byte[]> keyPair = Curve25519.createKeyPair(anotherSeed);
```
i-Alex: can be removed

sdk/src/test/java/com/horizen/transaction/RegularTransactionCreatorTest.java
```diff
@@ -159,9 +161,11 @@ public void beforeEachTest() {
     @Test
     public void RegularTransactionCreator_SuccessCreationTest() {
         List<Pair<PublicKey25519Proposition, Long>> to = new ArrayList<>();
+        List<Pair<MCPublicKeyHash, Long>> withdrawalRequests = new ArrayList<>();
```
i-Alex: We have tests for non-empty regular outputs and empty withdrawal. Please, add tests for both specified and only withdrawals specified

sdk/src/test/java/com/horizen/transaction/RegularTransactionSerializerTest.java
```diff
@@ -41,18 +43,22 @@ public void beforeEachTest() {
         to.add(new Pair<>(pk5.publicImage(), 20L));
         to.add(new Pair<>(pk6.publicImage(), 90L));
 
+        ArrayList<Pair<MCPublicKeyHash, Long>> withdrawalRequests = new ArrayList<>();
```
i-Alex: Also add tests similar as for comment above

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -37,6 +38,8 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage, over
   },
   s"Specified version is invalid. ${stateStorage.lastVersionId.map(w => bytesToVersion(w.data)).getOrElse(version)} != ${version}")
 
+  private var withdrawalEpochInfo : WithdrawalEpochInfo = WithdrawalEpochInfo(0, 0)
```
i-Alex: Each time we start the Node it will lead to setting `withdrawalEpochInfo` to initial value.
State should not keep `withdrawalEpochInfo` itself. Previous value should be retrieved from historyStorage.

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -116,6 +119,7 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage, over
   override def applyModifier(mod: SidechainBlock): Try[SidechainState] = {
     validate(mod).flatMap { _ =>
       changes(mod).flatMap(cs => {
+        this.withdrawalEpochInfo = WithdrawalEpochUtils.getWithdrawalEpochInfo(mod, this.withdrawalEpochInfo, this.params)
```
i-Alex: To avoid this operation here, BoxMinimalState trait can be reviewed. Even more, it seems that we can change it significantly or even remove.

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -44,30 +53,56 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
...
+      withdrawalRequestList.addAll(boxUpdateList
+        .filter(box => box.isInstanceOf[WithdrawalRequestBox])
+        .map(_.asInstanceOf[WithdrawalRequestBox])
+        .asJavaCollection)
+
+
+      updateList.add(new JPair(getWithdrawalRequestsKey(withdrawalEpochInfo),
+        new ByteArrayWrapper(withdrawalRequestSerializer.toBytes(withdrawalRequestList))))
+
+      updateList.add(new JPair(getWithdrawalRequestsKey(withdrawalEpochInfo),
```
i-Alex: key should be different, use `epochInformationKey`

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -44,30 +53,56 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
...
 
     for (r <- boxIdsRemoveList)
       removeList.add(calculateKey(r))
 
-    for (b <- boxUpdateList)
+    for (b <- boxUpdateList.filter(box => !box.isInstanceOf[WithdrawalRequestBox]))
       updateList.add(new JPair[ByteArrayWrapper, ByteArrayWrapper](calculateKey(b.id()),
         new ByteArrayWrapper(sidechainBoxesCompanion.toBytes(b))))
 
+    if (boxUpdateList.count(box => box.isInstanceOf[WithdrawalRequestBox]) > 0) {
```
i-Alex: I don't think that storage should do the withdrawal requests special calculation logic.
State itself should calculate box changes -> extract requests -> get from storage actual epoch info and requests list -> calculate boxes to add (remove requests), boxes to remove, new epoch info value, withdrawals to save for calculated epoch info.

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -53,38 +53,51 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
     }
   }
 
+  def getWithdrawalEpochInfo() : Option[WithdrawalEpochInfo] = {
+    storage.get(epochInformationKey) match {
+      case v if v.isPresent => WithdrawalEpochInfoSerializer.parseBytesTry(v.get().data).toOption
```
i-Alex: Add logs in case of parsing error like in `SidechainStateStorage.get`

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -53,38 +53,51 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
     }
   }
 
+  def getWithdrawalEpochInfo() : Option[WithdrawalEpochInfo] = {
```
i-Alex: `SidechainStateStorage.get` method name should be reviewed. Previously we had the only method for retrieving boxes, now we also retrieve epoch info.
What do you think about renaming `SidechainStateStorage.get` to `SidechainStateStorage.getBox`

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -53,38 +53,51 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
...
+    storage.get(epochInformationKey) match {
+      case v if v.isPresent => WithdrawalEpochInfoSerializer.parseBytesTry(v.get().data).toOption
+      case _ => None
+    }
+  }
+
+  def getWithdrawalRequests(withdrawalEpochInfo: WithdrawalEpochInfo) : JList[WithdrawalRequestBox] = {
+    storage.get(getWithdrawalRequestsKey(withdrawalEpochInfo)) match {
+      case v if v.isPresent => withdrawalRequestSerializer.parseBytes(v.get().data)
```
i-Alex: same as above

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -53,38 +53,51 @@ class SidechainStateStorage (storage : Storage, sidechainBoxesCompanion: Sidecha
...
-        .map(_.asInstanceOf[WithdrawalRequestBox])
-        .asJavaCollection)
-
+      withdrawalRequestList.addAll(withdrawalRequestAppendList.asJavaCollection)
 
       updateList.add(new JPair(getWithdrawalRequestsKey(withdrawalEpochInfo),
         new ByteArrayWrapper(withdrawalRequestSerializer.toBytes(withdrawalRequestList))))
 
-      updateList.add(new JPair(getWithdrawalRequestsKey(withdrawalEpochInfo),
+      updateList.add(new JPair(epochInformationKey,
```
i-Alex: This data should be updated in any case, no matter if withdrawalRequestAppendList is empty or not

sdk/src/test/java/com/horizen/transaction/RegularTransactionSerializerTest.java
```diff
@@ -91,6 +98,11 @@ public void RegularTransactionSerializerTest_RegressionTest() {
         assertEquals("Transaction serialization failed.", true, t.isSuccess());
 
         RegularTransaction parsedTransaction = t.get();
+        byte[] m1 = transaction.messageToSign();
```
i-Alex: should be removed?

PR #68, opened 2019-11-08
Jaxb api dependency removed.

g666:


sdk/src/main/java/com/horizen/utils/ByteArrayWrapper.java
```diff
@@ -13,4 +13,23 @@ public ByteArrayWrapper(Integer size) {
...
         this(wrapper.data());
     }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "[" + (
+                size() == 8 ? String.valueOf(getLong(data(), 0)) + "L" : BytesUtils.toHexString(data())) + "]";
+    }
+
+    private long getLong(byte[] buf, int pos) {
```
i-Alex: instead of defining getLong use BytesUtils.getLong

i-Alex:
As expected, without current changes we have a fail if `ByteArrayWrypper.ToString()` is called.
jaxb_error_1

PR #80, opened 2020-01-16
VRF Rust library initial implementation.

SergeyMakukha:


sdk/src/main/java/com/horizen/vrf/VRFKeyGenerator.java
```diff
@@ -0,0 +1,10 @@
+package com.horizen.vrf;
+
+import com.horizen.utils.Pair;
+
+public class VRFKeyGenerator {
+
+  public static Pair<VRFSecretKey, VRFPublicKey> generate(byte[] seed) {
+    return new Pair(new VRFSecretKey(new byte[32]) , new VRFPublicKey(new byte[32]));
```
i-Alex: Maybe, we need to do a jni call here that will return 2 byte arrays one for public another for private key

sdk/src/test/java/com/horizen/vrf/VRFProofTest.java
```diff
@@ -0,0 +1,45 @@
+package com.horizen.vrf;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class VRFProofTest {
+
+    static byte[] bhash = new byte[VRFProof.PROOF_LENGTH];;
```
i-Alex: `;;`

sdk/src/test/java/com/horizen/vrf/VRFProofTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @BeforeClass
+    public static void before() {
+        System.loadLibrary("vrflib");
+        for (byte i = 0; i < bhash.length; i++) bhash[i] = i;
```
i-Alex: Let me suggest to store `bhash` and `br` values as some predefined constants (can be even in hex form), so, later when we will have a real rust implementation we'll need just to change the these constants.

sdk/src/test/java/com/horizen/vrf/VRFPublicKeyTest.java
```diff
@@ -0,0 +1,45 @@
...
+import static org.junit.Assert.*;
+
+public class VRFPublicKeyTest {
+
+    static byte[] bkey = new byte[VRFPublicKey.KEY_LENGTH];
+
+    @BeforeClass
+    public static void before() {
+        System.loadLibrary("vrflib");
+        for (byte i = 0; i < bkey.length; i++) bkey[i] = i;
```
i-Alex: the same as for VRFProofTest

sdk/src/test/java/com/horizen/vrf/VRFSecretKeyTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.junit.Assert.*;
+
+public class VRFSecretKeyTest {
+
+    static byte[] bkey = new byte[VRFSecretKey.KEY_LENGTH];
+
+    @BeforeClass
+    public static void before() {
+        System.loadLibrary("vrflib");
+        for (byte i = 0; i < bkey.length; i++) bkey[i] = i;
```
i-Alex: the same as for VRFProofTest

PR #96, opened 2020-04-15
Ourboros vrf all nonce

oaleksandrov-ip:


sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +57,50 @@ trait ConsensusDataProvider {
...
+    }
+  }
 
-    require(nonceOpt.isDefined, "No mainchain reference had been found for whole consensus epoch") //crash whole world here?
-
-    NonceConsensusEpochInfo(bigIntToConsensusNonce(nonceOpt.get))
+  private def calculateNonceForNonGenesisEpoch(lastBlockIdInEpoch: ModifierId, lastBlockInfoInEpoch: SidechainBlockInfo): NonceConsensusEpochInfo = {
+    val previousEpoch = getPreviousConsensusEpochIdForBlock(lastBlockIdInEpoch, storage.blockInfoById(lastBlockIdInEpoch))
+
+    val allVrfOutputsWithSlots =
```
i-Alex, line 76: add type, please.

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +57,50 @@ trait ConsensusDataProvider {
...
+      foldEpochRight[ListBuffer[(VrfProof, VrfProofHash, ConsensusSlotNumber)]](ListBuffer(), lastBlockIdInEpoch, lastBlockInfoInEpoch) {
+      (blockId, blockInfo, accumulator) =>
+        (blockInfo.vrfProof, blockInfo.vrfProofHash, timeStampToSlotNumber(blockInfo.timestamp)) +=: accumulator
+    }.toList
+
+    //Hash function is applied to the concatenation of VRF values that are inserted into each block, using values from
+    //all blocks up to and including the middle ≈ 8k slots of an epoch that lasts approximately 24k slots in entirety
+    // https://eprint.iacr.org/2017/573.pdf p.23
+    val quietSlotsNumber = params.consensusSlotsInEpoch / 3
```
i-Alex, line 85: `quietSlotsNumber` What does it mean?

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +57,50 @@ trait ConsensusDataProvider {
...
+    val eligibleSlotsRange = (quietSlotsNumber to params.consensusSlotsInEpoch - quietSlotsNumber)
+    val eligibleForNonceCalculation =
+      allVrfOutputsWithSlots
+        .withFilter{case (_, _, slot) => eligibleSlotsRange.contains(slot)}
```
i-Alex, line 89: Why not just do `slot >= quietSlotsNumber && slot <= params.consensusSlotsInEpoch - quietSlotsNumber` instead of `.contains` loop?

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +57,50 @@ trait ConsensusDataProvider {
...
+        .map{case (proof,proofHash,  _) => ByteUtils.concatenate(proof.bytes(), proofHash.bytes())}
```
i-Alex, line 90: I'm not sure, but seems that we need to take only `proofHash.bytes()`

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +57,50 @@ trait ConsensusDataProvider {
...
+    else {
+      calculateNonceForNonGenesisEpoch(lastBlockIdInEpoch, lastBlockInfoInEpoch)
+    }
+  }
 
-    require(nonceOpt.isDefined, "No mainchain reference had been found for whole consensus epoch") //crash whole world here?
-
-    NonceConsensusEpochInfo(bigIntToConsensusNonce(nonceOpt.get))
+  private def calculateNonceForNonGenesisEpoch(lastBlockIdInEpoch: ModifierId, lastBlockInfoInEpoch: SidechainBlockInfo): NonceConsensusEpochInfo = {
+    val previousEpoch = getPreviousConsensusEpochIdForBlock(lastBlockIdInEpoch, storage.blockInfoById(lastBlockIdInEpoch))
```
i-Alex, line 74: can we use `lastBlockInfoInEpoch` instead of `storage.blockInfoById(lastBlockIdInEpoch)`?

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
oaleksandrov-ip: slots numbers which are not included in nonce calculation

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
oaleksandrov-ip: where is no loop in `contains` due we apply it to Range class

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
oaleksandrov-ip: according to whitepaper we take proof and proof hash

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
oaleksandrov-ip: yes, fixed

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
oaleksandrov-ip: check proof determinism, if Proof is not deterministic then do not include it for a nonce calculation

sdk/src/main/scala/com/horizen/block/SidechainBlock.scala
```diff
@@ -35,6 +36,7 @@ class SidechainBlock (
                        val sidechainTransactions: Seq[SidechainTransaction[Proposition, NoncedBox[Proposition]]],
                        val forgerBox: ForgerBox,
                        val vrfProof: VrfProof,
+                       val vrfProofHash: VrfProofHash,
```
albertog78, line 39: considering that vrfProofHash can be extracted by the vrfProof, storing it in any block would be a big waste of network resources. Moreover, considering our discussion about optimizing NonceConsensusEpochInfo calculation, the retrieval of the data to calculate the VRF Hash output from the vrfProof will be efficient enough to do it at block validation time and at block info calculation.

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
albertog78: only proofHash

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -52,27 +59,52 @@ trait ConsensusDataProvider {
...
+    else {
+      calculateNonceForNonGenesisEpoch(lastBlockIdInEpoch, lastBlockInfoInEpoch)
+    }
+  }
 
-    require(nonceOpt.isDefined, "No mainchain reference had been found for whole consensus epoch") //crash whole world here?
-
-    NonceConsensusEpochInfo(bigIntToConsensusNonce(nonceOpt.get))
+  private def calculateNonceForNonGenesisEpoch(lastBlockIdInEpoch: ModifierId, lastBlockInfoInEpoch: SidechainBlockInfo): NonceConsensusEpochInfo = {
+    val allVrfOutputsWithSlots: List[(VrfProof, VrfProofHash, ConsensusSlotNumber)] =
```
albertog78, line 76: simplify all the nonce calculation also avoiding to have the full list of slots related infos (proofHash) that are involved in nonce calculation. Nonce calculation can be managed by updating the digest at each slot with the current proofHash. Moreover its not needed to loop in the first third of slots.

oaleksandrov-ip:
No longer actual, see https://github.com/ZencashOfficial/Sidechains-SDK/pull/105

PR #98, opened 2020-04-21
SidechainApiRoute/ConsensusDataProvider tests update/creation

oaleksandrov-ip:


sdk/src/test/scala/com/horizen/consensus/ConsensusDataProviderTest.scala
```diff
@@ -0,0 +1,272 @@
...
+    //Full consensus info is the same for any block in epoch
+    assertEquals(consensusInfoForStartSecondEpoch, consensusInfoForEndSecondEpoch)
+
+    //Stake and root hash is the same for second and genesis epoch
+    assertEquals(consensusInfoForEndSecondEpoch.stakeConsensusEpochInfo, consensusInfoForGenesisEpoch.stakeConsensusEpochInfo)
+    //and stake is as expected
+    assertEquals(1000, consensusInfoForEndSecondEpoch.stakeConsensusEpochInfo.totalStake)
+    //nd stake root hash as expected
+    assertTrue(epochIds1.head.getBytes.take(merkleTreeHashLen).sameElements(consensusInfoForGenesisEpoch.stakeConsensusEpochInfo.rootHash))
+    //but nonce is the same as well // Is it acceptable? //
```
i-Alex, line 192: This should be discussed.
Previously nonce for genesis info for some "fixed" value. Nonce for second epoch was equal to mc ref hash in genesis block.
Now, nonce calculation algorithm is different and genesis block is a part of "quite" slots.
Maybe, we need to have a special case for second epoch nonce calculation and take genesis block vrf hash in consideration.

sdk/src/test/scala/com/horizen/consensus/ConsensusDataProviderTest.scala
```diff
@@ -0,0 +1,272 @@
...
+    val genesisBlockId = bytesToId(Utils.doubleSHA256Hash("genesis".getBytes()))
+    val genesisBlockTimestamp = 1000000
+    val networkParams = TestNetParams(
+      sidechainGenesisBlockId = genesisBlockId,
+      sidechainGenesisBlockTimestamp = genesisBlockTimestamp,
+      consensusSlotsInEpoch = slotsInEpoch,
+      consensusSecondsInSlot = 100
+    )
+
+    val tested1 = new TestedConsensusDataProvider(slotsPresentation, networkParams)
```
i-Alex, line 166: Please, check the names of the variables in test cases. It's not so clear the context, where they used.
For example:
`tested1` -> `dataProvider1` or `firstDataProvider`
`epochIds1` -> `firstDataProviderEpochIds`.

sdk/src/test/scala/com/horizen/consensus/ConsensusDataProviderTest.scala
```diff
@@ -0,0 +1,272 @@
...
+}
+
+class ConsensusDataProviderTest extends CompanionsFixture{
+  val generator: SidechainBlockFixture = new SidechainBlockFixture {}
+  val dummyWithdrawalEpochInfo = utils.WithdrawalEpochInfo(0, 0)
+
+
+  @Test
+  def test(): Unit = {
+    val slotsInEpoch = 10
```
i-Alex, line 140: In Ouroboros paper they told, that we need to have 24k of slots in epoch to be able to divide it by 3. So, the number of slots should be modulus 3.
Do we need to check `slotsInEpoch` value in `NetworkParams` or somewhere else?
The same here in current test - set `slotsInEpoch = 9`, for example.

PR #101, opened 2020-04-28
Proof shnorr

oaleksandrov-ip:


oaleksandrov-ip:
No longer actual https://github.com/ZencashOfficial/Sidechains-SDK/pull/102 shall be used instead

PR #110, opened 2020-05-11
Zendoo crypto lib integration

oaleksandrov-ip:


sdk/src/main/scala/com/horizen/SidechainHistory.scala
```diff
@@ -145,7 +145,7 @@ class SidechainHistory private (val storage: SidechainHistoryStorage,
       SidechainBlockInfo.mainchainHeaderHashesFromBlock(block),
       SidechainBlockInfo.mainchainReferenceDataHeaderHashesFromBlock(block),
       WithdrawalEpochUtils.getWithdrawalEpochInfo(block, parentBlockInfo.withdrawalEpochInfo, params),
-      vrfProofHash,
+      vrfOutput.getOrElse(new VrfOutput(Array())), //technically block is not correct from consensus point of view if vrfOutput is None
```
i-Alex, line 148: Do we use SidechainBlockInfo for verified block now or only after it was verified, so we can relay that proof is valid, so output exists? maybe we can store the output as Option or throw an exception directly here.

sdk/src/main/scala/com/horizen/SidechainHistory.scala
```diff
@@ -491,7 +491,7 @@ object SidechainHistory
       SidechainBlockInfo.mainchainHeaderHashesFromBlock(block),
       SidechainBlockInfo.mainchainReferenceDataHeaderHashesFromBlock(block),
       WithdrawalEpochInfo(1, block.mainchainBlockReferencesData.size), // First Withdrawal epoch value. Note: maybe put to params?
-      new VrfProofHash(Array()), //vrfProofHash is not used at all, so use just placeholder
+      new VrfOutput(Array()), //vrfOutput is not used at all, so use just placeholder
```
i-Alex, line 494: In case of genesis block - probably now `None` is better then `new VrfOutput(Array())`
To be discussed.

sdk/src/test/scala/com/horizen/block/SidechainBlockHeaderTest.scala
```diff
@@ -4,17 +4,34 @@ import java.io.{BufferedReader, BufferedWriter, FileReader, FileWriter}
...
 import org.junit.Test
 import org.scalatest.junit.JUnitSuite
 
 import scala.util.{Failure, Success}
 
 class SidechainBlockHeaderTest extends JUnitSuite with CompanionsFixture with SidechainBlockFixture {
+  val vrfGenerationDataSeed = 178
+  //set to true for update vrf related data
+  if (false) {
+    VrfGeneratedDataProvider.updateVrfProof(vrfGenerationDataSeed)
```
i-Alex, line 22: Can we add to the generated files names some suffix to describe where this fixed proof used, like "blockheader_vrfproof178"? What do you think about this?

sdk/src/test/scala/com/horizen/vrf/VrfFunctionsTest.scala
```diff
@@ -0,0 +1,67 @@
...
+  def sanityCheck(): Unit = {
+    assertNotEquals(vrfProofBytes.deep, vrfOutputBytes.deep)
+    assertTrue(VrfLoader.vrfFunctions.publicKeyIsValid(publicBytes))
+    assertTrue(vrfProofCheck)
+    assertTrue(vrfOutputBytes.nonEmpty)
+  }
+
+
+  @Test
+  def determinismCheck(): Unit = {
```
i-Alex, line 31: can be "vrfOutputDeterminismCheck".
I would suggest to add the comment that proof is not deterministic instead of commenting the check in line 40 or even add another test and mark in `@Ignored`

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
```diff
@@ -36,57 +36,89 @@ class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+        log.error(s"Failed to forge block for ${nextConsensusEpochNumber} epoch ${nextConsensusSlotNumber} slot due:" , ex)
+        ForgeFailed(ex)
     }
   }
 
-  private def getSecretsAndProof(wallet: SidechainWallet,
-                                 vrfMessage: VrfMessage,
-                                 forgerBox: ForgerBox,
-                                 merklePath: MerklePath): Option[(ForgerBox, MerklePath, PrivateKey25519, VrfProof, VrfProofHash)] = {
+  private def getSecretsAndProof(wallet: SidechainWallet, vrfMessage: VrfMessage, forgerBox: ForgerBox, merklePath: MerklePath) = {
```
i-Alex, line 89: please, define method return type as it was before

sdk/src/main/scala/com/horizen/validation/ConsensusValidator.scala
```diff
@@ -34,6 +34,9 @@ class ConsensusValidator extends HistoryBlockValidator with ScorexLogging {
 
 
   private def validateNonGenesisBlock(verifiedBlock: SidechainBlock, history: SidechainHistory): Unit = {
+    val parentBlockInfo: SidechainBlockInfo = history.blockInfoById(verifiedBlock.parentId)
```
i-Alex, line 37: something is wrong here after merge. Please, look deeply inside `ourboros_vrf_all_nonce` branch code

sdk/src/main/java/com/horizen/vrf/VrfFunctions.java
```diff
@@ -23,7 +23,7 @@
     boolean publicKeyIsValid(byte[] publicKeyBytes);
 
     //Return Vrf proof hash for given proof / key / message; return None if proof is not valid
-    Optional<byte[]> vrfProofToVrfOutput(byte[] publicKeyBytes, byte[] message, byte[] proofBytes);
+    Optional<byte[]> proofToOutput(byte[] publicKeyBytes, byte[] message, byte[] proofBytes);
```
i-Alex, line 26: `createVrfProof` -> `createProof` as well.

PR #112, opened 2020-05-19
Ommers from different epochs support added. Unit tests updated.

i-Alex:


sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
```diff
@@ -98,11 +100,21 @@ trait ConsensusDataProvider {
...
+                                       initialBlockInfo: SidechainBlockInfo,
+                                       eligibleSlotsRangeStart: Int,
+                                       eligibleSlotsRangeEnd: Int,
+                                       acc: Seq[(VrfOutput, ConsensusSlotNumber)]): MessageDigest = {
     require(!isGenesisBlock(initialBlockId)) //genesis nonce calculation shall be done in other way
 
     val digest: MessageDigest = MessageDigest.getInstance("SHA-256")
 
+    // Update digest with accumulated values first, that are stored in reverse order as well.
+    for(entry <- acc) {
```
i-Alex, line 113: Maybe, change "acc" to something with more context.
And set:
```
val output = entry._1
val slot = entry._2
```

sdk/src/main/scala/com/horizen/validation/ConsensusValidator.scala
```diff
@@ -71,42 +71,69 @@ class ConsensusValidator extends HistoryBlockValidator with ScorexLogging {
...
-        isPreviousEpochOmmer = true
-        val vrfOutput = history.getVrfOutput(ommer.header, previousFullConsensusEpochInfo.nonceConsensusEpochInfo)
-          .getOrElse(throw new IllegalStateException(s"VRF check for block ${ommer.header.id} had been failed"))
-        verifyForgerBox(ommer.header, previousFullConsensusEpochInfo.stakeConsensusEpochInfo, vrfOutput)
-        verifyOmmers(ommer, previousFullConsensusEpochInfo, null, history)
+      val ommerEpochAndSlot: ConsensusEpochAndSlot = history.timestampToEpochAndSlot(ommer.header.timestamp)
+
+      if(ommerEpochAndSlot.epochNumber < previousOmmerEpochNumber) {
+        // First ommer is from previous consensus epoch to Ommer Container epoch.
+        ommerFullConsensusEpochInfo = previousFullConsensusEpochInfo
```
oaleksandrov-ip, line 115: so when calling later `verifyOmmers(ommer, ommerFullConsensusEpochInfo, previousFullConsensusEpochInfo,	        bestKnownParentId, bestKnownParentInfo, history, accumulator)` we will pass the same data for `ommerFullConsensusEpochInfo` and for `ommerFullConsensusEpochInfo`. It is not an issue, but in such case we have inconsistency. I think it is a good style to use Option for `previousFullConsensusEpochInfo` then to avoid it.

sdk/src/test/scala/com/horizen/validation/ConsensusValidatorOmmersTest.scala
```diff
@@ -144,106 +138,369 @@ class ConsensusValidatorOmmersTest
...
+        (null, ConsensusSlotNumber @@ 5)
+      )
+      assertEquals("On calculate nonce: acc is different", expectedAcc, acc)
+
+      // Return nonce same as current epoch nonce
+      currentFullConsensusEpochInfo.nonceConsensusEpochInfo
+    })
+
+    var switchedEpochConsensusValidator = new ConsensusValidator {
+      /*override def verifyVrfProofAndHash(history: SidechainHistory, header: SidechainBlockHeader, message: VrfMessage): Unit = {
```
oaleksandrov-ip, line 276: do we need that comment?

sdk/src/main/scala/com/horizen/consensus/ConsensusDataProvider.scala
i-Alex: agree

sdk/src/test/scala/com/horizen/validation/ConsensusValidatorOmmersTest.scala
i-Alex: I just left it for further discussion. I guess, that this should be replaced somehow with VRFOuput check, but this is more difficult than for message.

sdk/src/main/scala/com/horizen/validation/ConsensusValidator.scala
```diff
@@ -71,42 +71,69 @@ class ConsensusValidator extends HistoryBlockValidator with ScorexLogging {
...
   private[horizen] def verifyOmmers(ommersContainer: OmmersContainer,
-                                     currentFullConsensusEpochInfo: FullConsensusEpochInfo,
-                                     previousFullConsensusEpochInfo: FullConsensusEpochInfo,
-                                     history: SidechainHistory): Unit = {
+                                    currentFullConsensusEpochInfo: FullConsensusEpochInfo,
+                                    previousFullConsensusEpochInfo: FullConsensusEpochInfo,
+                                    bestKnownParentId: ModifierId,
+                                    bestKnownParentInfo: SidechainBlockInfo,
+                                    history: SidechainHistory,
+                                    previousEpochOmmersInfoAccumulator: Seq[(VrfOutput, ConsensusSlotNumber)]
```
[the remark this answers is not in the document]
i-Alex: I would prefer to have a Seq, because having an immutable data structure is more likely to use by passing it recursively to several "branches" that will further modify it, without modification of original one.

sdk/src/main/scala/com/horizen/validation/ConsensusValidator.scala
i-Alex: Agree.
Let's expect `previousFullConsensusEpochInfo: Option[...]` and add `ommerPreviousFullConsensusEpochInfo: Option[...]` after line 108, so will set/modify it properly and pass it for subommers validation.

sdk/src/main/scala/com/horizen/validation/ConsensusValidator.scala
```diff
@@ -46,7 +46,7 @@ class ConsensusValidator extends HistoryBlockValidator with ScorexLogging {
 
     val lastBlockInPreviousConsensusEpochInfo: SidechainBlockInfo = history.blockInfoById(history.getLastBlockInPreviousConsensusEpoch(verifiedBlock.timestamp, verifiedBlock.parentId))
     val previousFullConsensusEpochInfo: FullConsensusEpochInfo = history.getFullConsensusEpochInfoForBlock(lastBlockInPreviousConsensusEpochInfo.timestamp, lastBlockInPreviousConsensusEpochInfo.parentId)
-    verifyOmmers(verifiedBlock, currentConsensusEpochInfo, previousFullConsensusEpochInfo, verifiedBlock.parentId, parentBlockInfo, history, Seq())
+    verifyOmmers(verifiedBlock, currentConsensusEpochInfo, Some(previousFullConsensusEpochInfo), verifiedBlock.parentId, parentBlockInfo, history, Seq())
```
oaleksandrov-ip, line 49: btw better use Option() instead of Some(), because in first case we get correct auto type Option, but not Some.

PR #114, opened 2020-05-22
Removed previousEndEpochBlockHash from Backward Transfer Certificates

oaleksandrov-ip:


sdk/src/main/scala/com/horizen/block/WithdrawalEpochCertificate.scala
```diff
@@ -76,11 +75,8 @@ object WithdrawalEpochCertificate {
       currentOffset += o.size
     }
 
-    val nounce: Array[Byte] = BytesUtils.reverseBytes(certificateBytes.slice(currentOffset, currentOffset + 32))
-    currentOffset += 32
-
     new WithdrawalEpochCertificate(certificateBytes.slice(offset, currentOffset), version,
-      sidechainId, epochNumber, endEpochBlockHash, Array(), Array(), totalAmount, fee, quality, transactionOutputs, outputs)
+      sidechainId, epochNumber, endEpochBlockHash, Array(), totalAmount, fee, quality, transactionOutputs, outputs)
```
i-Alex, line 79: @SergeyMakukha, the last version of `merge_support_branch` in zen core has different serialization:
https://github.com/ZencashOfficial/zen/pull/231/files#diff-35248149be7d2185b5cfa89c266e2da5R49
In general Certificate must be deserialized in the following order:
1. nVersion
2. scId
3. epochNumber
4. endEpochBlockHash
5. CTxIn vector
6. CTxOut vector
7. CBackwardTransferOut vector

Currently `quality` is not a parent of Certificate, so can be temporary hardcoded to some value, like 1.
Same for proof, can be hardcoded with zero-array with the length expected for zendoo.

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
```diff
@@ -161,14 +163,16 @@ class SidechainStateStorage(storage: Storage, sidechainBoxesCompanion: Sidechain
     }
   }
 
+  def getLastWithdrawalCertificatePreviousMcBlockHashOpt: Option[Array[Byte]] = storage.get(lastWithdrawalCertificatePreviousMcBlockHashKey).asScala.map(_.data)
```
i-Alex, line 166: I think, that for last certificate it should be not "previous". but "end mc block hash". It's previous for current certificate, but nor for last one.
`getLastWithdrawalCertificatePreviousMcBlockHashOpt` -> `getLastCertificateEndEpochMcBlockHashOpt`
same for `lastWithdrawalCertificatePreviousMcBlockHashKey` -> `lastCertificateEndEpochMcBlockHashKey`

sdk/src/test/scala/com/horizen/integration/storage/SidechainStateStorageTest.scala
```diff
@@ -212,12 +213,15 @@ class SidechainStateStorageTest
 
     val mod1Version = getVersion
     val mod1WithdrawalEpochInfo = WithdrawalEpochInfo(firstWithdrawalEpochNumber, 1)
+    val mod1mcBlockHashInCertificate = Option(Array[Byte](1, 2, 3, 4, 5, 6 ,7))
```
i-Alex, line 216: Let's have 32 bytes random here with predefined seed.

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -107,6 +107,7 @@ class SidechainApp @Inject()
       sidechainId = BytesUtils.fromHexString(sidechainSettings.genesisData.scId),
       sidechainGenesisBlockId = genesisBlock.id,
       genesisMainchainBlockHash = genesisBlock.mainchainHeaders.head.hash,
+      preGenesisMainchainBlockHash = genesisBlock.mainchainHeaders.head.hash,
```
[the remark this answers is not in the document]
oaleksandrov-ip: I prefer to leave it as is, the proposed name `genesisMainchainBlockParentHash` is good but it could be hard to distinguish `genesisMainchainBlockHash` and `genesisMainchainBlockParentHash`, but it easy to distinguish  `genesisMainchainBlockHash` and `preGenesisMainchainBlockHash`

sdk/src/main/scala/com/horizen/block/MainchainBlockReferenceData.scala
```diff
@@ -11,13 +11,12 @@ import scorex.util.serialization.{Reader, Writer}
...
-                                   @JsonSerialize(using = classOf[JsonMerklePathOptionSerializer])
-                                   mproof: Option[MerklePath],
+                                        headerHash: Array[Byte],
+                                        sidechainRelatedAggregatedTransaction: Option[MC2SCAggregatedTransaction],
+                                        @JsonSerialize(using = classOf[JsonMerklePathOptionSerializer])
+                                        mProof: Option[MerklePath],
                                         proofOfNoData: (Option[SidechainCommitmentEntryProof], Option[SidechainCommitmentEntryProof]),
-                                        backwardTransferCertificate: Option[WithdrawalEpochCertificate]
-                                 ) extends BytesSerializable {
+                                        backwardTransferCertificate: Option[WithdrawalEpochCertificate]) extends BytesSerializable {
```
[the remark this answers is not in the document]
oaleksandrov-ip: done

sdk/src/main/scala/com/horizen/block/WithdrawalEpochCertificate.scala
oaleksandrov-ip: let leave as is, and change it asap as all field will be supported in MC

sdk/src/main/scala/com/horizen/storage/SidechainStateStorage.scala
oaleksandrov-ip: agree

sdk/src/main/scala/com/horizen/SidechainApp.scala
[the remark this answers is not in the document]
i-Alex: Honestly, I don't like "pre" at all. Because it means "previous genesis mc block hash", that is not about block before - parent block, but about another genesis block, which is misleading here.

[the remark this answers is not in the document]
i-Alex: `genesisBlock.mainchainHeaders.head.hash` an error here -> must be `...head.hashPrevBlock`

sdk/src/main/scala/com/horizen/block/WithdrawalEpochCertificate.scala
i-Alex: ok. that's mostly for  @SergeyMakukha in STF branch with Certificate test case.

PR #116, opened 2020-05-25
STF/CCTP Testing

i-Alex:


sdk/src/main/scala/com/horizen/mainchain/api/RpcMainchainNodeApi.scala
```diff
@@ -48,17 +57,16 @@ class RpcMainchainNodeApi(val sidechainSettings: SidechainSettings)
...
-    } else ""
+
     val response = callRpc("send_certificate "
       + encloseStringParameter(BytesUtils.toHexString(certificateRequest.sidechainId)) + " "
       + certificateRequest.epochNumber + " "
-      + encloseStringParameter(BytesUtils.toHexString(certificateRequest.endEpochBlockHash))
-      + encloseJsonParameter(objectMapper.writeValueAsString(certificateRequest.backwardTransfers))
-      + feeParam
+      + encloseStringParameter(BytesUtils.toHexString(certificateRequest.endEpochBlockHash)) + " "
+      + "\"[]\" " //encloseJsonParameter(objectMapper.writeValueAsString(certificateRequest.backwardTransfers)) + " " // TODO: fix json serialization
```
i-Alex, line 65: Todo: need fix to return plain json representation of backward transfer entries (without endlines and tabs).

sdk/src/main/scala/com/horizen/utils/WithdrawalEpochUtils.scala
```diff
@@ -24,6 +24,6 @@ object WithdrawalEpochUtils {
   }
 
   def canSubmitCertificate(withdrawalEpochInfo: WithdrawalEpochInfo, params: NetworkParams): Boolean = {
-    (withdrawalEpochInfo.epoch > 0) && (withdrawalEpochInfo.lastEpochIndex < params.withdrawalEpochLength / 5)
+    (withdrawalEpochInfo.epoch > 0) && (withdrawalEpochInfo.lastEpochIndex <= params.withdrawalEpochLength / 5)
```
i-Alex, line 27: I think, that check must be inclusive. `lastEpochIndex` starts from 0, but when first mc block ref data arrived == 1

PR #117, opened 2020-05-27
Zendoo integration all

oaleksandrov-ip:


qa/readme.md
```diff
@@ -0,0 +1,36 @@
...
+        ```
+
+**Template configuration files**
+
+Template configuration files are located in directory resources. 
+
+File template.conf is template for tests of SC node(s) is connected to MC node(s).
+
+File template_predefined_genesis.conf is template for test of SC node(s) in standalone mode(without connection to MC node).
+
```
i-Alex, line 33: We need to add info how to set MC cli/zend and SC boot tool and simple app (or any other app)

sdk/pom.xml
```diff
@@ -44,7 +44,7 @@
         <dependency>
             <groupId>com.horizen</groupId>
             <artifactId>zendoo-sc-cryptolib</artifactId>
-            <version>0.2.1-SNAPSHOT</version>
+            <version>0.2-SNAPSHOT</version>
```
i-Alex, line 47: is it correct? Why not 0.2.6 generated yesterday?

sdk/pom.xml
```diff
@@ -311,6 +305,7 @@
                     <include>schnorr_sk00_hex</include>
                     <include>sample_proving_key_7_keys_with_threshold_5</include>
                     <include>sample_vk_7_keys_with_threshold_5</include>
+					<include>pippo_vk</include>
```
i-Alex, line 308: can be removed.
And moreover, let's have a regression test for proof verification.

sdk/src/main/java/com/horizen/transaction/mainchain/SidechainCreation.java
```diff
@@ -63,8 +58,8 @@ public ForgerBox getBox() {
         );
     }
 
-    public byte[] getBackwardTransferPoseidonRootHash() {
-        return new byte[0];
+    public byte[] genSysConstant() {
```
i-Alex, line 61: Probably,  `genSysConstant` -> `getSysConstant`

sdk/pom.xml
oaleksandrov-ip: we need to define it, current version in development branch in zensoo-sc-cryptolib is 0.2-SNAPSHOT

sdk/src/main/java/com/horizen/transaction/mainchain/SidechainCreation.java
oaleksandrov-ip: we use the same name `genSysConstant` everywhere, so it's not missprint

i-Alex: ok, i got it. Thought that `gen` is `generate`, but it's actually like get generated sys constant

PR #145, opened 2020-06-23
Unit tests were fixed.

SergeyMakukha:


qa/run_sc_tests.py
```diff
@@ -40,15 +40,15 @@ def run_tests(log_file):
     result = run_test(MCSCNodesAlive())
     assert_equal(0, result, "mc_sc_nodes_alive test failed!")
 
-    result = run_test(SCBackwardTransfer())
-    assert_equal(0, result, "sc_backward_transfer test failed!")
-
     result = run_test(SCBootstrap())
```
i-Alex, line 43: `MCSCNodesAlive` and `SCBootstrap` tests should be executed before `MCSCForging1`

qa/tests/sc_generate_test_data.py
```diff
@@ -0,0 +1,285 @@
...
+        assert_equal(0, we0_cert["cert"]["totalAmount"], "Sidechain total amount in certificate is wrong.")
+
+        # Generate MC block and verify that certificate is present
+        we1_2_mcblock_hash = mc_node.generate(1)[0]
+        assert_equal(0, mc_node.getmempoolinfo()["size"], "Certificate expected to be removed from MC node mempool.")
+        assert_equal(1, len(mc_node.getblock(we1_2_mcblock_hash)["tx"]), "MC block expected to contain 1 transaction.")
+        assert_equal(1, len(mc_node.getblock(we1_2_mcblock_hash)["cert"]), "MC block expected to contain 1 Certificate.")
+        assert_equal(we0_certHash, mc_node.getblock(we1_2_mcblock_hash)["cert"][0], "MC block expected to contain certificate.")
+
+        #Write test data - sidechain forward transfer transaction and block, containing it.
```
i-Alex, line 157: Comment message is inconsistent. BT?

qa/tests/sc_generate_test_data.py
```diff
@@ -0,0 +1,285 @@
...
+        bt_block = mc_node.getblock(we1_2_mcblock_hash, False)
+        bt_transaction = mc_node.gettransaction(we0_cert["certid"])["hex"]
+        self.write_test_data(test_data_file, "Sidechain backward transfer MC block", bt_block)
+        self.write_test_data(test_data_file, "Sidechain backward transfer MC transaction", bt_transaction)
+
+
+        # Generate SC block and verify that certificate is synced back
```
i-Alex, line 164: Why do we need the rest of code?
Honestly, I don't see any benefits by using this test at all. It's just a copy-paste of BT test with printing of some data.
So, we can just use the original test.
As for me, the idea of having python script is to help automatically regenerate **all data** that we need for `MC Block` / `MC Tx` / `MC Cert` Unit tests.
I suggest to delete this file.

sdk/pom.xml
```diff
@@ -316,8 +316,7 @@
 					<include>new_mc_blocks/mc_block_empty_sidechains</include>
 					<include>new_mc_blocks/mc_block_create_3_sidechains</include>
 					<include>new_mc_blocks/mc_block_forward_transfer_2_sidechains</include>
-					<include>new_mc_blocks/mc_block_forward_transfer_3_backward_transfer</include>
```
i-Alex, line 319: What was inside `mc_block_forward_transfer_2_sidechains` and `mc_block_forward_transfer_3_backward_transfer`?
What case was tested? Do we still have that test cases somewhere else?

i-Alex:
The changes in tests are outdated.

PR #146, opened 2020-06-24
BT destination address endianness fixed: LE for proof and schnorr cre…

i-Alex:
…ation and BE for certificate creation RPC command. Integration BT test updated.


sdk/src/main/java/com/horizen/utils/BytesUtils.java
```diff
@@ -190,4 +195,54 @@ public static boolean contains(Collection<byte[]> collection, byte[] value) {
...
         return false;
     }
+
+    private static int HorizenPublicKeyAddressPrefixLength = 2;
+    private static int HorizenPublicKeyAddressHashLength = 20;
+    private static int HorizenPublicKeyAddressChecksumLength = 4;
+
+    private static int HorizenPublicKeyAddressBase58CheckLength = 35;
+
+    private static byte[] PublicKeyMainNetPrefix = BytesUtils.fromHexString("2089"); // "zn"
```
oaleksandrov-ip, line 205: the name shall starts from a lower case or the whole name shall be in CAMEL_CASE

sdk/src/main/java/com/horizen/utils/BytesUtils.java
```diff
@@ -190,4 +195,54 @@ public static boolean contains(Collection<byte[]> collection, byte[] value) {
...
+        if(address.length() != HorizenPublicKeyAddressBase58CheckLength)
+            throw new IllegalArgumentException(String.format("Incorrect Horizen public key address length $d", address.length()));
+
+        byte[] addressBytesWithChecksum = Base58.decode(address).get();
+        byte[] addressBytes = Arrays.copyOfRange(addressBytesWithChecksum, 0, HorizenPublicKeyAddressPrefixLength + HorizenPublicKeyAddressHashLength);
+
+
+        // Check version
+        byte[] prefix = Arrays.copyOfRange(addressBytes, 0, HorizenPublicKeyAddressPrefixLength);
+        if(params instanceof MainNetParams) {
```
oaleksandrov-ip, line 221: better add parameter in params of public key address formatinstead of use instance of

sdk/src/main/java/com/horizen/utils/BytesUtils.java
```diff
@@ -190,4 +195,54 @@ public static boolean contains(Collection<byte[]> collection, byte[] value) {
...
+        byte[] checksum = Arrays.copyOfRange(addressBytesWithChecksum, HorizenPublicKeyAddressPrefixLength + HorizenPublicKeyAddressHashLength, addressBytesWithChecksum.length);
+        byte[] calculatedChecksum = Arrays.copyOfRange(Utils.doubleSHA256Hash(addressBytes), 0, HorizenPublicKeyAddressChecksumLength);
+        if(!Arrays.equals(calculatedChecksum, checksum))
+            throw new IllegalArgumentException("Broken Horizen public key address: checksum is wrong.");
+
+        return publicKeyHash;
+    }
+
+    public static String toHorizenPublicKeyAddress(byte[] publicKeyBytes, NetworkParams params) {
+        if(publicKeyBytes.length != 20)
```
oaleksandrov-ip, line 240: 20 to named constant?

sdk/src/main/java/com/horizen/utils/BytesUtils.java
i-Alex: changed to UPPER_CASE as in other places.

sdk/src/main/java/com/horizen/utils/BytesUtils.java
i-Alex: right. Used the constant defined above.

sdk/src/main/java/com/horizen/utils/BytesUtils.java
i-Alex: let's keep as is.
Maybe will be different after solving the main net case in Json encoder.

Issue #152, opened 2020-06-29
Provide a Bash script for Linux that automates steps during sidechain node setup


Cermelli:
**HDE task**

Starting points:

- https://github.com/HorizenOfficial/Sidechains-SDK/blob/master/README.md
- https://github.com/HorizenOfficial/zend_oo/blob/sidechains_testnet/README.md

Script should:

- Automate all the steps needed to declare sidechains and run a first sidechain node
- Generate all the needed keys
- Include customisation of the initial amount to be sent to the sidechain during sidechain declaration
- Declare the sidechain on the mainchain with the correct command
- Generate sidechain configuration files
- Start a sidechain node and start forging
- Include help message with detailed description of possible usage
- Be written in Bash on Linux

Installing additional tools is allowed (e.g. jq).

**For more info:** https://hde.horizen.global/tasks
***Limit first accepted submissions

morganblack25:
pero sera compatible también para todos los sistemas operativos  de 32 bits ?

rosariopabst:
Por ahora zend_oo y el sidechain-SDK solo es compatible con 64 bit. Vas a empezar a desarollar el bash script/bounty?

morganblack25:
si eso quiero realizar pero mi sistema operativo es de 32bits

jtoownus:
Hey @morganblack25 ,
if possible, please keep discussions in English so everybody can follow along. Thank you :)

Rifat0000:
We love u zen.... Every one know zen is best

salvatoretrimarchi:
Yes of course, how may I help you? I'm Bash Developer

ponta9:
I’ve got a working bash script that generates all keys and customizes a sidechain node with regtest flag at the moment(with custom amount). It starts the node and automatic forger. Just missing some good error logging and test it on testnet instead.

Does the script need to compile zend_oo and compile the sidechain app?

cronicc:
> Does the script need to compile zend_oo and compile the sidechain app?

@ponta9 compiling zend_oo in the script is optional, but nice to have. The script should at least have the option to provide a path to zend/zen-cli, or check that they are in $PATH. A check for the latest version should also be made (`zend -version`). You could also use the precompiled binaries from https://github.com/ZencashOfficial/zend_oo/releases/latest, but in that case authenticity of the binaries should be checked using gpg.

[...]

PR #181, opened 2020-08-02
Sidechainscript

ponta9:
Here is my bash script for the sidechain app setup. 

Additional tools used: jq

The script checks that zend_oo and sdk is up to date and compiles the sdk and app if there is no latest jar available in path .
The script also checks that java version >= 8 and maven version >= 3 is ínstalled.
#152


angievi:
Thank you!! @ponta9 we will assign this for review and get back to you!! :) -Angie.

angievi:
Hello! @ponta9, your script is being reviewed by @zennermac and will also get a couple of more reviews in the meantime. We will come back to you with any feedback and about your bounty in the next following days. thank you!

PR #183, opened 2020-08-28
Updated Readme with MALLOC_CHECK_ environment variable.

VolodymyrBabaiev:


README.md
```diff
@@ -30,6 +30,8 @@ For more details see [zendoo-sc-cryptolib](https://github.com/HorizenOfficial/ze
 * Python 2.7
 * Maven
 
+On Linux operation systems Sidechain requires to set environment variable `MALLOC_CHECK_` to 3 as well.
```
i-Alex, line 33: Let's put more comprehensive description, like:

Note: On some Linux OSs during backward transfers certificates proofs generation a extremely big RAM consumption may happen, that will lead to the process force killing by the OS. To solve this issue, set an environment variable `MALLOC_CHECK_ = 3` before starting the node.

@abi87 please, confirm if such description is fine.

i-Alex (changes requested):

README.md
abi87: a couple of three points from me:

1. export MALLOC_CHECK_=1 is a sufficient work around for us. MALLOC_CHECK_=3 would call abort upon errors and we do not need to be so drastic
2. I agree with @i-Alex on expanding explanation. I would replace 'To solve this issue, set an environment variable MALLOC_CHECK_ = 3 before starting the node.' with 'While our investigation is still ongoing, we found that setting the environment variable MALLOC_CHECK_=1 just before starting the node is a sufficiently good workaround.'
3. Should we update the simpleApp page too with this info?

Finally @lander86 and me are verifying the workaround on the our testnet machine. Maybe we could merge the update comment after we verify a few days that everything is ok?

abi87 (changes requested):
See comment for comments

README.md
i-Alex: 1,2 - Agree
3 - yeah we can to put there as well:
https://github.com/HorizenOfficial/Sidechains-SDK/blob/malloc_readme_update/examples/simpleapp/README.md
in Execution section for Linux

abi87 (approved):
with update of glibc parameter to set

albertog78 (approved):

abi87 (approved):

i-Alex (approved):

PR #202, opened 2020-09-22
Fixed wrong path

abi87:
Hello,
in building and rebuilding sidechains in testnet I noticed something wrong with the  `provingKeyFilePath `and `verificationKeyFilePath ` in the template for sidechain configurations `settings_basic.conf`.
I believe it should be fixed as I propose in this PR, so that the aforementioned files can be reach from the repo root.
Can you confirm my fix is correct?
If so, and if you are willing to accept the change, should master be targeted, or rather other branches?
Many thanks

### DIFF
--- a/examples/simpleapp/src/main/resources/settings_basic.conf
+++ b/examples/simpleapp/src/main/resources/settings_basic.conf
@@ -29,8 +29,8 @@ scorex {
     signersPublicKeys =
     signersThreshold =
     signersSecrets =
-    provingKeyFilePath = "../../sdk/src/test/resources/sample_proving_key_7_keys_with_threshold_5"
-    verificationKeyFilePath = "../../sdk/src/test/resources/sample_vk_7_keys_with_threshold_5"
+    provingKeyFilePath = "sdk/src/test/resources/sample_proving_key_7_keys_with_threshold_5"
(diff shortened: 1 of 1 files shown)


oaleksandrov-ip:
well, that path for proving files could be used as well. But, then we shall update all readme files as well.

abi87:
@oaleksandrov-ip  thanks for your reply.
Maybe I have misunderstood the workflow readme[https://github.com/HorizenOfficial/Sidechains-SDK/blob/master/examples/simpleapp/mc_sc_workflow_example.md], but I believe that starting from the repo root, the correct path does not require the '../..'.
If I have not to start from the repo root, can you specify from which folder should I start?

oaleksandrov-ip:
@abi87 for example command    java -cp ./examples/simpleapp/target/sidechains-sdk-simpleapp-0.2.4.jar;./examples/simpleapp/target/lib/* com.horizen.examples.SimpleApp ./examples/simpleapp/src/main/resources/my_settings.conf

will be incorrect, also other parts shall be verified.

abi87:
@oaleksandrov-ip ok I undestand what you mean and I agree with you.
Thanks

i-Alex:
Closed because the PR is outdated.

PR #211, opened 2020-10-12
Add pre_validate function to the ApplicationState interface

Mr-Leshiy:
I have added a new method into the ApplicationState interface - pre_validate.
This function has been executed during the block processing and BEFORE transactions validation, in contrast with the validate method, which has been executed AFTER transactions validation.
I think it is bring a more flexibility for the blockchain developers and allows to make some preparations before the transaction validation. 
As a reference to the known solutions I can refer to the Substrate https://substrate.dev/docs/en/knowledgebase/runtime/execution and their block execution. They also allows developers to bring some custom functionality before the transaction validation (on_initialize function) and after transaction validation (on_finalize  function).


i-Alex:
@Mr-Leshiy, thanks for your contribution!
Currently, if the block is going to be applied to the State, we separate the process into 2 phases: validation and application.
Validation steps are (roughly):
1) check block connectivity with the parent;
2) check that each transaction in the block is valid against the Core logic;
3) check that each transaction in the block is valid against the Application specific logic;
4) check that block is valid against the Application logic (here I mean the cases with bound transactions, for example).

Similar to the Substrate logic each of this verification point may fail, so the block will be discarded.
It's important, that validation must NOT change the State.

If the validation succeed, only then we will start to apply block to the State: at first to the application specific, then to the Core. And we expect no failures/exceptions during this phase.

It's worst to be mentioned, that there is no connectivity between Core and Application states during the block appending.
So actually, `on_initialize` and `on_finalize` can be implemented somewhere inside `ApplicationState.onApplyChanges` method without any knowledge to the core part about it.

Considering the words above, I think, that the proposed method `ApplicationState.pre_validate` has the same functionality as `ApplicationState.validate` and actually no matter what is the order in the overall validation procedure.

What do you think? Can you provide an example of possible usage?
Thanks!

Mr-Leshiy:
@i-Alex , thanks for the answer !
I can provide like an synthetic example of such case. Generally as I understood developer has an unlimited access to the third-party libraries, he can easily use some db, for example leveldb, rocksdb, sqlite etc., for custom indexing of his information from the boxes. So imagine that developer will has some  sophisticated algorithm for the fee calculation which will be defined for the whole block and will take information from the previous state from the storage (some db). And such fee should be validated for each transaction during the transaction process. So it seems naturally and natively to do such fee calculation before the transaction processing and reuse it during the transaction validation. 
I understand that it is not the breaking change suggestion, and there are another possible solutions exist, how to separate such calculation but on my opinion it seems naturally and natively to do it in the separate function that will be guarantied executes before transaction processing. But of course I can be mistaken, it is just a suggestion :).
What do you think about that ?
Thank you for your attention.

i-Alex:
To proceed with that approach, given that is not possible to store anything in the state during validation, it would be necessary to pass to all the validators an object that must be used as a transient storage for all the communications between the validators.
Instead with the current our approach the example with fees validation can be implemented by performing the fees calculation in the `ApplicationState.validate(mod)` method and in the same method then looping through the transactions to verify the correctness of the fees.
More in general if the application needs multiple validators and interacting between them this can be implemented by a single application by providing an implementation of `ApplicationState.validate(mod)` method that make use of different validators with the logic he wants. And leave only non-interactive checks of a single transaction in `ApplicationState.validate(tx)` method.

PR #218, opened 2020-10-19
Sc tx commitment

i-Alex:
Rough implementation of the ScTransactionCommitment with poseidon hash according to the `zend_oo ScCommitmentTree_to_cryptolib` draft implementation.


i-Alex:
This PR is outdated, since we calculate ScTxCommitment using sc-cryptolib.

PR #219, opened 2020-10-19
added application forger support (apps can add transactions on forge …

paolocappelletti:
added application forger support (apps can add transactions on forge time)

### DIFF
--- a/sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
+++ b/sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
@@ -17,10 +17,12 @@ import scorex.util.{ModifierId, ScorexLogging}
 import com.horizen.chain._
 import com.horizen.vrf.VrfOutput
 
+import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.util.{Failure, Success, Try}
 
 class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
                           companion: SidechainTransactionsCompanion,
+                          applicationForger: ApplicationForger,
                           val params: NetworkParams,
                           allowNoWebsocketConnectionInRegtest: Boolean) extends ScorexLogging with TimeToEpochSlotConverter {
   type ForgeMessageType = GetDataFromCurrentView[SidechainHistory, SidechainState, SidechainWallet, SidechainMemoryPool, ForgeResult]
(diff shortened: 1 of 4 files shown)


sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
```diff
@@ -225,9 +227,13 @@ class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
       if (branchPointInfo.referenceDataToInclude.size == withdrawalEpochMcBlocksLeft) { // SC block is going to become the last block of the withdrawal epoch
         Seq() // no SC Txs allowed
       } else { // SC block is in the middle of the epoch
-        nodeView.pool.take(SidechainBlock.MAX_SIDECHAIN_TXS_NUMBER) // TO DO: problems with types
+        val mempoolTransactions = nodeView.pool.take(SidechainBlock.MAX_SIDECHAIN_TXS_NUMBER) // TO DO: problems with types
           .map(t => t.asInstanceOf[SidechainTransaction[Proposition, NoncedBox[Proposition]]])
-          .toSeq
+
+        val applicationForgeTransactions = applicationForger.onForge(parentBlockInfo, blockSignPrivateKey.publicImage())
```
i-Alex, line 233: Are we sure that transactions proposed by ApplicationForger are valid?
Currently, if ApplicationForger will provide transaction that is not valid or which has a spent box intersections (double spend), it will lead to an invalid SC block proposed by the Forger.
Another point of discussion is what should be included first: `mempoolTransactions` or `applicationForgeTransactions`? We should keep in mind the max block size as well.

The first solution that come to my mind is:
1) get mempoolTransactions
2) pass them as an immutable collection as an argument to `applicationForger.onForge`
3) for the applicationForgeTransactions verify them with the compatibility checker against the previously added transactions, then against the SidechainState.
If everything is ok, forge a block.
What do you think?

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
paolocappelletti: I get the point and  agree.
I thought we could count on the check  at block acceptance by the other nodes, but in this way we can stop the process earlier and avoid to forge a wrong block at all

i-Alex:
We are not going to have this feature in the core. At least for now.

PR #220, opened 2020-10-22
Adding dependabot config

lander86:
Dependabot config added to properly configure the target branch of the PRs.


dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+
+version: 2
+updates:
+  # Enable version updates for maven
+  - package-ecosystem: "maven"
+    directory: "/"
+    target-branch: "dev"
+    reviewers:
+      - "i-Alex"
+      - "abi87"
```
i-Alex, line 11: abi87 is Alberto Benegiamo. Do we need him or Alberto Garoffolo (albertog78)?

lander86: Added Albene because don't like to have only one person as reviewer (and Albene volunteered).
If you think is necessary i can add also Algaro

cronicc (approved):

PR #235, opened 2021-02-11
issue 233: add helpers for submit tx to mempool and retrieve the NodeView

paolocappelletti:
added the 2 helpers requested by the issue, and made also these changes (after a sync with Algaro):
- added a "default" sidechain guice module which must be overridden by custom app module
- removed the inner ijector in SidechainApp  and added little more injection inside SidechainApp   (could be greater but would require a more heavy code-refactoring)

Note: I started the branch from the master branch and not  "dev" because I needed it urgently for Celsius and wanted to be based on a more  "stable" version

### DIFF
--- a/sdk/src/main/scala/com/horizen/SidechainApp.scala
+++ b/sdk/src/main/scala/com/horizen/SidechainApp.scala
@@ -36,13 +36,11 @@ import scorex.core.settings.ScorexSettings
 import scorex.core.transaction.Transaction
 import scorex.core.{ModifierTypeId, NodeViewModifier}
 import scorex.util.ScorexLogging
-
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Map
 import scala.collection.mutable
 import scala.io.Source
 import com.horizen.network.SidechainNodeViewSynchronizer
-
 import scala.util.Try
 
 
@@ -65,8 +63,10 @@ class SidechainApp @Inject()
    @Named("CustomApiGroups") val customApiGroups: JList[ApplicationApiGroup],
    @Named("RejectedApiPaths") val rejectedApiPaths : JList[Pair[String, String]]
   )
-  extends Application with ScorexLogging with Module
+  extends Application  with ScorexLogging with Module
 {
+
+
   override type TX = SidechainTypes#SCBT
   override type PMOD = SidechainBlock
   override type NVHT = SidechainNodeViewHolder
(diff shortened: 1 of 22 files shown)


sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -289,15 +279,11 @@ class SidechainApp @Inject()
...
-    binder.bind(classOf[SidechainProofsCompanion])
-      .toInstance(sidechainProofsCompanion)
-
-    binder.install(new FactoryModuleBuilder()
-      .build(classOf[SidechainCoreTransactionFactory]))
+  def getTransactionActorRef(): ActorRef = {
+    return sidechainTransactionActorRef
+  }
+  def getNodeViewActorRef(): ActorRef = {
+    return nodeViewHolderRef
```
i-Alex, line 286: I'm a bit worried, that we expose the NodeViewHolderRef outside the SidechainApp. Why not to return an instance of NodeViewHelper directly?

i-Alex (commented):
We need to update SDK version from 0.2.6 to 0.2.7 if we want to merge directly into master branch and release.

sdk/src/main/scala/com/horizen/SidechainApp.scala
paolocappelletti: if we return it directly, we would not be able to inject directly NodeViewHelper in the app because we would have a circular dependency (sidechainpp require our app module to start that in turns would require it to access to the viewhelper)

i-Alex: Ok. So can we return some wrapper on top of `NodeViewHolderRef` then? Like a NodeViewProvider with the "withNodeView` method, for example.

Why I'm talking about this, is that giving direct access to the `NodeViewHolderRef` outside the SidechainApp is strictly forbidden, because will give a direct access to the SidechainHistory/State/Mempool/Wallet that have modification methods. The usage of which will broke the node.

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -63,10 +60,14 @@ class SidechainApp @Inject()
    @Named("WalletForgingBoxesInfoStorage") val walletForgingBoxesInfoStorage: Storage,
    @Named("ConsensusStorage") val consensusStorage: Storage,
    @Named("CustomApiGroups") val customApiGroups: JList[ApplicationApiGroup],
-   @Named("RejectedApiPaths") val rejectedApiPaths : JList[Pair[String, String]]
+   @Named("RejectedApiPaths") val rejectedApiPaths : JList[Pair[String, String]],
+   val sidechainCoreTransactionFactory : SidechainCoreTransactionFactory,
+   val sidechainTransactionsCompanion : SidechainTransactionsCompanion
```
i-Alex, line 65: SimpleApp is broken. Please, run  python tests full suit and fix the issues.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/SidechainApp.scala
paolocappelletti: fixed

sdk/src/main/scala/com/horizen/SidechainApp.scala
paolocappelletti: yes, added the wrappers

i-Alex (dismissed):

i-Alex (approved):

PR #259, opened 2021-05-14
CoboundaryMarling PS dlog and snark keys initialization.

i-Alex:
Draft implementation: proper sc-cryplolib lib version and python tests update required."


i-Alex:
Cannibalized by https://github.com/HorizenOfficial/Sidechains-SDK/pull/252.

PR #270, opened 2021-07-28
Versioning

VolodymyrBabaiev:


sdk/src/main/java/com/horizen/transaction/BoxTransaction.java
```diff
@@ -43,6 +43,9 @@
     @JsonProperty("typeId")
     public abstract byte transactionTypeId();
 
+    @JsonProperty("version")
+    public abstract byte version();
```
i-Alex, line 47: add `@Override` since method was declared in the parent class.
And add the same for `transactionTypeId`, pls. That was missed.

sdk/src/test/scala/com/horizen/block/SidechainBlockHeaderTest.scala
```diff
@@ -137,8 +137,20 @@ class SidechainBlockHeaderTest extends JUnitSuite with CompanionsFixture with Si
...
+      case Success(_) =>
+        jFail("Signed header with negative timestamp expected to be semantically Invalid.")
+      case Failure(e) =>
+        assertEquals("Different exception type expected during semanticValidity.",
+          classOf[InvalidSidechainBlockHeaderException], e.getClass)
+    }
+
+    // Test 4: unsupported block version
+    val invalidHeader = baseUnsignedHeader.copy(version = Byte.MaxValue)
+    headerSignature = forgerMetadata.blockSignSecret.sign(invalidHeader.messageToSign)
```
i-Alex, line 152: Check `messageToSign` implementation for `SidechainBlockHeader`. It should consider `version`.

sdk/src/main/java/com/horizen/transaction/BoxTransaction.java
i-Alex: check `messageToSign` method. It must consider `version`.

i-Alex (changes requested):

sdk/src/main/java/com/horizen/transaction/BoxTransaction.java
VolodymyrBabaiev: Done.

sdk/src/test/scala/com/horizen/block/SidechainBlockHeaderTest.scala
VolodymyrBabaiev: Done.

sdk/src/main/scala/com/horizen/block/SidechainBlockHeader.scala
```diff
@@ -47,6 +47,7 @@ case class SidechainBlockHeader(
   lazy val messageToSign: Array[Byte] = {
     Bytes.concat(
       idToBytes(parentId),
+      Array[Byte]{version},
```
i-Alex, line 50: Do the same change for the `BoxTransaction messageToSign` method.

i-Alex (approved):

sdk/src/main/scala/com/horizen/block/SidechainBlockHeader.scala
VolodymyrBabaiev: Done.

i-Alex (approved):

PR #282, opened 2021-09-14
Enhance sc creation procedure

davirax:


examples/simpleapp/src/main/resources/sc_settings.conf
```diff
@@ -3,51 +3,62 @@ scorex {
     logDir = /tmp/scorex/data/log
 
     restApi {
-        bindAddress = "127.0.0.1:9085"
-        api-key-hash = ""
-        timeout = 5s
+      bindAddress = "127.0.0.1:9085"
```
i-Alex, line 6: check this file. Probably, you did the rollback to far in the past. This one has outdated info.
I would suggest to restore the file exactly how it was in a base branch.

i-Alex (changes requested):

i-Alex (approved):

PR #284, opened 2021-10-07
STF: python tests random failure fixes

i-Alex:


qa/sc_ceased.py
```diff
@@ -72,10 +72,10 @@ def run_test(self):
...
 
         # Generate MC blocks to reach the end of the withdrawal epoch
-        mc_block_hashes = mc_node.generate(mc_blocks_left_for_we)
-        # Generate SC blocks to sync with MC node
-        for x in range(int(math.ceil(len(mc_block_hashes)/3.0))):
-            generate_next_block(sc_node, "first node")
+        mc_node.generate(mc_blocks_left_for_we)
+
+        # Generate 1 more SC block to sync with MC
+        generate_next_block(sc_node, "first node")
```
i-Alex, line 78: Optimization: since we support multiple MC Block refs to be included into a single SC block, it is enough to generate only 1 SC block.
Same in other places.

VolodymyrBabaiev (approved):

PR #286, opened 2021-10-07
Api calling along node synchronization

davirax:
Created Prioritized mailbox for the sidechainNodeviewholder in order to give priority to:

ReceivableMessages.GetDataFromCurrentView -> 0     // for internal node operations

 ReceivableMessages.GetDataFromCurrentSidechainNodeView  -> 1   // for the api call

-  -> 100    // all the others

### DIFF
--- a/examples/simpleapp/.gitignore
+++ b/examples/simpleapp/.gitignore
@@ -9,3 +9,5 @@ src/main/resources/sc_settings_with_mc.conf
 .settings/
 .vscode/
 /src/main/MyOwnSettings/
+../Dockerfile
+../entrypoint.sh
--- a/examples/simpleapp/.gitignore
+++ b/examples/simpleapp/.gitignore
@@ -8,4 +8,3 @@ src/main/resources/sc_settings_with_mc.conf
 .classpath
 .settings/
 .vscode/
-../entrypoint.sh
--- a/examples/simpleapp/.gitignore
+++ b/examples/simpleapp/.gitignore
@@ -8,6 +8,4 @@ src/main/resources/sc_settings_with_mc.conf
 .classpath
 .settings/
 .vscode/
-/src/main/MyOwnSettings/
-../Dockerfile
 ../entrypoint.sh
--- a/examples/simpleapp/src/main/java/com/horizen/examples/SimpleAppModule.java
+++ b/examples/simpleapp/src/main/java/com/horizen/examples/SimpleAppModule.java
@@ -24,7 +24,6 @@ import com.horizen.transaction.BoxTransaction;
 import com.horizen.transaction.TransactionSerializer;
 import com.horizen.wallet.*;
 import com.horizen.utils.Pair;
-import com.typesafe.config.Config;
 
 public class SimpleAppModule extends SidechainAppModule
 {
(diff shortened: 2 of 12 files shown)


sdk/src/main/scala/com/horizen/mailbox/PrioritizedMailbox.scala
```diff
@@ -0,0 +1,17 @@
...
+import akka.dispatch.{PriorityGenerator, UnboundedPriorityMailbox}
+import com.horizen.SidechainNodeViewHolder.ReceivableMessages
+import com.typesafe.config.Config
+import scorex.core.NodeViewHolder.ReceivableMessages
+import scorex.util.ScorexLogging
+
+class PrioritizedMailbox (settings: Settings, cfg: Config) extends UnboundedPriorityMailbox (
+  PriorityGenerator {
+    case scorex.core.NodeViewHolder.ReceivableMessages.GetDataFromCurrentView => 0 // internal calls must go first
+    case com.horizen.SidechainNodeViewHolder.ReceivableMessages.GetDataFromCurrentSidechainNodeView => 1 // api calls
```
i-Alex, line 13: There are some other messages important for API: `ApplyFunctionOnNodeView`, `ApplyBiFunctionOnNodeView` and `LocallyGeneratedSecret`. Add them here as well with priority `1`.

sdk/src/main/scala/com/horizen/mailbox/PrioritizedMailbox.scala
```diff
@@ -0,0 +1,17 @@
+package com.horizen.mailbox
+
+import akka.actor.ActorSystem.Settings
+import akka.dispatch.{PriorityGenerator, UnboundedPriorityMailbox}
+import com.horizen.SidechainNodeViewHolder.ReceivableMessages
+import com.typesafe.config.Config
+import scorex.core.NodeViewHolder.ReceivableMessages
+import scorex.util.ScorexLogging
+
+class PrioritizedMailbox (settings: Settings, cfg: Config) extends UnboundedPriorityMailbox (
```
i-Alex, line 10: Check the usage of `UnboundedPriorityMailbox` here: https://doc.akka.io/docs/akka/current/typed/mailboxes.html
It seems we need to use `UnboundedStablePriorityMailbox` instead, because we must be sure if we get multiple messages with the same priority the order of execution is FIFO, otherwise it can lead to a random unexpected behavior.

examples/simpleapp/.gitignore
```diff
@@ -9,3 +9,5 @@ src/main/resources/sc_settings_with_mc.conf
 .settings/
 .vscode/
 /src/main/MyOwnSettings/
+../Dockerfile
```
i-Alex, line 12: these 2 lines duplicates the newly added records in `.gitignore` file.
Remove them from here.

examples/simpleapp/src/main/java/com/horizen/examples/SimpleAppModule.java
```diff
@@ -24,6 +24,7 @@
 import com.horizen.transaction.TransactionSerializer;
 import com.horizen.wallet.*;
 import com.horizen.utils.Pair;
+import com.typesafe.config.Config;
```
i-Alex, line 27: can be removed.

i-Alex (changes requested):

i-Alex:
Replaced by https://github.com/HorizenOfficial/Sidechains-SDK/pull/357

PR #297, opened 2021-11-04
Fee submitter configuration

VolodymyrBabaiev:


sdk/src/main/scala/com/horizen/websocket/client/MainchainNodeChannelImpl.scala
```diff
@@ -27,6 +27,15 @@ case class SendCertificateRequestPayload(scid: String,
                                          vFieldElementCertificateField: Seq[String], // Seq of compressed FE hex strings
                                          vBitVectorCertificateField: Seq[String] // Seq of compressed bitvectors hex strings
                                         ) extends RequestPayload
+case class SendCertificateAutoFeeRequestPayload(scid: String,
```
i-Alex, line 30: Do we need 2  different classes: `SendCertificateRequestPayload` and `SendCertificateAutoFeeRequestPayload`?
I would expect, if in `SendCertificateRequestPayload` we change `fee` type from `String` to `Option[String]` it will work for us in any case. So, in case there is auto fee -> `fee == None` -> field will not be serialized to JSON.

qa/sc_cert_fee_conf.py
```diff
@@ -0,0 +1,190 @@
...
+    SCNetworkConfiguration
+from SidechainTestFramework.sc_test_framework import SidechainTestFramework
+from test_framework.util import assert_false, assert_true, start_nodes, \
+    websocket_port_by_mc_node_index
+from SidechainTestFramework.scutil import bootstrap_sidechain_nodes, \
+    start_sc_nodes, connect_sc_nodes, generate_next_blocks, generate_next_block
+from SidechainTestFramework.sc_forging_util import *
+
+"""
+Check the bootstrap feature.
```
i-Alex, line 15: First few lines of test description are outdated (header and configuration). Check them and change.

i-Alex:
Note: must be merged after https://github.com/HorizenOfficial/zend_oo/pull/207

qa/sc_cert_fee_conf.py
```diff
@@ -0,0 +1,190 @@
...
+
+Configuration: bootstrap 1 SC node and start it with genesis info extracted from a mainchain node.
+    - Mine some blocks to reach hard fork
+    - Create 2 SC node(first nod with constant fee, second nod with automatic fee computation)
+
+Test:
+    - generate MC and SC blocks to reach the end of the Withdrawal epoch 0
+    - generate one more MC and SC block accordingly and await for certificate submission to MC node mempool
```
i-Alex, line 23: Specify that cert is generated by the first node with fixed fee.
Similar below, that we disable node 1 and enable node 2 cert submission. When we generate epoch 1 cert - that we use sc node 2 with automatic fee.

i-Alex (changes requested):
Seems fine. Just a few minor things left.

sdk/src/main/scala/com/horizen/websocket/client/MainchainNodeChannelImpl.scala
VolodymyrBabaiev: Mainchain after receiving send_certificate websocket command process arguments sequently. In case we don't send fee, we also shouldn't pass vFieldElementCertificateField and  vBitVectorCertificateField, because the information in these fields will be processed as fee by Mainchain.

i-Alex: as agreed, some changes on MC side need to be done. Then we will update SDK code.

i-Alex (approved):

albertog78 (approved):

PR #302, opened 2021-11-25
CSW

i-Alex:
sc-cryptolib branch: comm_tree_update


sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
```diff
@@ -22,6 +27,32 @@ public FieldElement getUtxoMerkleTreeLeaf(Box<Proposition> box) {
         return utxo.getNullifier();
     }
 
+    @Override
+    public byte[] transformPrivateKey25519(PrivateKey25519 pk) {
```
DanieleDiBenedetto, line 31: The function that performs this transformation seems to be public in Bouncy Castle:

https://github.com/bcgit/bc-java/blob/1ef1da015e116c1c2d09719674dfdd437971077f/core/src/main/java/org/bouncycastle/math/ec/rfc8032/Ed25519.java#L334

Called by:

https://github.com/bcgit/bc-java/blob/1ef1da015e116c1c2d09719674dfdd437971077f/core/src/main/java/org/bouncycastle/crypto/params/Ed25519PrivateKeyParameters.java#L61

Perhaps you can use directly that one instead ?

sdk/src/test/scala/com/horizen/csw/CswProofTest.scala
```diff
@@ -0,0 +1,23 @@
...
+  def transformPrivateKey25519(): Unit = {
+    val pk: PrivateKey25519 = PrivateKey25519Creator.getInstance().generateSecret("seed".getBytes)
+
+    val res = cswCircuit.transformPrivateKey25519(pk)
+
+    // Regression
+    assertEquals("Different pub key", "f165e1e5f7c290e52f2edef3fbab60cbae74bfd3274f8e5ee1de3345c954a166", BytesUtils.toHexString(pk.publicImage().pubKeyBytes()))
+    assertEquals("Different private key", "19b25856e1c150ca834cffc8b59b23adbd0ec0389e58eb22b3b64768098d002b", BytesUtils.toHexString(pk.privateKey()))
+    assertEquals("Different transformation result", "08eb1969be10581600c812f6ef0eea3b16c432854588698d78696e8ff7a7d163", BytesUtils.toHexString(res))
+  }
```
DanieleDiBenedetto, line 22: Can we add also a test case in which we take the transformedPrivateKey, apply the remaining part of the PublicKey generation process, and assert that the result would be equal to the one returned by the high-level APIs exposed by BouncyCastle ?

i-Alex: done

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
i-Alex: we can't use Bouncy Castle directly, because the methods are private.
I've added tests against original solution using reflection.

sdk/src/main/scala/com/horizen/api/http/SidechainCswApiRoute.scala
```diff
@@ -0,0 +1,163 @@
...
+   * return the csw info for given BoxId
+   */
+  def cswInfo: Route = (post & path("cswInfo")) {
+     entity(as[ReqCswInfo]) { body =>
+       Try {
+         Await.result(cswManager ? GetCswInfo(BytesUtils.fromHexString(body.boxId)), timeout.duration).asInstanceOf[CswInfo]
+       } match {
+         case Success(cswInfo: CswInfo) =>
+           ApiResponseUtil.toResponse(RespCswInfo(cswInfo))
+         case Failure(e) =>
```
i-Alex, line 80: We should separate Failure results from the CswManager from `Await` unexpected failures.
Same for `nullifier` endpoint.

sdk/src/main/scala/com/horizen/api/http/SidechainCswApiRoute.scala
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  /**
+   * return the csw info for given BoxId
+   */
+  def cswInfo: Route = (post & path("cswInfo")) {
+     entity(as[ReqCswInfo]) { body =>
+       Try {
+         Await.result(cswManager ? GetCswInfo(BytesUtils.fromHexString(body.boxId)), timeout.duration).asInstanceOf[CswInfo]
```
i-Alex, line 76: `.asInstanceOf[Try[CswInfo]]`

sdk/src/main/scala/com/horizen/api/http/SidechainCswApiRoute.scala
```diff
@@ -0,0 +1,163 @@
...
+    }
+  }
+
+  /**
+   * Return the nullifier for the given coin box id.
+   */
+  def nullifier: Route = (post & path("nullifier")) {
+    entity(as[ReqNullifier]) { body =>
+      Try {
+        Await.result(cswManager ? GetBoxNullifier(BytesUtils.fromHexString(body.boxId)), timeout.duration).asInstanceOf[Array[Byte]]
```
i-Alex, line 110: `.asInstanceOf[Try[Array[Byte]]]`

sdk/src/main/scala/com/horizen/api/http/SidechainCswApiRoute.scala
```diff
@@ -0,0 +1,163 @@
...
+  private[api] case class RespCswHasCeasedState(state: Boolean) extends SuccessResponse
+
+  @JsonView(Array(classOf[Views.Default]))
+  private[api] case class ReqGenerationCswState(boxId: String, mcAddress: String) {
+    require(boxId.length == 64, s"Invalid id $boxId. Id length must be 64")
+    require(mcAddress.length == 64, s"Invalid address $mcAddress. Address length must be 64")
+  }
+
+  @JsonView(Array(classOf[Views.Default]))
+  private[api] case class RespGenerationCswState(state: String) extends SuccessResponse
```
i-Alex, line 133: Let's change it with `state: String, description: String`.
Current answers like "Sidechain is alive" is a `description`, "SidechainIsAlive" is a `state`.
It will allow both users to read and programs to parse the state.

sdk/src/main/java/com/horizen/box/data/NoncedBoxData.java
```diff
@@ -1,10 +1,9 @@
 package com.horizen.box.data;
 
-import com.horizen.box.NoncedBox;
 import com.horizen.proposition.Proposition;
 import scorex.core.serialization.BytesSerializable;
 
-public interface NoncedBoxData<P extends Proposition, B extends NoncedBox<P>> extends BytesSerializable {
+public interface NoncedBoxData<P extends Proposition, B extends com.horizen.box.Box<P>> extends BytesSerializable {
```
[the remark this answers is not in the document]
i-Alex: Note: I will do NoncedBoxData -> BoxData, AbstractNoncedBoxData -> AbstractBoxData plus additional changes in a separate PR to avoid to many CSW non-related changes.

sdk/src/main/java/com/horizen/box/data/NoncedBoxData.java
```diff
@@ -1,10 +1,10 @@
 package com.horizen.box.data;
 
-import com.horizen.box.NoncedBox;
+import com.horizen.box.Box;
 import com.horizen.proposition.Proposition;
 import scorex.core.serialization.BytesSerializable;
 
-public interface NoncedBoxData<P extends Proposition, B extends NoncedBox<P>> extends BytesSerializable {
+public interface NoncedBoxData<P extends Proposition, B extends Box<P>> extends BytesSerializable {
```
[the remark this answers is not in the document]
i-Alex: Correct. I decided to put this changes into separate PR to prevent the current one growing. https://github.com/HorizenOfficial/Sidechains-SDK/pull/309

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -99,12 +110,26 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+  def hasCeased: Boolean = stateStorage.hasCeased
+
+  def certificate(referencedWithdrawalEpoch: Int): Option[WithdrawalEpochCertificate] = {
+    stateStorage.getTopQualityCertificate(referencedWithdrawalEpoch)
   }
 
-  def certificateTopQuality(epoch: Int): Long = {
-    stateStorage.getTopQualityCertificate(epoch) match {
+  def certificateTopQuality(referencedWithdrawalEpoch: Int): Long = {
+    stateStorage.getTopQualityCertificate(referencedWithdrawalEpoch) match {
```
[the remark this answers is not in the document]
i-Alex: Done

qa/sc_csw_ceased_at_epoch_1.py
```diff
@@ -0,0 +1,264 @@
...
+                     "FT spending Tx expected to be in the SC node mempool.")
+
+        # Generate 1 SC block to include Tx.
+        generate_next_blocks(sc_node, "first node", 1)[0]
+
+        # Check new box (utxo) id
+        zen_boxes_req = {"boxTypeClass": "ZenBox", "excludeBoxIds": [ft_box_2["id"]]}
+        utxo_box_1 = sc_node.wallet_allBoxes(json.dumps(zen_boxes_req))["result"]["boxes"][0]
+
+        # Generate 7 more MC blocks to finish the first withdrawal epoch, then generate 1 more SC block to sync with MC.
```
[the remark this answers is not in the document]
i-Alex: Done

alsala (approved):

VolodymyrBabaiev (approved):

sdk/src/main/java/com/horizen/cryptolibprovider/CommonCircuit.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class CommonCircuit {
+    // Note: supportedSegmentSize should correlate with the snark circuit complexity,
+    // but is always less or equal the one defined in the MC network (maxSegmentSize).
+    private static final int maxSegmentSize = (1 << 18);
+    private static final int supportedSegmentSize = (1 << 18);
```
i-Alex, line 15: Note: previously the `supportedSegmentSize` was `1 << 17`.
`1 << 18` is needed for CSW circuit, because with previous value CSW PK + VK was out of allowed size.

The drawback for the Cert circuit exists. Previously creation time was 9711ms, verification - 310ms.
Now: creation - 13850ms, verification - 689ms

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
```diff
@@ -0,0 +1,217 @@
...
+
+    @Override
+    public int utxoMerkleTreeHeight() {
+        return Constants.SC_MST_HEIGHT();
+    }
+
+    @Override
+    public FieldElement getUtxoMerkleTreeLeaf(Box<Proposition> box) {
+        ScUtxoOutput utxo = new ScUtxoOutput(box.proposition().bytes(), box.value(), box.nonce(), box.customFieldsHash());
+        return utxo.getNullifier();
```
albertog78, line 41: Even if internally in the current implementation getNullifier calls getHash, wouldn't be more appropriate to directly call getHash here? The tree leaves should represent and id whose pre-image should be used in the proving part. Nullifier could be also unique but constructed in a different way.

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public byte[] getCertDataHash(WithdrawalEpochCertificate cert) {
+        WithdrawalCertificate we = createWithdrawalCertificate(cert);
+
+        FieldElement hashFe = we.getHash();
+        byte[] hashBytes = hashFe.serializeFieldElement();
```
albertog78, line 49: missing a try finally for hashFe.freeFieldElement(); ?

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
```diff
@@ -0,0 +1,217 @@
...
+        try {
+            we.close();
+        } catch (Exception ignored) {
+        }
+
+        return hashBytes;
+    }
+
+    @Override
+    public byte[] privateKey25519ToScalar(PrivateKey25519 pk) {
```
albertog78, line 61: REACHED THIS POINT

albertog78 (changes requested):

sdk/src/main/java/com/horizen/cryptolibprovider/CommonCircuit.java
i-Alex: Yes. Can be fixed with https://github.com/HorizenOfficial/zendoo-sc-cryptolib/pull/77#pullrequestreview-843267005

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuit.java
```diff
@@ -0,0 +1,52 @@
...
+    int rangeSize(int withdrawalEpochLength);
+
+    boolean generateCoboundaryMarlinSnarkKeys(int withdrawalEpochLen, String provingKeyPath, String verificationKeyPath);
+
+    byte[] privateKey25519ToScalar(PrivateKey25519 pk);
+
+    byte[] utxoCreateProof(UtxoCswData utxo,
+                           WithdrawalEpochCertificate lastActiveCert,
+                           byte[] mcbScTxsCumComEnd,
+                           byte[] senderPubKeyHash,
```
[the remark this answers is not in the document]
i-Alex: Yes and no, as for me. That is the person who can spend the CSW coin in the MC as a CswInput. Because CSW is an input, it is more reasonable to call it "sender". Though  in design doc and sc-cryptolib it is called "receiver".  I can change to receiver, NP.

sdk/src/main/scala/com/horizen/utils/CswData.scala
```diff
@@ -0,0 +1,206 @@
...
+        amount == other.amount && nonce == other.nonce && customHash.sameElements(other.customHash) &&
+        utxoMerklePath.sameElements(other.utxoMerklePath)
+      case _ => false
+    }
+  }
+
+  override def toString = s"UtxoCswData(boxID = ${BytesUtils.toHexString(boxId)})"
+}
+
+case class ForwardTransferCswData(boxId: Array[Byte],
```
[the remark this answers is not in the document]
i-Alex: boxId is used as a key to store/access the CswData entry and to link it with the original box to be withdrawn when SC has ceased.

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
i-Alex: Ok. Would be nice to change method name to `getHash` in the sc-cryptolib as well.

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
i-Alex: Seems that `freeFieldElement` doesn't throw exceptions. It does nothing in case was freed before

sdk/src/main/java/com/horizen/cryptolibprovider/InMemorySparseMerkleTreeWrapper.java
```diff
@@ -0,0 +1,131 @@
...
+    // Returns the root of the merkle tree or null if tree state was not finalized
+    public byte[] calculateRoot() {
+        try {
+            merkleTree.finalizeInPlace();
+            FieldElement root = merkleTree.root();
+            byte[] rootBytes = root.serializeFieldElement();
+            root.freeFieldElement();
+            return rootBytes;
+        } catch (Exception e) {
+            return null;
```
albertog78, line 110: log exceptions

sdk/src/main/scala/com/horizen/SidechainWallet.scala
```diff
@@ -146,18 +154,73 @@ class SidechainWallet private[horizen] (seed: Array[Byte],
     // We keep forger boxes separate to manage forging stake delegation
     forgingBoxesInfoStorage.updateForgerBoxes(new ByteArrayWrapper(version), newDelegatedForgerBoxes, boxIdsToRemove).get
 
+    // In case utxoMerkleTreeViewOpt is defined, calculate and store the CSW data for every coin box in the Wallet
+    val utxoCswData: Seq[CswData] = utxoMerkleTreeViewOpt.map(view => calculateUtxoCswData(view)).getOrElse(Seq())
```
albertog78, line 158: REACHED THIS POINT

albertog78 (changes requested):

sdk/src/main/java/com/horizen/cryptolibprovider/CswCircuitImplZendoo.java
albertog78: the additional try finally was to ensure that hashFe.freeFieldElement is always executed even in case of any kind of error (including RuntimeExceptions) of hashFe.serializeFieldElement()

sdk/src/main/scala/com/horizen/csw/CswManager.scala
```diff
@@ -0,0 +1,520 @@
...
+  private def onGenerateCswProof: Receive = {
+    case GenerateCswProof(boxId: Array[Byte], senderAddress: String) =>
+      if (!hasSidechainCeased) {
+        sender() ! SidechainIsAlive
+      } else if (!isValidSenderAddress(senderAddress)) {
+        sender() ! InvalidAddress
+      } else {
+        val proofInfo: CswProofInfo = getProofInfo(boxId)
+        // Note: CSW proof result depends on the MC sender
+        val isSenderKnown = proofInfo.senderAddress.contains(senderAddress)
```
albertog78, line 154: isSameSender?

sdk/src/main/scala/com/horizen/storage/SidechainStateUtxoMerkleTreeStorage.scala
```diff
@@ -0,0 +1,139 @@
...
+  require(storage != null, "Storage must be NOT NULL.")
+
+  private def loadMerkleTree(): InMemorySparseMerkleTreeWrapper = {
+    val treeHeight: Int = CryptoLibProvider.cswCircuitFunctions.utxoMerkleTreeHeight()
+    val merkleTree = new InMemorySparseMerkleTreeWrapper(treeHeight)
+
+    val newLeaves: Map[java.lang.Long, FieldElement] = getAllLeavesInfo.map(leafInfo => {
+      long2Long(leafInfo.position) -> FieldElement.deserialize(leafInfo.leaf)
+    }).toMap
+    merkleTree.addLeaves(newLeaves.asJava)
```
albertog78, line 34: try finally to avoid memory leaks

albertog78 (changes requested):

albertog78 (approved):

PR #304, opened 2021-11-29
As/log improvements

alsala:
Added log4j2 logging system to SimpleApp as well as log file handling.
Removed noisy warnings/pkg info from stdout; chenged log file handling for py tests using automatic name generation

### DIFF
--- a/examples/simpleapp/pom.xml
+++ b/examples/simpleapp/pom.xml
@@ -17,27 +17,6 @@
 			<groupId>io.horizen</groupId>
 			<artifactId>sidechains-sdk</artifactId>
 			<version>0.2.7</version>
-
-			<!-- exclude the default simple-sl4j library and include the bridge to log4j -->
-			<exclusions>
-				<exclusion>
-					<groupId>org.slf4j</groupId>
-					<artifactId>slf4j-simple</artifactId>
-				</exclusion>
-			</exclusions>
-
-		</dependency>
-
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
-			<artifactId>log4j-slf4j-impl</artifactId>
-			<version>2.14.0</version>
-		</dependency>
-
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
-			<artifactId>log4j-core</artifactId>
-			<version>2.14.0</version>
 		</dependency>
 
 		<dependency>
--- a/examples/simpleapp/pom.xml
+++ b/examples/simpleapp/pom.xml
@@ -17,6 +17,27 @@
 			<groupId>io.horizen</groupId>
 			<artifactId>sidechains-sdk</artifactId>
 			<version>0.2.7</version>
+
+			<!-- exclude the default simple-sl4j library and include the bridge to log4j -->
+			<exclusions>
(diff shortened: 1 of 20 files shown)


examples/simpleapp/pom.xml
```diff
@@ -17,6 +17,27 @@
 			<groupId>io.horizen</groupId>
 			<artifactId>sidechains-sdk</artifactId>
 			<version>0.2.7</version>
+
+			<!-- exclude the default simple-sl4j library and include the bridge to log4j -->
+			<exclusions>
```
i-Alex, line 22: Instead of excluding the dependency of `sidechains-sdk` and replacing it with another one, we may directly do it in core sdk pom.
So, other developers will have it correct of out of the box.

examples/simpleapp/src/main/java/com/horizen/examples/SimpleApp.java
```diff
@@ -19,13 +23,33 @@ public static void main(String[] args) {
             System.out.println("File on path " + args[0] + " doesn't exist");
             return;
         }
-
         String settingsFileName = args[0];
 
+        // default name for the log file
+        String logFileName = "logs/debugLog.txt";
```
i-Alex, line 29: Let's try to use the directory path specified in a configuration file: `scorex { logDir = "..." }`

examples/simpleapp/pom.xml
alsala: done

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -67,6 +67,7 @@ class SidechainApp @Inject()
   )
   extends Application  with ScorexLogging
 {
+  // cannot replace ScorexLogging with LazyLogging since Application already inherits from ScorexLogging
```
i-Alex, line 70: Good idea to keep logger initialization in Settings as soon as we've read the docs.
This comment can be removed.

sdk/src/main/resources/sidechain-sdk-settings.conf
```diff
@@ -2,6 +2,13 @@ scorex {
   dataDir = /tmp/scorex/data/blockchain
   logDir = /tmp/scorex/data/log
 
+  logInfo {
```
i-Alex, line 5: add the same block to `settings_basic.conf` and `sc_settings.conf`
even if it is a duplication with default values, but very useful to know for developers/users, because we refer to this files in our internal docs. So, they will see the these configs.

i-Alex (commented):
Great!

sdk/src/main/scala/com/horizen/SidechainApp.scala
alsala: Done

sdk/src/main/resources/sidechain-sdk-settings.conf
alsala: Done

i-Alex (approved):

VolodymyrBabaiev (approved):

albertog78 (approved):

PR #312, opened 2021-12-17
STF switched to Python 3

i-Alex:


qa/SidechainTestFramework/websocket_client.py
```diff
@@ -26,7 +26,7 @@ def create_connection(self, url):
     def sendMessage(self, ws, msgType, requestId, requestType, requestPayload):
         ws.send('{"msgType":'+str(msgType)+', "requestId":'+str(requestId)+', "requestType":'+str(requestType)+', "requestPayload":'+requestPayload+' }')
         timeout = 500
-        for x in xrange(1, timeout):
+        for x in range(1, timeout):
```
VolodymyrBabaiev, line 29: I think qa/test_framework/util.py should also be modified this way:
413:  for _ in xrange(1, timeout):

i-Alex: done

VolodymyrBabaiev (approved):

PR #314, opened 2021-12-20
As/issue 301

alsala:
Substitution of IODBStoreAdapter with VersionedLevelDbStorageAdapter, a LevelDB based storage


i-Alex (commented):
Seems fine.
Now, I'm thinking about the removal of iodb dependency from pom files.
But `io.iohk.iodb.ByteArrayWrapper` prevents us.
Let's remove it as well and implement missing parts in the descendant class: `com.horizen.utils.ByteArrayWrapper` written by us.
Initially we were forced to use iodb wrapper because of IODB storage.
To be discussed.

i-Alex (approved):

albertog78 (approved):

PR #344, opened 2022-03-24
Fix for dlog key cryptolib initialization in bootstrapping tool

alsala:
This is necessary while using the tool in interactive mode since initialization of dlog key must be done only once by the rust cryptolib. Such data is stored in memory and is used in both generateCertProofInfo and generateCswProofInfo cmds.


saratnt (approved):

tools/sctool/src/main/java/com/horizen/ScBootstrappingTool.java
```diff
@@ -1,10 +1,15 @@
 package com.horizen;
 import java.io.File;
 import java.util.Scanner;
+import java.util.Set;
+import java.util.TreeSet;
+
+import com.horizen.cryptolibprovider.CryptoLibProvider;
```
VolodymyrBabaiev, line 7: These imports are unused. Can be removed.

i-Alex: agree. Revert the file to previous version

i-Alex (changes requested):

tools/sctool/src/main/java/com/horizen/ScBootstrappingTool.java
alsala: ok, done

VolodymyrBabaiev (approved):

i-Alex (approved):

Issue #310, opened 2021-12-13
Allow `SidechainApp.stopAll()` extension


i-Alex:
Would be useful to allow custom app developers to extend `stopAll` method. For example, to close custom storage databases in a safe way when App is going to stop. To be discussed.

Issue #193, opened 2020-09-17
Sidechain Node safe stop


i-Alex:
We need to add a safe way to stop the node execution, similar to how it's done in the zend.
Currently, we are able only to terminate the process, that may lead to the Node persistent data corruption/inconsistency.
Let's add the API call `stop` method, which will do it safely.

PR #347, opened 2022-03-30
Added http api for stopping the SC node and obj hook for calling custom application stop procedure

alsala:
Addresses issue #193 and #310

### DIFF
--- a/sdk/src/main/scala/com/horizen/SidechainApp.scala
+++ b/sdk/src/main/scala/com/horizen/SidechainApp.scala
@@ -68,6 +68,7 @@ class SidechainApp @Inject()
    @Named("ConsensusStorage") val consensusStorage: Storage,
    @Named("CustomApiGroups") val customApiGroups: JList[ApplicationApiGroup],
    @Named("RejectedApiPaths") val rejectedApiPaths : JList[Pair[String, String]],
+   @Named("ApplicationStopper") val applicationStopper : SidechainAppStopper
   )
   extends Application  with ScorexLogging
 {
@@ -270,8 +271,7 @@ class SidechainApp @Inject()
     timeProvider,
     applicationWallet,
     applicationState,
-    genesisBlock,
-    this
+    genesisBlock
     ) // TO DO: why not to put genesisBlock as a part of params? REVIEW Params structure
 
   def modifierSerializers: Map[ModifierTypeId, ScorexSerializer[_ <: NodeViewModifier]] =
@@ -336,7 +336,7 @@ class SidechainApp @Inject()
   var coreApiRoutes: Seq[SidechainApiRoute] = Seq[SidechainApiRoute](
     MainchainBlockApiRoute(settings.restApi, nodeViewHolderRef),
     SidechainBlockApiRoute(settings.restApi, nodeViewHolderRef, sidechainBlockActorRef, sidechainBlockForgerActorRef),
-    SidechainNodeApiRoute(peerManagerRef, networkControllerRef, timeProvider, settings.restApi, nodeViewHolderRef),
+    SidechainNodeApiRoute(peerManagerRef, networkControllerRef, timeProvider, settings.restApi, nodeViewHolderRef, this),
     SidechainTransactionApiRoute(settings.restApi, nodeViewHolderRef, sidechainTransactionActorRef, sidechainTransactionsCompanion, params),
     SidechainWalletApiRoute(settings.restApi, nodeViewHolderRef),
     SidechainSubmitterApiRoute(settings.restApi, certificateSubmitterRef, nodeViewHolderRef),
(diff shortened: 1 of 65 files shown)


sdk/src/main/scala/com/horizen/api/http/SidechainBlockActor.scala
```diff
@@ -35,6 +35,14 @@ class SidechainBlockActor[PMOD <: PersistentNodeViewModifier, SI <: SidechainSyn
     context.system.eventStream.subscribe(self, classOf[ChangedHistory[HR]])
   }
 
+  override def postStop(): Unit = {
+    log.info("SidechainBlock Actor is stopping...")
+    context.system.eventStream.unsubscribe(self, classOf[SemanticallyFailedModification[PMOD]])
```
i-Alex, line 40: do we need to unsubscribe to avoid `dead letters` messages?

sdk/src/main/scala/com/horizen/api/http/SidechainNodeApiRoute.scala
```diff
@@ -114,6 +116,26 @@ case class SidechainNodeApiRoute(peerManager: ActorRef,
...
   }
 
+  def stop: Route = (post & path("stop")) {
+    try {
+      // we are stopping the node, and since we will shortly be closing network services lets do in a separate thread
+      // and give some time to the HTTP reply to be transmitted immediately in this thread
+      new Thread( new Runnable() {
+        override def run(): Unit = {
+          log.info("Sleeping 50 msec...")
+          sleep(50)
```
i-Alex, line 126: 50 seconds can be not enough on weak PCs. like on our CI. So can lead to random failures when of API.
I'm curious to know if AKKA http has something like "onComplete" event that can be used for a particular request. If yes we may try to use it and then execute `stopAll`.
Let's check it.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/api/http/SidechainBlockActor.scala
alsala: That was I was trying to avoid, but it seems not the case. I think is anyway cleaner unregistering from what we registered in the start phase, for this reason I kept it.

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -356,7 +357,13 @@ class SidechainApp @Inject()
   override val swaggerConfig: String = Source.fromResource("api/sidechainApi.yaml")(Codec.UTF8).getLines.mkString("\n")
 
   override def stopAll(): Unit = {
+    log.info("Calling custom application stopAll...")
+    applicationStopper.stopAll()
```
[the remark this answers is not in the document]
alsala: As discussed I reimplemented the base class stopAll() in order to customize the behaviour.

sdk/src/main/scala/com/horizen/SidechainAppStopper.scala
```diff
@@ -0,0 +1,8 @@
+package com.horizen
+
+
+abstract class SidechainAppStopper {
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/api/http/SidechainBlockActor.scala
alsala: I removed all unnecessary unsubscribe() calls

sdk/src/main/scala/com/horizen/api/http/SidechainNodeApiRoute.scala
alsala: I found no alternative, for the time being the number of millisec has been increased to 500

sdk/src/main/scala/com/horizen/api/http/SidechainNodeApiRoute.scala
```diff
@@ -114,6 +116,34 @@ case class SidechainNodeApiRoute(peerManager: ActorRef,
     }
   }
 
+  def stop: Route = (post & path("stop")) {
+    if (app.stopAllInProgress.get() ) {
```
saratnt, line 120: If the stop method is called twice at the same time, it is possible that both threads can fork the stop thread. Maybe it is better to use app.stopAllInProgress.compareAndSet, that it is an atomic operation and so only 1 thread can fork the stop thread.

i-Alex (approved):

sdk/src/main/scala/com/horizen/api/http/SidechainNodeApiRoute.scala
alsala: Ok, done

saratnt (approved):

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -68,6 +72,7 @@ class SidechainApp @Inject()
    @Named("ConsensusStorage") val consensusStorage: Storage,
    @Named("CustomApiGroups") val customApiGroups: JList[ApplicationApiGroup],
    @Named("RejectedApiPaths") val rejectedApiPaths : JList[Pair[String, String]],
+   @Named("ApplicationStopper") val applicationStopper : SidechainAppStopper
```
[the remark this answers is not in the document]
alsala: Done on branch `sdk_031_features`

VolodymyrBabaiev (approved):

PR #357, opened 2022-04-29
Api node synchronization

VolodymyrBabaiev:
Fixed issue of api freezing during node synchronization.
In ModifiersCache was fixed issue of cache cleaning.


paolocappelletti (approved):

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
```diff
@@ -302,6 +348,7 @@ object SidechainNodeViewHolder /*extends ScorexLogging with ScorexEncoding*/ {
     case class ApplyFunctionOnNodeView[HIS, MS, VL, MP, A](f: java.util.function.Function[SidechainNodeView, A])
     case class ApplyBiFunctionOnNodeView[HIS, MS, VL, MP, T, A](f: java.util.function.BiFunction[SidechainNodeView, T, A], functionParameter: T)
     case class LocallyGeneratedSecret[S <: SidechainTypes#SCS](secret: S)
+    case class ApplyModifier(applied: Seq[SidechainBlock])
```
i-Alex, line 351: let's make this one as a part of internal interface, because we don't want anyone outside to use this message. See `InternalReceivableMessages` in `CertificateSubmitter` object as an example.

sdk/src/test/scala/com/horizen/SidechainNodeViewHolderTest.scala
```diff
@@ -42,6 +47,7 @@ class SidechainNodeViewHolderTest extends JUnitSuite
 
   val genesisBlock: SidechainBlock = SidechainBlockFixture.generateSidechainBlock(sidechainTransactionsCompanion)
   val params: NetworkParams = RegTestParams(initialCumulativeCommTreeHash = FieldElementFixture.generateFieldElement())
+  implicit lazy val timeout: Timeout = Timeout(100000 milliseconds)
```
i-Alex, line 50: do we need it to be 100s? In case of any bug here, we may wait 100s to fail the test.
Let's have 10s, we still will be safe even when running this test on the weak PC.

sdk/src/test/scala/com/horizen/SidechainNodeViewHolderTest.scala
```diff
@@ -448,4 +455,225 @@ class SidechainNodeViewHolderTest extends JUnitSuite
...
+    val eventListener = TestProbe()
+    actorSystem.eventStream.subscribe(eventListener.ref, classOf[ModifiersProcessingResult[(Seq[SidechainBlock], Seq[SidechainBlock])]])
+
+    mockedNodeViewHolderRef ! ModifiersFromRemote(firstRequestBlocks)
+    mockedNodeViewHolderRef ! ModifiersFromRemote(secondRequestBlocks)
+
+    eventListener.fishForMessage(timeout.duration) {
+      case m =>
+        m match {
+          case ModifiersProcessingResult(applied, _) => {
```
i-Alex, line 555: same here and in all the places below, check that cleared is empty.

sdk/src/test/scala/com/horizen/SidechainNodeViewHolderTest.scala
```diff
@@ -448,4 +455,225 @@ class SidechainNodeViewHolderTest extends JUnitSuite
...
+            assertTrue("Applyed block sequence is differ", applied.toSet.equals(Set(block3, block4, block5, block6)))
+            true
+          }
+          case _ => false
+        }
+    }
+  }
+
+  @Test
+  def remoteModifiersCacheClean(): Unit = {
```
i-Alex, line 641: let's add test description as was did for the tests above. It is very helpful!

sdk/src/test/scala/com/horizen/utils/SDKModifiersCacheTest.scala
```diff
@@ -0,0 +1,62 @@
...
+    val modifiersCache = new SDKModifiersCache[SidechainBlock, HIS](cacheSize)
+    val blocksNumber = 200
+    var blocks = generateSidechainBlockSeq(blocksNumber, sidechainTransactionsCompanion, params, Some(genesisBlock.id))
+
+    blocks.foreach(block => modifiersCache.put(block.id, block))
+    var expectedChacheSize = 200
+    assertEquals("Cache size is differ", expectedChacheSize, modifiersCache.size)
+
+    blocks = generateSidechainBlockSeq(blocksNumber, sidechainTransactionsCompanion, params, Some(blocks.last.id))
+    blocks.takeRight(200).foreach(block => modifiersCache.put(block.id, block))
```
i-Alex, line 48: `.takeRight(200)` can be removed.

i-Alex (changes requested):
The changes of the PR seems completely fine.
I put changes notes to cleanup a few points and add some comments.
PR requires the merge with dev. After it, please, check the behavior of `sc_cert_submitter_after_sync_1.py` and `sc_cert_submitter_after_sync_2.py` tests: I don't expect to see any "error" log there during the node synchronization.

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
VolodymyrBabaiev: Cannot make it private because ApplyModifier is also used in com.horizen.mailbox.PrioritizedMailbox.

qa/sc_node_response_along_sync.py
```diff
@@ -0,0 +1,135 @@
...
+def print_peers_connections(sc_nodes):
+    print("\n")
+    for i in range(len(sc_nodes)):
+        print("Peer #" + str(i))
+        print("node " + "127.0.0.1:" + str(sc_p2p_port(i)) + " peers:")
+        print_peers_addresses(sc_nodes[i])
+    return ""
+
+
+def sc_create_multiple_not_connected_nodes_network(mc_node_1, num_of_nodes_to_start, mc_node_index):
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Yes, it's a relic. SCMultiNetworkConfiguration was removed and used SCNetworkConfiguration instead.

sdk/src/main/scala/com/horizen/utils/SDKModifiersCache.scala
```diff
@@ -0,0 +1,25 @@
...
+ * @tparam PMOD - type of a persistent node view modifier (or a family of modifiers).
+ */
+
+class SDKModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD, _]] (override val maxSize: Int)
+  extends DefaultModifiersCache[PMOD, H] (maxSize) {
+  @Override
+  override def put(key: K, value: V): Unit = {
+    if (!contains(key)) {
+      cache.put(key, value)
+      onPut(key)
```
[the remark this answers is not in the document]
VolodymyrBabaiev: Done.

sdk/src/test/scala/com/horizen/SidechainNodeViewHolderTest.scala
VolodymyrBabaiev: Done.

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
VolodymyrBabaiev: Visibility was restricted to com.horizen package.

Issue #184, opened 2020-09-07
wallet storage not updated correctly with box having custom proposition


paolocappelletti:
**Issue:**
the API endpoint /allBoxes does not show boxes owned by the user that have custom proposition.
(for example CarSellOrderBox  in the LamboRegistry app)

**Cause:**
SidechainWallet  uses this code to check if a box can be opened by the user:
 pubKeys.contains(box.proposition())
(see: [here](https://github.com/HorizenOfficial/Sidechains-SDK/blob/master/sdk/src/main/scala/com/horizen/SidechainWallet.scala#L121) )
This relies on the the .equals() method of the Proposition class, but this can cause a failing check if the box has a custom proposition that overrides the equals method (as in the case of the [SellOrderProposition  ](https://github.com/HorizenOfficial/lambo-registry/blob/master/src/main/java/io/horizen/lambo/car/proposition/SellOrderProposition.java#L95)of CarSellOrderBox  in the lamboregistry)

**Proposed solution:**
exend the Proposition interface with a new method 
`public boolean matches(Proposition otherProposition)`
and modifiy the SidechainWallet  to use it  for the box-owning check.

i-Alex:
Thanks! I agree with the proposal.

i-Alex:
The proposed changes are not fully enough for us. We should review the way how we store secrets and retrieve secrets for given proposition. The corner case is shown by the SellOrderProposition, which is actually a complex Proposition, that can have multiple related Secrets.

paolocappelletti:
this bug afftects also the  getNodeWallet().boxesOfType(...) method of SidechainNodeView : it does not  returns the boxes locked by a custom proposition

PR #364, opened 2022-05-12
Custom propositions in node wallet

paolocappelletti:


paolocappelletti:
https://github.com/HorizenOfficial/Sidechains-SDK/issues/184
https://horizenlabs.atlassian.net/browse/SCHAINS-421

MarcoOl94 (approved):

VolodymyrBabaiev (approved):

sdk/src/main/scala/com/horizen/csw/CswManager.scala
```diff
@@ -449,7 +449,7 @@ class CswManager(settings: SidechainSettings,
     }
 
     def getOwner(sidechainNodeView: View): Option[PrivateKey25519] = {
-      sidechainNodeView.vault.secretByPublicKey(publicKey25519Proposition).asScala.map(_.asInstanceOf[PrivateKey25519])
+      sidechainNodeView.vault.secretByPublicKey25519Proposition(publicKey25519Proposition).asScala.map(_.asInstanceOf[PrivateKey25519])
```
i-Alex, line 452: I think we don't need `.map(_.asInstanceOf[PrivateKey25519])` anymore

sdk/src/test/java/com/horizen/customtypes/CustomPublicKeyProposition.java
```diff
@@ -58,4 +62,15 @@ public String toString() {
     public byte[] pubKeyBytes() {
         return Arrays.copyOf(pubKeyBytes, PUBLIC_KEY_LENGTH);
     }
+
+    @Override
+    public ProvableCheckResult<CustomPrivateKey> canBeProvedBy(List<Secret> secrectList){
```
i-Alex, line 67: Can we just let CustomPublicKeyProposition extend `AbstractSingleSecretProofOfKnowledgeProposition` to avoid code duplication?

sdk/src/test/java/com/horizen/proposition/CustomPropositionTest.java
```diff
@@ -0,0 +1,47 @@
...
+import com.horizen.secret.Secret;
+import com.horizen.utils.Ed25519;
+import com.horizen.utils.Pair;
+import org.junit.Test;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import static org.junit.Assert.*;
+
+public class CustomPropositionTest {
```
i-Alex, line 14: I think, that we should add custom proposition that can be opened by k of n secrets. Like what we have in Lambo Registry for sell order, that can be opened or by the owner or by the buyer.
Then add the test similar to this one.

sdk/src/main/scala/com/horizen/api/http/SidechainTransactionApiRoute.scala
```diff
@@ -175,7 +175,9 @@ case class SidechainTransactionApiRoute(override val settings: RESTApiSettings,
       applyOnNodeView { sidechainNodeView =>
         val wallet = sidechainNodeView.getNodeWallet
         val inputBoxes = wallet.allBoxes().asScala
-          .filter(box => body.transactionInputs.exists(p => p.boxId.contentEquals(BytesUtils.toHexString(box.id()))))
+          .filter(box =>
+            (box.isInstanceOf[ZenBox] || box.isInstanceOf[ForgerBox]) &&
```
i-Alex, line 179: why do we filter with ZenBox and ForgerBox only?
Actually we can spend any box locked by PublicKey25519Proposition.

i-Alex (changes requested):
Seems good.
I think we need to extend tests a bit.

sdk/src/main/scala/com/horizen/api/http/SidechainTransactionApiRoute.scala
paolocappelletti: we are creating a core transaction here: are'nt the only two types of boxes we can have in input in this tx?

i-Alex: was agreed to keep only ZenBoxes and ForgerBoxes available.

i-Alex (approved):

PR #378, opened 2022-05-24
Refactored Wallet classe hierarchy for AccountWallet.

saratnt:


sdk/src/main/scala/com/horizen/account/wallet/AccountWallet.scala
```diff
@@ -1,53 +1,52 @@
...
-                                           secretStorage: SidechainSecretStorage,
-                                           params: NetworkParams): Try[AccountWallet] = Try {
-    //TODO: uncomment
-    // new AccountWallet(seed, secretStorage, params)
-    new AccountWallet()
+                                           secretStorage: SidechainSecretStorage
+                                          ): Try[AccountWallet] = Try {
+
+    if (secretStorage.isEmpty) {
+      new AccountWallet(seed, secretStorage)
```
i-Alex, line 47: secret storage can be empty even for long-running node.
`new AccountWallet(seed, secretStorage)` is enough for this method

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/wallet/AccountWallet.scala
saratnt: OK.

i-Alex (approved):

PR #379, opened 2022-05-25
Removed transactions limit inside a block

MarcoOl94:


lander86 (approved):

i-Alex:
Closed. We don't want this to keep this changes in the dev, only for test purposes

PR #380, opened 2022-05-25
EVM statedb

JcBernack:
- refactor libevm interface to simplify Java interop
  - this will greatly reduce development overhead when adding new methods to the library
- add StateDB class that manages an underlying instance of the go-ethereum StateDB


libevm/lib/service_state.go
```diff
@@ -0,0 +1,249 @@
...
+		if end > length {
+			end = length
+		}
+		chunk := statedb.GetState(params.Address, getChunkKey(params.Key, i+1))
+		copy(data[i:end], skipLeadingZeroes(chunk))
+	}
+	return nil, data
+}
+
+// StateSetStorage writes values of arbitrary length to the storage trie of given account
```
JcBernack, line 220: The algorithm for storing values larger than 32 bytes works but is not particularly pretty.

evm/src/main/java/com/horizen/evm/interop/EvmLog.java
```diff
@@ -0,0 +1,19 @@
+package com.horizen.evm.interop;
+
+import com.horizen.evm.utils.Address;
+import com.horizen.evm.utils.Hash;
+import com.horizen.evm.utils.HexBytes;
+
+import java.math.BigInteger;
+
+public class EvmLog {
```
JcBernack, line 9: Currently, this contains everything the log in Ethereum has to offer. Depending on how we create and store reciepts in the core we might be able to reduce this.

evm/src/main/java/com/horizen/evm/LibEvm.java
```diff
@@ -0,0 +1,154 @@
...
+    public static EvmResult evmExecute(int handle, byte[] from, byte[] to, BigInteger value, byte[] input)
+        throws Exception {
+        var cfg = new EvmParams.EvmConfig();
+        cfg.origin = Address.FromBytes(from);
+        cfg.value = value;
+        cfg.gasLimit = 100000;
+        return invoke("EvmExecute", new EvmParams(handle, cfg, to, input), EvmResult.class);
+    }
+
+    public static EvmApplyResult evmApply(int handle, byte[] from, byte[] to, BigInteger value, byte[] input)
```
JcBernack, line 145: This is an alternative method to call the EVM which uses more of the code-path in GETH when applying a block. It has more validation and also handles gas fees (Intrinsic Gas + moving gas from the caller balance to the coinbase account). Which one we're going to use, or if we need something a bit different depends on how exactly we want to handle gas fees.

evm/src/main/java/com/horizen/evm/LibEvm.java
```diff
@@ -0,0 +1,154 @@
...
+import com.horizen.evm.interop.*;
+import com.horizen.evm.utils.Address;
+import com.horizen.evm.utils.Hash;
+import com.sun.jna.Library;
+import com.sun.jna.Native;
+import com.sun.jna.Pointer;
+
+import java.math.BigInteger;
+
+final class LibEvm {
```
rockNicktive, line 13: For gas estimation, is the intention to "throw the state chages away" on the java side and simply use evmExecute?

rockNicktive (approved):
I've read through the code changes. I'll test it on my machine later, but so far everything is looking great and I don't expect any issues!

evm/src/main/java/com/horizen/evm/LibEvm.java
JcBernack: For gas estimation I expect to re-use parts or even all of the geth implementation. So I think it makes more sense to do that on the go-library side.

evm/src/main/java/com/horizen/evm/interop/EvmLog.java
JcBernack: Removed all derived fields.

i-Alex (changes requested):
Looks good.
The only point about the physical storage usage.

i-Alex (approved):

PR #383, opened 2022-05-26
mo/Schains-474

MarcoOl94:
## Description
This PR adds a new endpoint **/backup/getRestoredBoxes** used by the Explorer to retrieve the boxes restored with the backup/restore procedure (See https://github.com/HorizenOfficial/Sidechains-SDK/pull/351).

## Jira ticket
[474](https://horizenlabs.atlassian.net/jira/software/c/projects/SCHAINS/boards/3?modal=detail&selectedIssue=SCHAINS-474)

## Checks
- [✔] Project Builds
- [✔] Project passes Unit tests and checks
- [✘] Project passes Python tests
- [✔] Sidechain Explorer consistency


sdk/src/main/scala/com/horizen/backup/BoxIterator.java
```diff
@@ -54,6 +56,24 @@ public Optional<BackupBox> nextBox() throws RuntimeException {
         return nextBox(false);
     }
 
+    public List<Box<Proposition>> getNextBoxes(int nElement, Optional<byte []> keyToSeek) {
+        if (keyToSeek.isPresent()) {
+            nElement += 1;
```
i-Alex, line 61: Do we need to increment `nElement`, then add key to seek box to the resutl list and afterwords - remove that element?
Can we just do:
```
this.seekIterator(Utils.calculateKey(keyToSeek.get()).data());
this.nextBox();
```

i-Alex (changes requested):

paolocappelletti (approved):
agree on the Olek note, maybe we can make that method more readble.
apart from that looks good!

sdk/src/main/scala/com/horizen/backup/BoxIterator.java
MarcoOl94: Yes this is much better!

sdk/src/main/scala/com/horizen/backup/BoxIterator.java
```diff
@@ -54,6 +56,22 @@ public Optional<BackupBox> nextBox() throws RuntimeException {
         return nextBox(false);
     }
 
+    public List<Box<Proposition>> getNextBoxes(int nElement, Optional<byte []> keyToSeek) {
```
i-Alex, line 59: Let's add the UTs for this method with the real storage usage, where we will put real boxes (both coin and non-coin) and also some trash-records (not a boxes).

i-Alex (changes requested):

i-Alex (approved):

alsala (approved):

PR #392, opened 2022-06-03
Eth rpc

akrckn:
WIP


examples/evmapp/src/main/java/com/horizen/examples/EvmAppModule.java
```diff
@@ -42,6 +45,7 @@ public void configureApp() {
 
         // Here I can add my custom rest api and/or override existing one
         List<ApplicationApiGroup> customApiGroups = new ArrayList<>();
+        customApiGroups.add(new RpcController(new RpcHandler(new EthService())));
```
i-Alex, line 48: Let's define RPC server as a core api group instead of custom api group, since we plan it to be used for every application.

For this, we may keep `EthService()` as is, but replace `RpcController` and `RpcHandler` classes with a single scala class like we did for `AccountTransactionApiRoute`.
It requires to define a route with only one method inside. That method will parse the args of type `RpcRequest` and invoke `RpcService.execute`.

i-Alex (changes requested):

PR #397, opened 2022-06-08
Mo/sdk 47

MarcoOl94:
## Description
This PR adds some new wallet endpoints for the export/import of private keys.

## Jira Ticket
[47](https://horizenlabs.atlassian.net/jira/software/c/projects/SDK/boards/14?modal=detail&selectedIssue=SDK-47)

## Changes

- Added new endpoint **wallet/importSecret**
- Added new endpoint **wallet/exportSecret**
- Added new endpoint **wallet/dumpSecrets**
- Added new endpoint **wallet/importSecrets**

All of these endpoints require authentication. This means that the POST request should add an additional header (**api_key**) that contains a string which its hash is included in the config file.
The sc bootstrap tool was updated with a new command (**encodeString**) that calculates this hash needed inside the config file for the API authentication.

## Breaking Changes
- No breaking changes.

## Checks
- [✔] Project Builds
- [✔] Project passes tests and checks
- [✔] Breaking changes have been correctly tagged and notified


qa/SidechainTestFramework/sidechainauthproxy.py
```diff
@@ -71,7 +71,7 @@ def __getattr__(self, name):
             name = "%s.%s" % (self.__service_name, name)
         return SidechainAuthServiceProxy(self.__service_url, name, connection=self.__conn)
 
-    def _request(self, method, path, postdata):
+    def _request(self, method, path, postdata, additional_header):
```
i-Alex, line 74: I would suggest to update the constructor of the class with "api key" argument. The value is known at the place of proxy instance creation.

Then, for every call to the API automatically add `apiKey` to the header.
So, inside the particular test we don't need to care to set:
```
auth = {
         "api_key": api_key
      }
```
For the authorization checks we may keep this argument `additional_header` but with a different context - "customApiKey".
By default , method looks like `sidechainNode.wallet_exportSecret(request)`, but if you want to specify custom api key it will look like `sidechainNode.wallet_exportSecret(request, customApiKey)`.

qa/sc_import_export_keys.py
```diff
@@ -0,0 +1,252 @@
...
+from httpCalls.wallet.balance import http_wallet_balance
+from httpCalls.transaction.sendCoinsToAddress import sendCoinsToAddress
+from SidechainTestFramework.sidechainauthproxy import SCAPIException
+
+"""
+    - Setup 2 SC Node.
+    - Create a new address on node2
+    - Export its private key and import it inside the node1
+    - Verify that node1 owns this address
+    - Send coins to this address and verify that the balance on node1 is unchanged and node2 owns the new amount
```
i-Alex, line 22: I guess here: "Send coins from node1 to this address...."
That is the reason why the balance of node1 is unchanged. Correct?

qa/sc_import_export_keys.py
```diff
@@ -0,0 +1,252 @@
...
+            MCConnectionInfo(address="ws://{0}:{1}".format(mc_node_1.hostname, websocket_port_by_mc_node_index(0))),
+            True,
+            automatic_fee_computation=False,
+            api_key_hash=self.API_KEY_HASH
+        )
+        sc_node_2_configuration = SCNodeConfiguration(
+            MCConnectionInfo(address="ws://{0}:{1}".format(mc_node_1.hostname, websocket_port_by_mc_node_index(0))),
+            True,
+            automatic_fee_computation=False,
+            api_key_hash=self.API_KEY_HASH
```
i-Alex, line 60: API key and key hash can be generated automatically during node bootstrap. Use bootstrapping tool new command.

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
```diff
@@ -134,6 +138,122 @@ case class SidechainWalletApiRoute(override val settings: RESTApiSettings,
...
+    }
+  }
+
+  /**
+   * Export a private key from the wallet based on its public key
+   */
+  def exportSecret: Route = (post & path("exportSecret")) {
+    withAuth {
+      entity(as[ReqExportSecret]) { body =>
+        val proposition = PublicKey25519PropositionSerializer.getSerializer.parseBytes(BytesUtils.fromHexString(body.publickey))
```
i-Alex, line 165: User may request an export of VRF key or any custom key for example.
We need to consider:
1) new proposition <-> secret interface defined by Paolo C.
2) the case of "complex" proposition that may be related to multiple secrets. (Maybe not a case for this method, since it is all about the keys only)

A new wallet method that takes public key bytes and compares with  `secret.publicImage().publicKeyBytes()` can help.

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
```diff
@@ -134,6 +138,122 @@ case class SidechainWalletApiRoute(override val settings: RESTApiSettings,
...
+              case Success(value) =>
+                if(!BytesUtils.toHexString(value.publicImage().bytes()).equals(keyPair(1))) {
+                  log.error(s"Import Wallet: Public key doesn't match: ${BytesUtils.toHexString(value.publicImage().bytes())}  ${keyPair(1)}")
+                  error = JOptional.of(ErrorPropositionNotMatch(s"Public key doesn't match on line ${lineNumber}", JOptional.empty()))
+                } else {
+                  secrets.add((value, lineNumber))
+                }
+              case Failure(e) =>
+                log.error(s"Import Wallet: Failed to parse the secret: ${keyPair(0)}")
+                error = JOptional.of(ErrorFailedToParseSecret(s"Failed to parse the secret at line ${lineNumber}", JOptional.of(e)))
```
i-Alex, line 226: in case we have multiple errors, we track only the last one.
Let's track the whole list, or stop parsing after the first error.

i-Alex, line 226: let's extend python test, where we corrupt the file with keys (append a random line to the file) and check that we track error as expected during the import.

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
```diff
@@ -8,26 +8,30 @@ import com.horizen.SidechainNodeViewHolder.ReceivableMessages
...
 
 case class SidechainWalletApiRoute(override val settings: RESTApiSettings,
-                                   sidechainNodeViewHolderRef: ActorRef)(implicit val context: ActorRefFactory, override val ec: ExecutionContext)
+                                   sidechainNodeViewHolderRef: ActorRef,
+                                   sidechainSecretsCompanion: SidechainSecretsCompanion)(implicit val context: ActorRefFactory, override val ec: ExecutionContext)
   extends SidechainApiRoute {
 
   override val route: Route = pathPrefix("wallet") {
-    allBoxes ~ coinsBalance ~ balanceOfType ~ createPrivateKey25519 ~ createVrfSecret ~ allPublicKeys
+    allBoxes ~ coinsBalance ~ balanceOfType ~ createPrivateKey25519 ~ createVrfSecret ~ allPublicKeys ~ importSecret  ~ exportSecret ~ dumpSecrets ~ importSecrets
```
i-Alex, line 34: add authorization to `createPrivateKey25519` and `createVrfSecret`  methods.
The same for `transaction` api where we spend coins or send raw tx: `createCoreTransaction ~ createCoreTransactionSimplified ~  sendCoinsToAddress ~ sendTransaction ~ withdrawCoins ~ makeForgerStake ~ spendForgingStake`

i-Alex (changes requested):
API interface looks good.
Let's improve a bit STF usage.
Note: update `sidechainApp.yaml` file with the new API endpoints info. Same for documentation.

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
```diff
@@ -66,6 +68,9 @@ public void processCommand(String input) throws IOException {
             case "generateCswProofInfo":
                 processGenerateCswProofInfo(command.data());
                 break;
+            case "encodeString":
```
paolocappelletti, line 71: do we really need this change to the bootstrapping tool?
looks like it just performs a standar Blake2b256, which should be something standard and could be generated by some online tools as well

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
paolocappelletti: agree. all these  endpoints implementations are only related to 25519 keys
we could maybe change the endpoints path to something like:
wallet/secrets/25519/import
wallet/secrets/25519/export
wallet/secrets/25519/importList
wallet/secrets/25519/exportList

or add an additional  "type" paremeter to every endpoint , and just raise an exception if type != 25519

otherwise we have to change the implementation to be more generic: for the generic  proposition->secret there is this new method I added in wallet (note that 1 proposition could be unlocked by more than 1 secret)
< S extends Secret> List< S > secretsByProposition(ProofOfKnowledgeProposition< S > proposition)

paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
MarcoOl94: I tought this could be done as a separate task, but it's fine for me to do it now.

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
MarcoOl94: Agree but i think we should give the possibility to the user to calculate it using also our tools.

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
i-Alex: I missed that point.
Honestly, "error" processing flow can be simplified just by `return ApiResponseUtil.toResponse(error)` here.
Currently there is one place where we declare optional error field, another where we check if it is not empty in the while loop, another where we setup, and one more where we check and return error result.

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
i-Alex: I would prefer to keep it as out-of-the-box utility.
Need to update the internal doc with the usage info.

MarcoOl94: Which doc are you referring to @i-Alex ?

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
MarcoOl94: Done

qa/sc_import_export_keys.py
MarcoOl94: Yes correct, I updated the comment.

qa/sc_import_export_keys.py
MarcoOl94: Done

qa/SidechainTestFramework/sidechainauthproxy.py
MarcoOl94: Done

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
MarcoOl94: Done

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
i-Alex: right, let's do it in a separate PR. Also, we need to have a separate python or unit test that verifies authentication flow.

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
MarcoOl94: Updated also the README with the ScBoostrapping tool instruction.

qa/httpCalls/wallet/dumpSecrets.py
```diff
@@ -4,10 +4,6 @@ def http_wallet_dumpSecrets(sidechainNode, file_path, api_key):
       j = {
         "path": file_path
       }
-      auth = {
-         "api_key": api_key
-      }
       request = json.dumps(j)
-      authHeader = json.dumps(auth)
-      response = sidechainNode.wallet_dumpSecrets(request, authHeader)
+      response = sidechainNode.wallet_dumpSecrets(request, api_key)
```
i-Alex, line 8: I think, the method argument "api_key" can have defalut value "None".
In case of None we call just `sidechainNode.wallet_dumpSecrets(request)`
Same in other methods like this.
This will simplify the life for the user.

sdk/src/main/java/com/horizen/node/NodeWallet.java
```diff
@@ -33,6 +33,8 @@ public interface NodeWallet {
 
     <S extends Secret> List<S> secretsByProposition(ProofOfKnowledgeProposition<S> proposition);
 
+    <S extends Secret> Optional<S> secretByProposition(byte[] proposition);
```
i-Alex, line 36: let's call it `secretByPublicKeyBytes(byte[] publicKey)`
since this method is not about all propositions, but only for the case of PrivateKey <-> PublicKey relations.
For custom proposition, for example, with multiple public keys inside (K-of-N multisig) this method has no sense.

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
i-Alex: good

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
```diff
@@ -134,6 +138,121 @@ case class SidechainWalletApiRoute(override val settings: RESTApiSettings,
...
+   * Import a private key inside the wallet
+   */
+  def importSecret: Route = (post & path("importSecret")) {
+    withAuth {
+      entity(as[ReqImportSecret]) { body =>
+        val secret = sidechainSecretsCompanion.parseBytes(BytesUtils.fromHexString(body.privKey))
+        val future = sidechainNodeViewHolderRef ? LocallyGeneratedSecret(secret)
+        Await.result(future, timeout.duration).asInstanceOf[Try[Unit]] match {
+          case Success(_) =>
+            ApiResponseUtil.toResponse(RespCreatePrivateKey25519(secret.publicImage()))
```
i-Alex, line 151: RespCreatePrivateKey25519 name seems outdated.
Probably we should name it just `RespCreatePrivateKey`

i-Alex (changes requested):

qa/httpCalls/wallet/dumpSecrets.py
MarcoOl94: Done

sdk/src/main/java/com/horizen/node/NodeWallet.java
MarcoOl94: Yes better, done.

sdk/src/main/scala/com/horizen/api/http/SidechainWalletApiRoute.scala
MarcoOl94: Done

i-Alex (approved):

paolocappelletti (approved):

PR #398, opened 2022-06-08
MessageProcessor base implementation

i-Alex:


sdk/src/main/java/com/horizen/account/state/MessageProcessor.java
```diff
@@ -0,0 +1,17 @@
+package com.horizen.account.state;
+
+public interface MessageProcessor {
+    // Initialization is going to happen only once at genesis State creation.
+    // Common pattern: declare a new fake smart contract account in the View
+    void init(AccountStateView view) throws MessageProcessorInitializationException;
```
alsala, line 6: It is not clear to me why we need the view as input parameter. For instance we will use this init() call to setup the fake smart contract address which is not dependent on the view

i-Alex: In this example we should store the new Account in the view.
So the can be modified during the initialization.

sdk/src/main/java/com/horizen/account/state/ExecutionSucceeded.java
```diff
@@ -0,0 +1,38 @@
+package com.horizen.account.state;
+
+import java.math.BigInteger;
+
+
+// Message was executed as expected.
+public class ExecutionSucceeded implements ExecutionResult {
+    private final BigInteger gasUsed; // Total used gas during Message processing
+    private final byte[] returnData; // Returned data from invocation(function result or data supplied with revert opcode)
```
JcBernack, line 9: Where do we put the logs? Also here?

sdk/src/main/java/com/horizen/account/state/MessageProcessor.java
JcBernack: This is the place where e.g. the ForgerStakesProcessor can make sure that its account exists and is not declared as "empty" by setting some (arbitrary) code hash, right?

i-Alex: right

sdk/src/main/java/com/horizen/account/state/ExecutionSucceeded.java
i-Alex: I think, that logs are add to the View directly while processing message. Similar to the Geth implementation.
So, inside StateDB we cache the logs linked to tx id.

alsala (approved):

saratnt (approved):

JcBernack (approved):

PR #402, opened 2022-06-16
EVM-161 chain id usage

rockNicktive:
Updated version of Web3j 
Added semantic validator for EIP1559 transactions
Added support for EIP1559 transactions
Added chainId param to network params

Missing:
Tests not fully updated yet, coverage low
Semantic Validator needs to be added to the sidechain app
Need to check getData() in EthereumTransaction
Need to fix message in account/state/Message.java


sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -44,7 +44,7 @@ public void BeforeEachTest() throws InvalidAlgorithmParameterException, NoSuchAl
     }
 
     @Test
-    public void ethereumRawTransactionTest() throws TransactionSemanticValidityException {
+    public void ethereumLegacyRawTransactionTest() throws TransactionSemanticValidityException {
```
i-Alex, line 47: Add also separate serialization test class, like was done for `SidechainCoreTransactionSerializerTest`.
It may contain: serialization and regression tests for legacy and eip1559 txs. Both signed an non-signed versions.

i-Alex (changes requested):

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
akrckn: Yes, there is already a serializer test for signed and unsigned eth txs - just needs to be adjusted according to the latest changes

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -47,105 +87,151 @@ public byte transactionTypeId() {
...
     public String getData() {
-        return transaction.getData();
+        return this.transaction.getData();
     }
 
     @Override
     public SignatureSecp256k1 getSignature() {
-        return signature;
+        if (this.isSigned())
+            return new SignatureSecp256k1(((SignedRawTransaction) this.transaction).getSignatureData());
```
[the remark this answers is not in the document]
rockNicktive: That could happen, as this returns the raw signature data. Alternatively we could create the signature from 
`getRealV(), getR(), getS()`. There's also a simple `verify` call we could expose

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -159,11 +245,8 @@ public String toString() {
      */
     @Override
     public byte[] messageToSign() {
-        return Bytes.concat(getAddress().getBytes(StandardCharsets.UTF_8),
-                getGasPrice().toByteArray(),
-                getGasLimit().toByteArray(),
-                getTo() != null ? getTo().address() : "".getBytes(StandardCharsets.UTF_8),
-                getValue().toByteArray(),
-                getData().getBytes(StandardCharsets.UTF_8));
+        if (this.isSigned() && ((SignedRawTransaction) this.transaction).getChainId() != null)
```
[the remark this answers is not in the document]
rockNicktive: I'd propose just leaving out the null check for the chainID

sdk/src/main/java/com/horizen/account/transaction/EthereumTransactionSerializer.java
```diff
@@ -26,30 +25,19 @@ public static EthereumTransactionSerializer getSerializer() {
...
-            Sign.SignatureData signatureData = new Sign.SignatureData(transaction.getSignature().getV(), transaction.getSignature().getR(), transaction.getSignature().getS());
-            var rlpValues = TransactionEncoder.asRlpValues(transaction.getTransaction(), signatureData);
-            RlpList rlpList = new RlpList(rlpValues);
-            encodedMessage = RlpEncoder.encode(rlpList);
-        } else {
-            encodedMessage = TransactionEncoder.encode(transaction.getTransaction());
-        }
-
+    public void serialize(EthereumTransaction tx, Writer writer) {
+        byte[] encodedMessage = tx.messageToSign();
```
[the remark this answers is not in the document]
akrckn: Yes, signature is missing here - need to decode based on tx type, maybe like before.

akrckn (changes requested):
Added small todos regarding account/state/message in code+PR description

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -47,113 +88,153 @@ public byte transactionTypeId() {
...
-                if (signature != null && this.getFrom().address().length <= Account.ADDRESS_SIZE) {
-                    throw new TransactionSemanticValidityException("Cannot create signed transaction without valid from address");
-                } else if (!signature.isValid(getFrom(), "test".getBytes(StandardCharsets.UTF_8))) {
-                    throw new TransactionSemanticValidityException("Cannot create transaction with invalid signature");
-                }
-            } catch (TransactionSemanticValidityException e) {
-                throw new TransactionSemanticValidityException(e);
-            }
+        if (this.isSigned()) {
+            if (this.getFrom().address().length <= Account.ADDRESS_SIZE)
```
JcBernack, line 106: Why `<= ADDRESS_SIZE`? Shouldn't it be `!=`?

JcBernack (commented):
I'm not too deep into this end, but to me it feels like the `EthereumTransaction` class tries to be multiple things at once (legacy and EIP1559). Does it make sense to separate this? In GETH there is a Transaction interface and multiple implementations, see the comments here:

```go
// TxData is the underlying data of a transaction.
//
// This is implemented by DynamicFeeTx, LegacyTx and AccessListTx.
type TxData interface {
	txType() byte // returns the type ID
	copy() TxData // creates a deep copy and initializes all fields

	chainID() *big.Int
	accessList() AccessList
	data() []byte
	gas() uint64
	gasPrice() *big.Int
	gasTipCap() *big.Int
	gasFeeCap() *big.Int
	value() *big.Int
	nonce() uint64
	to() *common.Address

	rawSignatureValues() (v, r, s *big.Int)
	setSignatureValues(chainID, v, r, s *big.Int)
}
```

The `LegacyTx` e.g. just returns `gasPrice` for all three of the gas-price related fields.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -47,113 +88,153 @@ public byte transactionTypeId() {
...
+    }
+
+    @JsonProperty("from")
+    public String getFromAddress() {
+        try {
+            if (this.isSigned()) return ((SignedRawTransaction) this.transaction).getFrom();
+        } catch (SignatureException e) {
+            return null;
+        }
+        return "";
```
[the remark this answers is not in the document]
rockNicktive: Fixed

sdk/src/main/java/com/horizen/account/transaction/EthereumTransactionSerializer.java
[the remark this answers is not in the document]
rockNicktive: Signature is now part of the serialization, so this issue should be resolved

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -44,66 +45,107 @@ public void BeforeEachTest() throws InvalidAlgorithmParameterException, NoSuchAl
...
-        // Test 7: ethereum transaction instance returns Signature correctly
-        assertEquals(null, ethereumTransaction.getSignature());
+        // Test 7: getMaxFeePerGas should be null, as this is a legacy transaction
+        assertNull(ethereumTransaction.getMaxFeePerGas());
+        // Test 8: getMaxPriorityFeePerGas should be null, as this is a legacy transaction
+        assertNull(ethereumTransaction.getMaxPriorityFeePerGas());
+        // Test 9: getChainId should be null, as this is a legacy transaction
+        assertNull(ethereumTransaction.getChainId());
+        // Test 10: getFrom should be null, as this is an unsigned tx
+        assertNull(ethereumTransaction.getFrom());
```
JcBernack, line 100: This will always fail now as `getFrom()` returns an empty string and not null anymore, right? :sweat_smile: 
With my comment earlier I wasn't implying that anything was wrong, I was just curious if there was a difference between an empty-string and null. Maybe always use a proper address or null, and don't use empty-string.

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -44,66 +45,107 @@ public void BeforeEachTest() throws InvalidAlgorithmParameterException, NoSuchAl
...
+        // Test 13: ethereum transaction instance returns null as to is empty or has false size
+        assertNull(ethereumTransaction.getTo());
 
-        // Test 10: ethereum transaction instance returns to proposition address correctly
+        // Test 14: ethereum transaction instance returns to proposition address correctly
         RawTransaction toTx = RawTransaction.createTransaction(someValue,
                 someValue, someValue, "00112233445566778899AABBCCDDEEFF01020304", someValue, "");
         EthereumTransaction toEthereumTransaction = new EthereumTransaction(toTx);
         AddressProposition toProposition = new AddressProposition(BytesUtils.fromHexString("00112233445566778899AABBCCDDEEFF01020304"));
         assert (Arrays.equals(toEthereumTransaction.getTo().address(), toProposition.address()));
```
JcBernack, line 116: Could use `assertArrayEquals()` everywhere there is an `assert(Arrays.equals(x,y))`.

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -157,7 +206,18 @@ public void ethereumEIP1559RawTransactionTest() throws TransactionSemanticValidi
...
+    public void ethereumSignedRawTransactionTest() {
+        // Test 0: direct constructor
+        try {
+            var someTx = new EthereumTransaction(
+                    "0x", someValue, someValue, someValue,
+                    someValue, "", signedRawTransaction.getSignatureData()
+            );
+            assertNotNull(someTx.getSignature());
+        } catch (Exception ignored) {
+            fail("Should not happen");
```
JcBernack, line 218: Hm, if an exception means the test failed, I don't think you need or should catch it at all. Just let it throw and the test will fail. If an exception is expected for the test to succeed I have been using `assertThrows()`.

@i-Alex What do you think is best practice here?

JcBernack (changes requested):

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
rockNicktive: Oh yes, you're right. Missed that when I fixed it

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
i-Alex: `assertThrows` is good. It will allow to fail the test with a proper message and result structure.

i-Alex (approved):

i-Alex:
> I'm not too deep into this end, but to me it feels like the `EthereumTransaction` class tries to be multiple things at once (legacy and EIP1559). Does it make sense to separate this? In GETH there is a Transaction interface and multiple implementations, see the comments here:
> 
> ```go
> // TxData is the underlying data of a transaction.
> //
> // This is implemented by DynamicFeeTx, LegacyTx and AccessListTx.
> type TxData interface {
> 	txType() byte // returns the type ID
> 	copy() TxData // creates a deep copy and initializes all fields
> 
> 	chainID() *big.Int
> 	accessList() AccessList
> 	data() []byte
> 	gas() uint64
> 	gasPrice() *big.Int
> 	gasTipCap() *big.Int
> 	gasFeeCap() *big.Int
> 	value() *big.Int
> 	nonce() uint64
> 	to() *common.Address
> 
> 	rawSignatureValues() (v, r, s *big.Int)
> 	setSignatureValues(chainID, v, r, s *big.Int)
> }
> ```
> 
> The `LegacyTx` e.g. just returns `gasPrice` for all three of the gas-price related fields.

In SDK we follow the rule that

akrckn (approved):
Changes regarding the test will be made as discussed, the rest seems ok.

JcBernack (approved):

akrckn (approved):

PR #404, opened 2022-06-16
Release 0.3.5: Snark keys generation fix

i-Alex:


DanieleDiBenedetto (approved):
Everything fine, but I wonder if we could ever make this supportedSegmentSize parameteric. It's true that the performance gap is not huge, but since maxKeys is customizable then also supportedSegmentSize should be

alsala (approved):

cronicc (approved):

PR #406, opened 2022-06-17
EoaMessageProcessor introduced

i-Alex:


sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  override def init(view: AccountStateView): Unit = {
+    // No actions required for transferring coins during genesis state initialization.
+  }
+
+  override def canProcess(msg: Message, view: AccountStateView): Boolean = {
+    // Can process only EOA to EOA transfer, so when:
+    // 1. "from" account must be an EOA account - no "code" set;
+    // 2. transaction "data" is empty.
+    msg.getData.isEmpty && view.getCodeHash(msg.getFrom.address()).isEmpty
```
JcBernack, line 24: I just checked: A TX with data is still a valid EOA to EOA transaction in Ethereum. When the `to` address has no code the data is ignored. It is just very uncommon because adding data increases the intrinsic gas required for the transaction.

sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  override def process(msg: Message, view: AccountStateView): ExecutionResult = {
+    view.subBalance(msg.getFrom.address(), msg.getValue) match {
```
JcBernack, line 28: I assume we have already validated that the sender has enough balance for the `value` and gas fees, correct?
So far `subBalance` does not do any validation and will also happily set the accounts balance to a negative value.

sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
```diff
@@ -0,0 +1,44 @@
...
+      case Failure(reason) =>
+        log.error(s"Unable to subtract ${msg.getValue} wei from ${msg.getFrom}", reason)
+        return new ExecutionFailed(gasUsed, new Exception(reason))
+      case _ => // do nothing
+    }
+
+    view.addBalance(msg.getTo.address(), msg.getValue) match {
+      case Failure(reason) =>
+        log.error(s"Unable to add ${msg.getValue} wei to ${msg.getTo}", reason)
+        return new ExecutionFailed(gasUsed, new Exception(reason))
```
JcBernack, line 38: The only way a balance modification can really fail is a programming error on our side, e.g. if the StateDB handle is invalid or something completely explodes - do we even want to handle that as an "execution failed"?

sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
JcBernack: I think we could decide not to support that, but to be as compatible as possible we should allow data and ignore it.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
i-Alex: good, thanks. Let's be more compatible: so it is allowed but ignored.
I will update the spec

sdk/src/main/scala/com/horizen/account/state/EoaMessageProcessor.scala
i-Alex: right, it is checked outside in AccountState, and theoretically never can occur. But considering that interface allow it to fail -> better to process this case.

JcBernack (approved):

PR #408, opened 2022-06-17
add simple python rpc test interface and example with stf

akrckn:


akrckn:
As we are using one rpc route and methods are defined in the json body, should we although stay to the specified format defined in the jira issue?
That would mean, that we have to add support for all methods in the framework then - that's no problem at all, just wanted to ask before doing so.

```
res = sc_node1.eth_getTransactionByHash("0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b")
hash = res["blockHash"]
...
```

qa/sc_evm_bootstrap.py
```diff
@@ -70,6 +71,20 @@ def run_test(self):
         print("mempool contents:")
         pprint.pprint(response)
 
+        # request chainId via rpc route
+        params = []
+        request = test_framework.util.create_json2_rpc_request("eth_chainId", params, "1")
+        response = sc_node.ethv1(request)
```
i-Alex, line 77: here I would expect to have a syntax like:
`chain_id = sc_node.eth_chainId()`

i-Alex (changes requested):

i-Alex:
> As we are using one rpc route and methods are defined in the json body, should we although stay to the specified format defined in the jira issue? That would mean, that we have to add support for all methods in the framework then - that's no problem at all, just wanted to ask before doing so.
> 
> ```
> res = sc_node1.eth_getTransactionByHash("0x88df016429689c079f3b2f6ad39fa052532c56795b733da78a91ebe6a713944b")
> hash = res["blockHash"]
> ...
> ```

actually we don't need to support all methods. We have everything in STF to make it recognized in a runtime.

i-Alex (approved):

PR #410, opened 2022-06-17
St/withdrawal msg proc

saratnt:
Still todos:
- define a proper amount of gas spent for each operation
- check if in Ethereum there are predefined types of errors that we should use
- log


sdk/src/main/scala/com/horizen/account/state/WithdrawalMsgProcessor.scala
```diff
@@ -13,14 +12,20 @@ import scorex.util.serialization.{Reader, Writer}
 import scala.collection.JavaConverters.seqAsJavaListConverter
 import scala.util.{Failure, Success, Try}
 
-object WithdrawalMsgProcessor extends AbstractFakeSmartContractMsgProcessor {
+trait WithdrawalRequestProvider extends AbstractFakeSmartContractMsgProcessor {
```
i-Alex, line 15: I think WithdrawalRequestProvider should NOT extend AbstractFakeSmartContractMsgProcessor
Instead `object WithdrawalMsgProcessor extends WithdrawalRequestProvider with WithdrawalRequestProvider`

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -1,27 +1,30 @@
...
-                        val stateDb: StateDB,
-                        messageProcessors: Seq[MessageProcessor]) extends StateView[SidechainTypes#SCAT, AccountStateView] with AccountStateReader {
+class AccountStateView(val metadataStorageView: AccountStateMetadataStorageView,
+                       val stateDb: StateDB,
+                       messageProcessors: Seq[MessageProcessor]) extends StateView[SidechainTypes#SCAT, AccountStateView] with AccountStateReader {
   view: AccountStateView =>
 
   override type NVCT = this.type
 
+  lazy val withdrawalReqProvider: WithdrawalRequestProvider = messageProcessors.find(_.isInstanceOf[WithdrawalRequestProvider]).get.asInstanceOf[WithdrawalRequestProvider]
```
i-Alex, line 26: why not just return `WithdrawalMsgProcessor` object?
Seems safe

i-Alex (commented):
Just a few comments.
Then we may merge it to the base branch for the more comprehensive review.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
saratnt: I wanted to keep the possibility to "inject" the withdrawalmesgprocessor from outside, using the messageProcessors list.
Using directly WithdrawalMsgProcessor object would have removed this possibility.

i-Alex: Good. Then no observations.

sdk/src/main/scala/com/horizen/account/state/WithdrawalMsgProcessor.scala
saratnt: Done

i-Alex (approved):

PR #412, opened 2022-06-17
add EvmMessageProcessor

JcBernack:


examples/evmapp/src/main/java/com/horizen/examples/EvmMessageProcessor.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Can process messages that require an EVM invocation:
+     * <ol>
+     * <li>when "to" address is empty; deploying a new contract</li>
+     * <li>when the account at the "to" address has non-empty code; invoke smart contract</li>
+     * </ol>
+     */
+    @Override
+    public boolean canProcess(Message msg, AccountStateView view) {
+        return msg.getTo() == null || view.getCodeHash(msg.getTo().address());
```
i-Alex, line 21: @alsala  @saratnt `msg.getTo() != null` must be accounted by other msg processors

i-Alex, line 21: `view.getCodeHash(msg.getTo().address())` can be replaced with `view.isSmartContractAccount(msg.getTo.address())` (was defined in the base branch recently)

libevm/lib/service_state.go
```diff
@@ -231,17 +241,11 @@ func (s *Service) StateRemoveStorage(params StorageParams) error {
...
-
-// chunk keys are generated by hashing the original key and the chunk index
+// chunk keys are generated by hashing a salt, the original key and the chunk index
+// the salt was added to reduce the risk of accidental hash collisions because similar strategies
+// to generate storage keys might be used by the caller
 func getChunkKey(key common.Hash, chunkIndex int) common.Hash {
-	chunkIndexBytes := big.NewInt(int64(chunkIndex)).Bytes()
-	hashedData := append(key.Bytes(), chunkIndexBytes...)
-	return crypto.Keccak256Hash(hashedData)
+	return crypto.Keccak256Hash(chunkKeySalt.Bytes(), key.Bytes(), common.BigToHash(big.NewInt(int64(chunkIndex))).Bytes())
```
i-Alex, line 248: We may keep `chunkKeySalt` already in bytes, to avoid getting bytes representation every time we use read\write chunked record.
Also for `chunkIndex` do we really need this heavy construction `common.BigToHash(big.NewInt(int64(chunkIndex))).Bytes()`? Is there any simpler solution?

libevm/lib/service_state.go
```diff
@@ -112,6 +116,14 @@ func (s *Service) StateCommit(params HandleParams) (error, common.Hash) {
 	return nil, hash
 }
 
+// TODO: we might want to change this to "StateEmpty()" because the logic behind Exists() is a bit unpredictable
```
i-Alex, line 119: good, let's do it and change in the places of usage.

libevm/lib/service_state.go
JcBernack: Regarding `chunkKeySalt`: This is internally already a byte array. This would only create a new go-slice, basically a view, of the array. I changed it anyway.

Regarding the int conversion:
That construction is heavy and ugly, I agree, but it does a bunch of things:
- Using a `big.Int` takes care of encoding/decoding incl. endianness and the inner `int64(chunkIndex)` is just a type cast because the constructor only accepts `int64`.
- The outermost ` common.BigToHash()` takes care of properly padding to the correct length (32 bytes).

I refactored it to make clearer what is happening.

libevm/lib/service_state.go
JcBernack: done

i-Alex (approved):

alsala (approved):

examples/evmapp/src/test/java/com/horizen/examples/EvmMessageProcessorIntegrationTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class EvmMessageProcessorIntegrationTest extends MessageProcessorTestBase {
+    private Database storage;
+    private AccountStateView stateView;
+
+    private static final byte[] deployCode = BytesUtils.fromHexString(
```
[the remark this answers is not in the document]
JcBernack: yes, I added a comment that it's coming from `libevm/contracts/Storage.sol`

PR #415, opened 2022-06-21
Mempool upper bound limit

paolocappelletti:
added conf parameters:
mempool.maxSize 
mempool.minFeeRate
To limit maximum mempool size and minimim tx fee to be inserted in mempool


sdk/src/test/scala/com/horizen/SidechainMemoryPoolTest.scala
```diff
@@ -123,4 +156,59 @@ class SidechainMemoryPoolTestTestsExtension
...
+    assertEquals("MemoryPool must have correct size ", list.size-1, memoryPool.size)
+    assertEquals("Lowest fee transaction must be present ", true, memoryPool.getTransactionById(lowestFeeTx.id()).isPresent)
+
+    //add one more tx, causing the maxPoolSize to be reached and lowest one to be evicted
+    assertEquals("Put tx operation must be success.", true, memoryPool.put(list.apply(list.size - 1)).isSuccess)
+    assertEquals("MemoryPool must have correct size ", list.size-1, memoryPool.size)
+    assertEquals("Lowest fee transaction must not be present ", false, memoryPool.getTransactionById(lowestFeeTx.id()).isPresent)
+
+  }
+
```
MarcoOl94, line 207: If you have time i would like to add a test case here where we add a tx with feeRate equals to the tx with the lowest feeRate already in the mempool.

qa/sc_mempool_min_fee_rate.py
```diff
@@ -0,0 +1,139 @@
...
+        #add tx with higher feerate
+        error = False
+        try:
+            txHighId = sendCoinsToAddress(sc_node1, sc_address_2, 1000, 200)
+            #this tx has size 174 and feerate 1149
+        except:
+            error = True
+        assert_false(error)
+
+        time.sleep(5)
```
MarcoOl94, line 123: I guess you are not using sync_mempool() because the nodes mempool will never be the same. Please add a comment.

MarcoOl94 (changes requested):

MarcoOl94 (approved):

qa/resources/template.conf
```diff
@@ -30,6 +30,14 @@ scorex {
       wsServerPort = 0
   }
 
+  mempool {
```
i-Alex, line 33: Check also the `template_predefined_genesis.conf` and its usage.
I think, we need to add the same there

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -64,9 +70,20 @@ class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry]
     filter(t => !txs.exists(_.id == t.id))
   }
 
+  /**
+   * Filters the txs cointained in the mempool, recalculates the used bytes, and return an instance of itself.
+   * @param condition to use to filter the transactions
+   */
   override def filter(condition: SidechainTypes#SCBT => Boolean): SidechainMemoryPool = {
+    usedPoolSizeBytes = 0
     unconfirmed.retain { (k, v) =>
```
i-Alex, line 79: I'm just thinking if it is correct that we use `retain` function here?
I guess it must be `filter` and then return `new SidechainMemoryPool(res)`
otherwise, anyone from outside may wrongly remove some txs.

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -6,17 +6,23 @@ import com.horizen.node.NodeMemoryPool
...
 
-class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry])
+class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry], mempoolSettings: MempoolSettings)
   extends scorex.core.transaction.MemoryPool[SidechainTypes#SCBT, SidechainMemoryPool]
   with SidechainTypes
   with NodeMemoryPool
+  with ScorexLogging
 {
+  var maxPoolSizeBytes : Long =  mempoolSettings.maxSize * 1024 * 1024
+  var usedPoolSizeBytes : Long = 0
```
i-Alex, line 23: why `0`?
I guess we may take in consideration the `unconfirmed` (can be not empty)

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -109,12 +132,37 @@ class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry]
...
+   * Add tx to the mempool.
+   * Lowest fee-rate txs are removed in case max mempool size is reached.
+   * @return true if the transaction has been inserted succesfully
+   */
+  def addWithSizeCheck(entry: SidechainMemoryPoolEntry) : Boolean = {
+    val orderedEntries = unconfirmed.values.toSeq.sortWith((a,b) => {
+      a.feeRate.getFeeRate() < b.feeRate.getFeeRate()
+    })
+    while (usedPoolSizeBytes + entry.feeRate.getSize() > maxPoolSizeBytes){
+      val lastEntry = orderedEntries.take(1).last
```
i-Alex, line 153: `orderedEntries` are ordered from lowest fee rate to highest one, right?

`orderedEntries.take(1).last` can be replaced with `orderedEntries.head`

Moreover, in case mempool was empty and the very first tx exceed the limit (for example, it was set to 0 Mb), we will get an exception here.
Use `headOption` and check if entry present.

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -129,14 +177,18 @@ class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry]
...
+        addWithSizeCheck(entry)
+      }
     }
 
     this
   }
 
   override def remove(tx: SidechainTypes#SCBT): SidechainMemoryPool = {
     unconfirmed.remove(tx.id)
+    usedPoolSizeBytes = usedPoolSizeBytes - tx.size()
```
i-Alex, line 191: Instead of using `tx.size()` we may, take the result of `unconfirmed.remove(tx.id)`.
If result is `Some(entry)` -> get entry.size() and decrease `usedPoolSizeBytes`.
In case we tried to remove nonexistent tx we don't decrease  `usedPoolSizeBytes`

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -109,12 +132,37 @@ class SidechainMemoryPool(unconfirmed: TrieMap[String, SidechainMemoryPoolEntry]
...
     new Success[SidechainMemoryPool](this)
   }
 
+  /**
+   * Add tx to the mempool.
+   * Lowest fee-rate txs are removed in case max mempool size is reached.
+   * @return true if the transaction has been inserted succesfully
+   */
+  def addWithSizeCheck(entry: SidechainMemoryPoolEntry) : Boolean = {
+    val orderedEntries = unconfirmed.values.toSeq.sortWith((a,b) => {
```
i-Alex, line 149: Just want to point that every time we add new Tx, we sort the whole collection from scratch. It costs.
Can we use some other data type?

sdk/src/test/scala/com/horizen/SidechainMemoryPoolTest.scala
```diff
@@ -0,0 +1,222 @@
...
+    assertEquals("Put operation must be success.", memoryPool.put(tx2).isSuccess,
+      true)
+    assertEquals("Size must be 2.", memoryPool.size, 2)
+    assertEquals("usedPoolSizeBytes must be equals to tx1+tx2 size", memoryPool.usedPoolSizeBytes,
+      tx.size() + tx2.size())
+
+    memoryPool.remove(tx)
+    assertEquals("Size must be 1.", memoryPool.size, 1)
+    assertEquals("usedPoolSizeBytes must be equals to tx2 size", memoryPool.usedPoolSizeBytes, tx2.size())
+
```
i-Alex, line 47: add the test case where trying to remove nonexisting Tx, check resulting usedPoolSizeBytes

sdk/src/test/scala/com/horizen/SidechainMemoryPoolTest.scala
```diff
@@ -0,0 +1,222 @@
...
+import com.horizen.transaction.RegularTransaction
+import com.horizen.utils.FeeRate
+import org.mockito.Mockito
+import scorex.util.ModifierId
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
+class SidechainMemoryPoolTest
```
i-Alex, line 18: A few more test cases to check:
1) Create mempool with non empty map if Txs, check initial size
2) Create mempool with settings `mempool.maxSize = 0` try to add tx.

i-Alex (changes requested):

qa/resources/template.conf
paolocappelletti: we have default values, no needed to add explicitly the conf there

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/utils/MempoolMap.scala
```diff
@@ -0,0 +1,98 @@
...
+      case Some(value) => map.get(value.txid)
+      case None => None
+    }
+  }
+
+  /**
+   * Take n lowest entries sorted by feeRate (ascending)
+   */
+  def takeLowest(n: Int) : Seq[SidechainMemoryPoolEntry]   = {
+    idsSortedByFeeRate.toList.take(n).map(item => map.get(item.txid).get)
```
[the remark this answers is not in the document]
paolocappelletti: if I remove it it expects  a Set as return, but we want a Seq to preserve order

i-Alex (approved):

PR #422, opened 2022-06-28
Evm-184 update http api

rockNicktive:


sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -68,28 +73,144 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+      val ethTx = EthereumTransactionSerializer.getSerializer.parseBytes(body.payload)
+      // lock the view and try to create CoreTransaction
+      applyOnNodeView { sidechainNodeView =>
+        validateAndSendTransaction(ethTx)
+      }
+    }
+  }
+
+  def signTransaction: Route = (post & path("signTransaction")) {
+    entity(as[ReqCreateRawTransaction]) {
```
i-Alex, line 200: I expect that we need to add one more argument - account address.

So, we need to sign Tx with the real account key that belongs to the node.
Signing with random key has no sence, because tx validation will fail because of balance is 0

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -68,28 +73,144 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+          new SignatureData(
+            body.payload.signature_v.get,
+            body.payload.signature_r.get,
+            body.payload.signature_s.get)
+        else
+          null
+      )
+      // lock the view and try to create CoreTransaction
+      applyOnNodeView { sidechainNodeView =>
+        validateAndSendTransaction(ethTx)
```
i-Alex, line 153: I think, in case tx was not yet signed, we need to find the proper account and sign tx, otherwise `validateAndSendTransaction` will always return an error, because tx has no valid signature.
We need to do something like in `sendCoinsToAddress` method.

Same in `createLegacyTransaction`

i-Alex (changes requested):
Looks good in general. Some points are missed.

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
rockNicktive: Would it make sense in your opinion to generally add an optional param `from` which can be used to create the signature, so that it goes like this

`is from specified? 
   yes => look for the corresponding account and use that 
   no  => pick the first account you find which satisfies needed conditions (gas cost in the future, and value of tx)`

i-Alex: If user want to choose first one from the wallet there is a endpoint for it `sendCoinsToAddress`

rockNicktive: I was thinking of unifying the behavior between those three functions instead, so that the user knows that it's either the first one that fits or the specific one they chose, no matter which function they're calling to do it

rockNicktive: In the new changes, the behavior is the same for all the routes with ethereum transactions. This makes it easy for the user to handle these, since they all act in the same way

i-Alex (approved):
Great!

PR #426, opened 2022-06-29
Created new trait to be used by Message Processors for accessing Acco…

saratnt:
…untStateView


sdk/src/main/scala/com/horizen/account/state/BaseAccountStateView.scala
```diff
@@ -0,0 +1,34 @@
+package com.horizen.account.state
+
+import com.horizen.evm.StateDB
+
+import java.math.BigInteger
+import scala.util.Try
+
+trait BaseAccountStateView extends AccountStateReader {
+  val stateDb: StateDB
```
i-Alex, line 9: Let's replace `val stateDB` with `getStateDbHandler(): Int` method.
Then in EvmMessageProc use it instead of `view.stateDB`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/state/BaseAccountStateView.scala
saratnt: Done

i-Alex (approved):

saratnt:
I have modified BaseAccountStateView in order to expose ResourceHandle instead of the actual handle.

JcBernack (approved):
Much better! :+1:

PR #429, opened 2022-06-30
account block root hashes

JcBernack:
- use libevm `TrieHasher` to generate AccountBlock transactions merkle root hash
- wip: add some tests to check if root hashes using GETH transactions and web3j transactions are identical


sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -351,5 +352,38 @@ public void ethereumSignedEIP1559TransactionTest() {
...
+            var txs = generateTransactions(count);
+            var rlpTxs = Arrays
+                    .stream(txs)
+                    .map(tx -> TransactionEncoder.encode(
+                            tx,
+                            // make sure we also encode a signature, even if it is empty to make the results comparable to GETH
+                            new Sign.SignatureData(new byte[0], new byte[0], new byte[0])
+                    ))
+                    .toArray(byte[][]::new);
+            System.out.format("transactions root hash (%4d) %s%n", count, Numeric.toHexString(TrieHasher.Root(rlpTxs)));
```
i-Alex, line 386: Do we have some regression data in GETH tests to compare with?

sdk/src/main/scala/com/horizen/account/block/AccountBlock.scala
```diff
@@ -138,6 +139,8 @@ object AccountBlock extends ScorexEncoding {
 
   def calculateTransactionsMerkleRootHash(sidechainTransactions: Seq[SidechainTypes#SCAT]): Array[Byte] = {
     // TODO: Use Eth friendly way to calculate merkle tree root.
-    Utils.ZEROS_HASH
+    // TODO: this assumes the binary representation of transactions exactly match the ethereum RLP encoding,
+    //  which is not true currently because the serializer prepends the payload with the length
+    TrieHasher.Root(sidechainTransactions.map(tx => tx.bytes).toArray)
```
i-Alex, line 144: check also `AccountBlock.verifyTransactionsDataConsistency` method. Where we should verify transactions to be compatible with the root in the header

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
JcBernack: yes, added

i-Alex (approved):

alsala (approved):

PR #434, opened 2022-07-01
Account state & view improvements

i-Alex:


sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -103,44 +98,111 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
...
+      throw new TransactionSemanticValidityException(s"Transaction ${tx.id} is invalid: signature is invalid")
+
+    // Check that "from" is EOA address
+    if(!isEoaAccount(tx.getFrom.address()))
+      throw new TransactionSemanticValidityException(s"Transaction ${tx.id} is invalid: from account is not EOA")
+
+    // Check the nonce
+    val stateNonce: BigInteger = getNonce(tx.getFrom.address())
+    val txNonce: BigInteger = tx.getNonce
+    stateNonce.compareTo(txNonce) match {
```
alsala, line 139: Apparently this match has problems and throws a scala.MatchError: 0 (of class java.lang.Integer), I tried using:

```
    val result = stateNonce.compareTo(txNonce)
    if (result > 0) {
      throw new TransactionSemanticValidityException(s"Transaction ${tx.id} is invalid: nonce ${txNonce} is to high (expected nonce is $stateNonce)")
    } else if (result < 0) {
      throw new TransactionSemanticValidityException(s"Transaction ${tx.id} is invalid: nonce ${txNonce} is to low (expected nonce is $stateNonce)")
    }
```

sdk/src/main/scala/com/horizen/account/state/Account.scala
```diff
@@ -1,6 +0,0 @@
-package com.horizen.account.state
-
-case class Account(nonce: Long,
```
alsala, line 3: The UT WithdrawalMsgProcessorTest.scala still uses this class, we can remove the testInit() there, we have the analogous UT which targets stateDb

sdk/src/main/java/com/horizen/account/node/NodeAccountState.java
```diff
@@ -1,9 +1,8 @@
 package com.horizen.account.node;
 
-import com.horizen.account.state.Account;
 import com.horizen.account.state.AccountStateReader;
 import com.horizen.node.NodeStateBase;
 
 public interface NodeAccountState extends NodeStateBase, AccountStateReader {
```
alsala, line 6: Do we still need NodeStateBase? Now both its derived traits NodeAccountState and NodeState are empty

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -103,44 +98,111 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
...
+
+    // Do the checks and prepay gas
+    val bookedGasPrice: BigInteger = preCheck(ethTx)
+
+    // Set Tx context for stateDB, to know where to keep EvmLogs
+    setupTxContext(ethTx)
+    val message: Message = Message.fromTransaction(ethTx)
+
+    // Increase the nonce by 1
+    increaseNonce(message.getFrom.address())
```
alsala, line 187: Shall we increase the nonce here of at the very bottom of the method when we are sure we will not thrown any exception?

i-Alex: we need to increase nonce in both success and fail.
That's why I put it before doing a snapshot

sdk/src/main/java/com/horizen/account/node/NodeAccountState.java
i-Alex: in future we may put into `NodeStateBase` something common to both UTXO and Account model, for example withdrawal epoch info.
At the moment we keep it empty because there is no UTXO state implemented compatible to the new model

sdk/src/main/scala/com/horizen/account/state/Account.scala
i-Alex: Done

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
i-Alex: right, in this case block is invalid and rejected.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -115,16 +115,52 @@ public TransactionSerializer serializer() {
...
-        if (this.isSigned()) {
-            if (this.getFrom().address().length != Account.ADDRESS_SIZE)
-                throw new TransactionSemanticValidityException("Cannot create signed transaction without valid from address");
-            if (!this.getSignature().isValid(this.getFrom(), this.messageToSign()))
-                throw new TransactionSemanticValidityException("Cannot create signed transaction with invalid " +
-                        "signature");
-
+            throw new TransactionSemanticValidityException(String.format("Transaction [%s] is semantically invalid: " +
+                    "non-positive gas limit", id()));
+        if (getTo() == null && getData().length == 0)
```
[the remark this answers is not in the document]
i-Alex: good point.
`getTo` expect to return `null` in case of smart contract declaration.
And now we probably have a bug in our EthTransaction implementation. Must return `null` as soon as `getToAddress` return null

alsala (approved):

saratnt (approved):

PR #435, opened 2022-07-04
SDK-166 - add custom storage names parameter to DbTool

davehorizen:
In this pull request:
- added custom storage names parameter to DbTool
- python test _sc_storage_recovery_with_csw_ updated with SimpleApp custom storage names


paolocappelletti (changes requested):
You are assuming that we have always custom storages, but that may not happen.
Since we have already some optional parameters I think we have to switch from "positional" parameters to something identified by a param name:
like: 
 <program name> [-s=<storages data dir path>] [-n=<custom storage names>] [-f=<path to json file>] <command name> [<json data>]\n" 

Please also update the printed hellp:
https://github.com/HorizenOfficial/Sidechains-SDK/blob/db/SDK-166/tools/dbtool/src/main/java/com/horizen/DbToolCommandProcessor.java#L107

davehorizen:
> You are assuming that we have always custom storages, but that may not happen. Since we have already some optional parameters I think we have to switch from "positional" parameters to something identified by a param name: like: < program name> [ -s=< storages data dir path > ] [ -n=< custom storage names > ] [ -f=< path to json file > ] < command name > [< json data >]
> 
> Please also update the printed hellp: https://github.com/HorizenOfficial/Sidechains-SDK/blob/db/SDK-166/tools/dbtool/src/main/java/com/horizen/DbToolCommandProcessor.java#L107

Right now the DB tool is only used in the python tests and not by end users. Moreover, changes in the DB tool interface will introduce a difference with the Bootstrapping tools (of the same package) that still reads the parameters in a positional manner.
The interface changes is a great idea but it is better to take care of it in a separate task that change the DB and Bootstrapping tools interface.

With the last commit I updated the printed helper and the python test _sc_storage_recovery_without_csw_

alsala (approved):

i-Alex (approved):
It works.

In general I'm thinking why do we need at all the checks against the `storageNames` set in the DB pool. Instead we can set the only argument - full path to the storage, then inside we will check if the file exists and will try to open this level db based storage.
@alsala what do you think?

i-Alex:
Replaced by https://github.com/HorizenOfficial/Sidechains-SDK/pull/509
Merged as a part of 0.4.1

PR #438, opened 2022-07-04
[SDK-213] Added python test to check FeePaymentBoxes inside CSW

MarcoOl94:
Verified that the FeePaymentBoxes are included in the SidechainWalletCswDataStorage with a new Python test.

### DIFF
--- a/qa/sc_csw_in_fee_payment.py
+++ b/qa/sc_csw_in_fee_payment.py
@@ -153,13 +153,14 @@ class ScCSWInFeePaymentTest(SidechainTestFramework):
 
         fee_payments = http_block_getFeePayments(sc_node, sc_block_id)
         assert_equal(len(fee_payments["feePayments"]), 1)
+        fee_payment_box_id = fee_payments["feePayments"][0]["id"]
 
         all_boxes = http_wallet_allBoxes(sc_node)
-        fee_payment_box_id = ""
+        found = False
         for box in all_boxes:
-            if (box["value"] == self.FEE*2):
-                fee_payment_box_id = box["id"]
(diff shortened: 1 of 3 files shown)


i-Alex (approved):
Good to know that there is no bug for this case.

qa/sc_csw_in_fee_payment.py
```diff
@@ -0,0 +1,317 @@
...
+        we0_end_epoch_cum_sc_tx_comm_tree_root = we0_end_mcblock_json["scCumTreeHash"]
+        print("End cum sc tx cum comm tree root hash in withdrawal epoch 0 = " + we0_end_epoch_cum_sc_tx_comm_tree_root)
+        sc_block_id = generate_next_block(sc_node, "first node")
+        check_mcreferencedata_presence(we0_end_mcblock_hash, sc_block_id, sc_node)
+
+        # Verify that we have FeePayment Box in the last block and in the wallet
+        block = http_block_findById(sc_node, sc_block_id)
+        assert_false(block["block"]["header"]["feePaymentsHash"] == "0000000000000000000000000000000000000000000000000000000000000000")
+
+        fee_payments = http_block_getFeePayments(sc_node, sc_block_id)
```
paolocappelletti, line 154: more safe to extract the boxid from fee_payments and use it below to search inside all_boxes instead of using the value

paolocappelletti (changes requested):

qa/sc_csw_in_fee_payment.py
MarcoOl94: Right, thanks

paolocappelletti (approved):

PR #439, opened 2022-07-05
As/state and receipts roots

alsala:


sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -192,17 +195,38 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
     val processor = messageProcessors.find(_.canProcess(message, this)).getOrElse(
       throw new IllegalArgumentException(s"Transaction ${ethTx.id} has no known processor.")
     )
+
+    val receipt = new EthereumReceipt()
+    receipt.setTransactionType(ethTx.version())
+    receipt.setTransactionHash(txHash)
+
+    if (!isEoaAccount(ethTx.getTo.address()) )
```
akrckn, line 203: Possible Nullpointer exception if getTo is null

alsala: Fixed, thanks

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  def update(txHash: Array[Byte], txIndex: Int, blHash: Array[Byte], blNumber: Int, idx: Int): EthereumLog =
+    EthereumLog(this.consensusDataLog, txHash, txIndex, blHash, blNumber, idx, removed=0)
+
+  override def toString: String = {
+
+    var txHashStr : String = "null"
+    var blockHashStr: String = "null"
+
+    if (transactionHash != null)
```
i-Alex, line 37: can it be null? same for `blockHash`

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+    for (i <- 0 until topicsArraySize) {
+      writer.putBytes(log.consensusDataLog.topics(i).toBytes)
+    }
+
+    val data = log.consensusDataLog.data
+    writer.putInt(data.length)
+    writer.putBytes(data)
+
+    // derived
+    // TODO (shall we put these 4? They are the same as belonging receipt
```
i-Alex, line 68: I think we can derive.

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+    writer.putBytes(log.blockHash)
+    writer.putInt(log.blockNumber)
+
+    writer.putInt(log.logIndex)
+    writer.putInt(log.removed)
+  }
+
+  override def parse(reader: Reader): EthereumLog = {
+    // consensus data
+    val address: Array[Byte] = reader.getBytes(20)
```
i-Alex, line 80: we may use address length constant

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+    writer.putBytes(data)
+
+    // derived
+    // TODO (shall we put these 4? They are the same as belonging receipt
+    writer.putBytes(log.transactionHash)
+    writer.putInt(log.transactionIndex)
+    writer.putBytes(log.blockHash)
+    writer.putInt(log.blockNumber)
+
+    writer.putInt(log.logIndex)
```
i-Alex, line 74: I think that log index can be derived from the receipt as it knows all the logs.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -111,14 +116,52 @@ class AccountState(val params: NetworkParams,
       stateView.applyMainchainBlockReferenceData(mcBlockRefData).get
     }
 
-    for (tx <- mod.sidechainTransactions) {
-      stateView.applyTransaction(tx).get
+    // get also list of receipts, useful for computing the receiptRoot hash
+    val receiptList = new ListBuffer[EthereumReceipt]()
+    val blockNumber = stateView.getHeight
```
i-Alex, line 121: probably `stateView.getHeight + 1`

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -111,14 +116,52 @@ class AccountState(val params: NetworkParams,
...
+          // update cumulative gas used so far
+          cumGasUsed = consensusDataReceipt.cumulativeGasUsed
+          val ethTx = tx.asInstanceOf[EthereumTransaction]
+
+          val txHash = idToBytes(ethTx.id)
+
+          val contractAddress = if (ethTx.getTo != null && !stateView.isEoaAccount(ethTx.getTo.address()) ) {
+            ethTx.getTo.address()
+          } else {
+            new Array[Byte](Address.LENGTH)
```
i-Alex, line 138: how can we get the real contract address?
maybe it is returned by EVM message processor in "returnData" bytes?

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -111,14 +116,52 @@ class AccountState(val params: NetworkParams,
...
+          }
+
+          // get a receipt obj with non consensus data too (logs updated too)
+          val fullReceipt = EthereumReceipt(consensusDataReceipt,
+                      txHash, txIndex, blockHash, blockNumber, txGasUsed, contractAddress).updateLogs()
```
i-Alex, line 143: I think that we can avoid calling the `.updateLogs()` constructor is enough.
Receipt should contain only a `EthereumConsensusDataLog` and have a lazy method to get the `EthereumLog`

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -192,17 +196,34 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
...
+
+    val consensusDataReceipt : EthereumConsensusDataReceipt = processor.process(message, this) match {
       case success: ExecutionSucceeded =>
-        success.gasUsed()
+        val evmLogs = getLogs(txHash)
+        val logs = evmLogs.map(element => new EthereumLog(
+          new EthereumConsensusDataLog(element)))
+        val gasUsed = success.gasUsed()
+        new EthereumConsensusDataReceipt(
+          ethTx.version(), ReceiptStatus.SUCCESSFUL.id, prevCumGasUsed.add(gasUsed), logs)
```
i-Alex, line 207: Good example here.
We may use the EvmLog class directly, no need to wrap it by `EthereumConsensusDataLog`.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -331,12 +361,15 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
     stateDb.getNonce(address)
   }
 
-  override def getAccountStateRoot: Array[Byte] = metadataStorageView.getAccountStateRoot
+  override def getLogs(txHash: Array[Byte]): Array[EvmLog] = {
+    stateDb.getLogs(txHash)
```
i-Alex, line 365: I think, to make it work we need to implement `setupTxContext` that will call stateDB.txLookup() method.
Otherwise, statedb doesn't know how to store the logs.

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -256,6 +283,14 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
       case _ => // do nothing
     }
 
+    receiptsOpt.foreach(receipts => {
+        for (r <- receipts) {
+          val key = r.transactionHash
```
i-Alex, line 288: let's construct the key using as a `hash( "receipt" | r.transactionHash)` similar to other cases.

sdk/src/test/scala/com/horizen/account/receipt/EthereumLogTest.scala
```diff
@@ -0,0 +1,60 @@
...
+import scorex.crypto.hash.Keccak256
+
+class EthereumLogTest
+  extends JUnitSuite
+    with MockitoSugar
+{
+
+  @Test def receiptSimpleEncodeDecodeTest(): Unit = {
+    val ethereumLog = createTestEthereumConsensusDataLog
+    println(ethereumLog)
```
i-Alex, line 19: let's remove prints

sdk/src/test/scala/com/horizen/account/receipt/EthereumLogTest.scala
```diff
@@ -0,0 +1,60 @@
...
+    val decodedConsensusDataLog = EthereumConsensusDataLog.rlpDecode(dataBytes)
+    System.out.println(decodedConsensusDataLog)
+    assertEquals(
+      BytesUtils.toHexString(ethereumLog.address.toBytes),
+      BytesUtils.toHexString(decodedConsensusDataLog.address.toBytes))
+  }
+
+}
+
+object EthereumLogTest {
```
i-Alex, line 33: let's put it as a trait into `Fixtures` package

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -62,11 +75,23 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
-    val stateRoot: Array[Byte] = new Array[Byte](MerkleTree.ROOT_HASH_LENGTH)
-    val receiptsRoot: Array[Byte] = new Array[Byte](MerkleTree.ROOT_HASH_LENGTH)
+    // 1. create a view and apply all transactions in the list.
+    // (see also comments in collectTransactionsFromMemPool)
+    // At the end we will have the stateRoot to be stored in the block header
+    //  Also, for every processed tx, store in a list the receipt consensus data, they will be used for build the receiptsRoot:
+    //   - cumulative Gas Used
+    //   - Status (success or failed)
+    //   - Logs
+    val dummyView = nodeView.state.getView
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataLog.scala
```diff
@@ -0,0 +1,87 @@
...
+      topicsStr = topicsStr.concat(" " + BytesUtils.toHexString(t.toBytes))
+    }
+    topicsStr = topicsStr.concat("}")
+    String.format(
+      s"EthereumLog (log consensus data) { address=${BytesUtils.toHexString(address.toBytes)}, topics=$topicsStr, data=${BytesUtils.toHexString(data)}}"
+    )
+  }
+}
+
+object EthereumConsensusDataLog {
```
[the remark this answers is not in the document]
alsala: For the time being I left it here, go has its own lib which is compatible

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -0,0 +1,219 @@
...
+  def getStatus: ReceiptStatus = {
+    status match {
+      case 0 => ReceiptStatus.FAILED
+      case 1 => ReceiptStatus.SUCCESSFUL
+    }
+  }
+
+  def isStatusOK: Boolean = status == ReceiptStatus.SUCCESSFUL.id
+
+  def setLogsBloom(): EthereumConsensusDataReceipt = {
```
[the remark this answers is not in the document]
alsala: Removed, it was actually not used

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -0,0 +1,219 @@
...
+import com.horizen.account.receipt.ReceiptTxType.ReceiptTxType
+import com.horizen.utils.BytesUtils
+import org.web3j.rlp._
+
+import java.math.BigInteger
+import java.nio.ByteBuffer
+import java.util
+import scala.collection.mutable.ListBuffer
+
+case class EthereumConsensusDataReceipt(transactionType: Int, status: Int, cumulativeGasUsed: BigInteger, logs: Seq[EthereumLog], logsBloom: Array[Byte]) {
```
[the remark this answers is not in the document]
alsala: Now it has a Seq[EvmLog]

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+                         removed : Int
+                       ) extends BytesSerializable  {
+
+
+  override type M = EthereumLog
+
+  override def serializer: ScorexSerializer[EthereumLog] = EthereumLogSerializer
+
+  def this(consensusDataLog : EthereumConsensusDataLog) {
+    this(consensusDataLog, new Array[Byte](0), -1, new Array[Byte](0), -1, -1, -1)
```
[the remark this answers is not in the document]
alsala: Ok

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+                         logIndex : Int,
+                         removed : Int
+                       ) extends BytesSerializable  {
+
+
+  override type M = EthereumLog
+
+  override def serializer: ScorexSerializer[EthereumLog] = EthereumLogSerializer
+
+  def this(consensusDataLog : EthereumConsensusDataLog) {
```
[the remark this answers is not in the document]
alsala: Correct, I removed it

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -0,0 +1,106 @@
...
+    this(consensusDataLog, new Array[Byte](0), -1, new Array[Byte](0), -1, -1, -1)
+  }
+
+  def update(txHash: Array[Byte], txIndex: Int, blHash: Array[Byte], blNumber: Int, idx: Int): EthereumLog =
```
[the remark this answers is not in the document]
alsala: After discussion, I removed this method and added another one for deriving full log info from receipt

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
alsala: Should not be, but for the time being I left that check

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
alsala: Done

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
alsala: Now non consensus info of a log are derived from belonging receipt

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: We already updated the height at this point

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
alsala: Ok

sdk/src/test/scala/com/horizen/account/receipt/EthereumLogTest.scala
alsala: Done

libevm/lib/service_hash_test.go
```diff
@@ -57,15 +59,26 @@ func generateReceipts(count int) types.Receipts {
...
+			Data:    common.Hex2Bytes("aabbccdd"),
+		}
+		receipt.Logs[0] = log
 		receipt.Bloom = types.CreateBloom(types.Receipts{receipt})
 		// These three are non-consensus fields:
 		//receipt.BlockHash
 		//receipt.BlockNumber
-		receipt.TransactionIndex = uint(v)
+		//receipt.TransactionIndex = uint(v)
+		//fmt.Printf("receipt cumGas %d\n", receipt.CumulativeGasUsed)
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
alsala: Ok, I added the implementation. I also added the txIndex to the signature since evm needs it too

sdk/src/test/scala/com/horizen/account/receipt/EthereumLogTest.scala
```diff
@@ -0,0 +1,60 @@
...
+  @Test def receiptSimpleEncodeDecodeTest(): Unit = {
+    val ethereumLog = createTestEthereumConsensusDataLog
+    println(ethereumLog)
+    val encodedLog = EthereumConsensusDataLog.rlpEncode(ethereumLog)
+    println(BytesUtils.toHexString(encodedLog))
+    // read what you write
+    val dataBytes = encodedLog
+    val decodedConsensusDataLog = EthereumConsensusDataLog.rlpDecode(dataBytes)
+    System.out.println(decodedConsensusDataLog)
+    assertEquals(
```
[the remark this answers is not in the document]
alsala: Ok. done

sdk/src/test/scala/com/horizen/account/receipt/EthereumLogTest.scala
alsala: Ok, done

evm/src/main/java/com/horizen/evm/utils/Hash.java
```diff
@@ -32,6 +32,19 @@ public byte[] toBytes() {
         return Arrays.copyOf(bytes, LENGTH);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Hash h = (Hash) o;
+        return Arrays.equals(this.toBytes(), h.toBytes());
```
i-Alex, line 40: let's use `return Arrays.equals(this.bytes, h.bytes);` to prevent unnecessary copy operations

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: The problem was that at the genesis initialization phase we incremented height twice, I added a check that the block hash used as version is not the null array. The correct line here is `stateView.getHeight + 1` as you pointed out.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: After discussion and some research on docs and geth code I fixed it getting the contractAddress via w3j util function, equivalent to the createAddress() in geth (triggered also by CREATE opcode).
Note: geth has also a CREATE2 opcode which may be optionally used in a smart contract solidity implementation in order to deploy another contract with a predefined address, created with a different formula.

evm/src/main/java/com/horizen/evm/utils/Hash.java
```diff
@@ -32,6 +32,19 @@ public byte[] toBytes() {
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Arrays.hashCode(toBytes());
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -41,43 +28,60 @@ case class EthereumLog(
...
-object EthereumLogSerializer extends ScorexSerializer[EthereumLog]{
+object EthereumLog {
+  def derive(consensusDataLog : EvmLog, txHash: Array[Byte], txIndex: Int, blHash: Array[Byte], blNumber: Int, idx: Int): EthereumLog =
+    EthereumLog(consensusDataLog, txHash, txIndex, blHash, blNumber, idx)
+}
 
-  override def serialize(log: EthereumLog, writer: Writer): Unit = {
-    // consensus data
-    writer.putBytes(log.consensusDataLog.address.toBytes)
+object EvmLogSerializer extends ScorexSerializer[EvmLog]{
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataLog.scala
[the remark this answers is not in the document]
alsala: Ok, Done

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -132,9 +133,14 @@ class AccountState(val params: NetworkParams,
...
-            ethTx.getTo.address()
+          // The contract address created, if the transaction was a contract creation
+          val contractAddress = if (ethTx.getTo == null) {
+            // this w3j util method is equivalent to the createAddress() in geth triggered also by CREATE opcode.
+            // Note: geth has also a CREATE2 opcode which may be optionally used in a smart contract solidity implementation
+            // to deploy another contract with a predefined address.
+            generateContractAddress(ethTx.getFrom.address, ethTx.getNonce)
           } else {
+            // otherwise a zero filled address
             new Array[Byte](Address.LENGTH)
```
i-Alex, line 144: I think in case to is not null we should return empty array, not a 20 bytes of zeros. 20 zeros is a valid contract address name.
official docs say that we need to return `null` https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionreceipt.
Check how `null` is encoded

i-Alex (changes requested):

i-Alex (approved):

PR #446, opened 2022-07-08
wip: move eth service to scala class, add base logic for sendRawTransaction

akrckn:


i-Alex (approved):
I think we may stop here an merge the current progress.

PR #447, opened 2022-07-08
Evm 185 stf smart contract env

rockNicktive:


i-Alex (commented):
Good
Let's have a python test similar to the `sc_evm_bootstrap.py`, where we will send a tx that deploys the smart contact using this tool, then invoke methods of that smart contract

rockNicktive:
Sounds good. I will get that done after the evm staticcal is done

rockNicktive:
Once eth_call is fixed and works properly, I would like to automatically parse outputs of smart contract calls and things like that, so that the parsing and getting of outputs doesn't have to be done manually again. See the TODOs @i-Alex

qa/SidechainTestFramework/account/requirements.txt
```diff
@@ -0,0 +1,4 @@
+eth_abi
+pycryptodome
+rlp
+eth_utils
```
i-Alex, line 4: For me the names were "eth-abi" and "eth-utils"

qa/SidechainTestFramework/account/smart_contract_resources/contracts/folderthing/Solfile.sol
```diff
@@ -0,0 +1,15 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract ExampleERC20Contract is ERC20 {
```
i-Alex, line 6: why do we have this contract declared twice here and in `contracts/folderthing/Othercontract.sol`?

qa/SidechainTestFramework/account/smart_contract_resources/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "smart_contract_resources",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "engines": {
+    "npm": ">=8.1.2",
+    "node": ">=16.13.2"
```
i-Alex, line 11: Let's add the requirements info to the `README.md` file.
I want to see there that we need to install `npm`, `node` and `yarn` of the given versions (`>=...`).
Then python requirements specified in `account/requirements.txt`

qa/sc_evm_smart_contract_transaction.py
```diff
@@ -0,0 +1,216 @@
...
+    - Extract genesis info
+    - Start SC node with that genesis info
+
+Test:
+    For the SC node:
+        - verify the MC block is included
+"""
+
+
+class SCEvmBootstrap(SidechainTestFramework):
```
i-Alex, line 36: need to rename the class to avoid the name collisions later, for example:
SCEvmBootstrap -> SCEvmStorageContract

qa/sc_evm_smart_contract_transaction.py
```diff
@@ -0,0 +1,216 @@
...
+Configuration: bootstrap 1 SC node and start it with genesis info extracted from a mainchain node.
+    - Mine some blocks to reach hard fork
+    - Create 1 SC node
+    - Extract genesis info
+    - Start SC node with that genesis info
+
+Test:
+    For the SC node:
+        - verify the MC block is included
+"""
```
i-Alex, line 33: 1) Let's update the test description.
2) Let's simplify the test itself, here we want to verify the behavior of the StorageContract only. So need to send some FT coins first, then do the StorageContract related checks. No need to to Eoa transfers, check balances, network and chain ids, etc.

qa/SidechainTestFramework/account/ac_use_smart_contract.py
```diff
@@ -0,0 +1,220 @@
...
+        return encoded_string
+
+    def decode(self, output):
+        return decode_abi(self.outputs, output)
+
+    def __str__(self):
+        return f'{self.sigHash} :: {self.name}({",".join(self.inputs)}) -> ({",".join(self.outputs)})'
+
+
+class SmartContract:
```
i-Alex, line 32: let's add the simple description and docs for the public methods of this class: what are inputs, outputs, etc.

i-Alex (changes requested):
Looks good.
Just a few comments.
One more thing, that I can see the following warning while running the test: "Solidity 0.8.14 is not fully supported yet. You can still use Hardhat, but some features, like stack traces, might not work correctly." Can we change the solidity version to the one supported by hardhat?

qa/SidechainTestFramework/account/requirements.txt
rockNicktive: I needed to specifically install pycryptodome, the others were not enough for me. These are the packages I'm using, so while rlp is probably always going to be a dependency for eth_abi and eth_util I'd still rather keep it as an extra mention.

qa/SidechainTestFramework/account/smart_contract_resources/contracts/folderthing/Solfile.sol
rockNicktive: This was for testing purposes - the way smart contracts are searched for, every contract needs to have a unique name. I did this to check if / how this works and if the error returned makes sense.

i-Alex (approved):

PR #454, opened 2022-07-18
As/forger roots handling

alsala:


evm/src/main/java/com/horizen/evm/StateDB.java
```diff
@@ -10,6 +10,12 @@ public class StateDB extends ResourceHandle {
     public static byte[] EMPTY_CODE_HASH =
             Converter.fromHexString("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
 
+    /**
+     * TrieHasher.Root() of an empty byte array
+     */
+    public static byte[] EMPTY_ROOT_HASH =
+            Converter.fromHexString("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421");
```
saratnt, line 17: Aren't Converter.fromHexString/toHexString the same as BytesUtils.fromHexString/toHexString?

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -44,15 +44,31 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
   type MS = AccountState
   type MP = AccountMemoryPool
 
-  def computeReceiptRoot(receiptList: Seq[EthereumReceipt]) : Array[Byte] = {
+  def computeReceiptRoot(receiptList: Seq[EthereumConsensusDataReceipt]) : Array[Byte] = {
     // 1. for each receipt item in list rlp encode and append to a new leaf list
     // 2. compute hash
-    TrieHasher.Root(receiptList.map(r => EthereumConsensusDataReceipt.rlpEncode(r.consensusDataReceipt)).toArray)
+    TrieHasher.Root(receiptList.map(EthereumConsensusDataReceipt.rlpEncode).toArray)
```
saratnt, line 50: There is this exact piece of code  in AccountBlock at line 51. Could it be possible to put it in a single place in order to avoid code duplication?

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -400,4 +359,109 @@ object AccountState {
...
+              throw new IllegalArgumentException("Could not apply tx, block gas limit exceeded")
+            }
+          }
+
+          val ethTx = tx.asInstanceOf[EthereumTransaction]
+
+          val txHash = idToBytes(ethTx.id)
+
+          // The contract address created, if the transaction was a contract creation
+          val contractAddress = if (ethTx.getTo == null) {
```
saratnt, line 435: Are you sure ethTx.getTo == null is enough for selecting a tx creating a smart contract? In EthereumTransaction constructor, if to param is null it is changed in "".
Maybe creating a method in EthereumTransaction class for testing if it is a tx creating a smart contract could be useful

saratnt (approved):

sdk/src/main/scala/com/horizen/account/state/MessageProcessorUtil.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import scala.collection.JavaConverters.asScalaBufferConverter
+
+object MessageProcessorUtil {
+  def getMessageProcessorSeq(params: NetworkParams, customMessageProcessors: java.util.List[MessageProcessor]): Seq[MessageProcessor] = {
+    Seq(
+      EoaMessageProcessor,
+      WithdrawalMsgProcessor,
+      ForgerStakeMsgProcessor(params),
+    ) ++ customMessageProcessors.asScala
```
i-Alex, line 13: So we may do `.asScala` in bootstrapping tool class.

tools/sctool/src/main/java/com/horizen/CommandProcessor.java
```diff
@@ -598,12 +604,25 @@ private void processGenesisInfo(JsonNode json) {
...
 
-                byte[] stateRoot = new byte[MerkleTree.ROOT_HASH_LENGTH];
-                byte[] receiptsRoot = new byte[MerkleTree.ROOT_HASH_LENGTH];
-                AddressProposition forgerAddress = new AddressProposition(new byte[Account.ADDRESS_SIZE]);
+                byte[] receiptsRoot = StateDB.EMPTY_ROOT_HASH; // empty root hash (no receipts)
+
+                AddressProposition forgerAddress =
+                  new AddressProposition(Arrays.copyOf(
+                          sidechainCreation.getScCrOutput().address(),
+                          com.horizen.account.utils.Account.ADDRESS_SIZE)); // taken from the creation
```
i-Alex, line 625: This is a tricky point. I mean, which part of original 32 bytes take as an eth address.
In the AccountStateView we have the following code:
`val ownerAddressProposition = new AddressProposition(BytesUtils.reverseBytes(scOut.address.take(com.horizen.account.utils.Account.ADDRESS_SIZE)))`
Let's add a utility function that will extract EthAddress from FT and ScCreation address field.
Then use it in all places: here and twice in `applyMainchainBlockReferenceData`

tools/sctool/src/main/java/com/horizen/CommandProcessor.java
```diff
@@ -571,6 +573,10 @@ private void processGenesisInfo(JsonNode json) {
 
             MainchainBlockReference mcRef = MainchainBlockReference.create(mcBlockBytes, params, versionsManager).get();
 
+            // TODO for the time being we have the sc creation only, but we might support also FT
```
i-Alex, line 576: do we need this todo? I guess ScCr+FT case is supported even now.

evm/src/main/java/com/horizen/evm/StateDB.java
alsala: Yes, they are, and there is another similar util in package Numeric. I used this one for being consistent with the overall usage in this package, but probably we could live with one- util package only.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -400,4 +359,109 @@ object AccountState {
...
+
+  def blockSizeExceeded(blockSize: Int, txCounter: Int): Boolean = {
+    if (txCounter > SidechainBlockBase.MAX_SIDECHAIN_TXS_NUMBER || blockSize > SidechainBlockBase.MAX_BLOCK_SIZE)
+      true // stop data collection
+    else {
+      false // continue data collection
+    }
+  }
+
+  def applyAndGetReceipts(stateView: AccountStateView,
```
i-Alex, line 376: I honestly vote to separate the logic for state and forger.
As you may see this method is full of `if`s and there will be even more when forger will have more complex approach of transaction selection.
Let's keep the original code in the state as is, and have a separate consistent one for the Forger.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -62,45 +78,54 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
 
+    // dispose of the view
     dummyView.close()
 
     // 2. Compute the receipt root
     val receiptsRoot: Array[Byte] = computeReceiptRoot(receiptList)
 
-    val forgerAddress: AddressProposition = new AddressProposition(new Array[Byte](Account.ADDRESS_SIZE))
+    // 3. As forger address take first address from the wallet
+    val firstAddress = nodeView.vault.allSecrets().asScala
```
i-Alex, line 113: let's use `secretsOfType`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: Should be safe, after checking the EthereumTransaction object creation code the address is null in case of 'to' string null, "", and also "0x"

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/block/AccountBlock.scala
```diff
@@ -47,15 +47,24 @@ class AccountBlock(override val header: AccountBlockHeader,
...
       //throw new InconsistentSidechainBlockDataException("invalid receipt root hash")
     }
   }
 
+  @throws(classOf[InconsistentSidechainBlockDataException])
+  def verifyStateRootDataConsistency(stateRoot: Array[Byte]): Unit = {
+    if (!java.util.Arrays.equals(stateRoot, header.stateRoot)) {
+      log.error("CHECK IS DISABLED: update forger & bootstrapping tool first!")
+      // TODO: uncomment when ready
+      //throw new InconsistentSidechainBlockDataException("invalid receipt root hash")
```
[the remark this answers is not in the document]
alsala: Ok, done. I uncommented the code too.

sdk/src/main/scala/com/horizen/account/AccountSidechainNodeViewHolder.scala
```diff
@@ -31,7 +31,7 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
                                      consensusDataStorage: ConsensusDataStorage,
                                      stateMetadataStorage: AccountStateMetadataStorage,
                                      stateDbStorage: Database,
-                                     customMessageProcessors: Seq[MessageProcessor],
+                                     customMessageProcessors: java.util.List[MessageProcessor],
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/MessageProcessorUtil.scala
alsala: ok

tools/sctool/src/main/java/com/horizen/CommandProcessor.java
alsala: Ok, done

tools/sctool/src/main/java/com/horizen/CommandProcessor.java
alsala: Removed

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: Now using two separate methods

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
i-Alex: Good.
I still tempted to avoid static methods here, but return it back to the AccountState class as it was before.
The same for the forger, put `blockGasLimitExceeded`, `blockSizeExceeded` and `tryApplyAndGetReceipts` methods to the Forger.

sdk/src/main/scala/com/horizen/account/AccountSidechainApp.scala
```diff
@@ -49,7 +49,7 @@ class AccountSidechainApp @Inject()
    @Named("CustomApiGroups") customApiGroups: JList[ApplicationApiGroup],
    @Named("RejectedApiPaths") rejectedApiPaths : JList[Pair[String, String]],
    @Named("ChainInfo") chainInfo : ChainInfo,
-   @Named("CustomMessageProcessors") customMessageProcessors: JList[MessageProcessor]
+   @Named("CustomMessageProcessors") customMessageProcessors: Seq[MessageProcessor]
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
alsala: Ok done, for the time being I just left the dummy `blockGasLimitExceeded` since it will be used by both modules

evm/src/main/java/com/horizen/evm/StateDB.java
saratnt: ok

i-Alex (approved):

PR #455, opened 2022-07-19
St/missing api

saratnt:


sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -254,14 +261,235 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+        }
+        //TODO Probably getFittingSecret would need to take into account also gas
+        val secret = getFittingSecret(sidechainNodeView, None, valueInWei)
+        secret match {
+          case Some(secret) =>
+            val to = BytesUtils.toHexString(ForgerStakeMsgProcessor.ForgerStakeSmartContractAddress.address())
+            val nonce = body.nonce.getOrElse(sidechainNodeView.getNodeState.getNonce(secret.publicImage.address))
+            val data = encodeSpendStakeCmdRequest(body.spendForgerStakeInfo)
+            val tmpTx: EthereumTransaction = new EthereumTransaction(
+              body.chainId,
```
i-Alex, line 329: it think, we should retrieve `chainId` from NetworkParams instead of from user input.
Same in other places.

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -254,14 +261,235 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+  def encodeAddNewStakeCmdRequest(forgerStakeInfo: TransactionForgerOutput): String = {
+    val blockSignPublicKey = new PublicKey25519Proposition(BytesUtils.fromHexString(forgerStakeInfo.blockSignPublicKey.getOrElse(forgerStakeInfo.ownerAddress)))
+    val vrfPubKey = new VrfPublicKey(BytesUtils.fromHexString(forgerStakeInfo.vrfPubKey))
+    val addForgerStakeInput = AddNewStakeCmdInput(ForgerPublicKeys(blockSignPublicKey, vrfPubKey), new AddressProposition(BytesUtils.fromHexString(forgerStakeInfo.ownerAddress)))
+    val data = BytesUtils.toHexString(Bytes.concat(BytesUtils.fromHexString(ForgerStakeMsgProcessor.AddNewStakeCmd), addForgerStakeInput.encode()))
+    data
+  }
+
+  def encodeSpendStakeCmdRequest(spendForgerStake: TransactionSpendForgerStake): String = {
+    val signature = new SignatureSecp256k1(BytesUtils.fromHexString(spendForgerStake.v),
```
i-Alex, line 469: We don't expect to retrieve the signature from the user.
User give us only `stakeId`. Then we shoudl take the owner from the State, and create a signature using the corresponding private key from the wallet.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
saratnt: Done

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
saratnt: Done

i-Alex (approved):

alsala (approved):

PR #456, opened 2022-07-19
EVM Event Creation Utility

akrckn:
<WIP>

### DIFF
--- a/sdk/src/test/java/com/horizen/account/event/EthereumEventTest.java
+++ b/sdk/src/test/java/com/horizen/account/event/EthereumEventTest.java
@@ -1,0 +1,130 @@
+package com.horizen.account.event;
+
+import com.horizen.account.MyEvent3;
+import org.junit.Before;
+import org.junit.Test;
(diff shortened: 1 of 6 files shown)


sdk/src/test/java/com/horizen/account/event/EthereumEventTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.web3j.abi.datatypes.Address;
+import org.web3j.abi.datatypes.generated.Uint256;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.math.BigInteger;
+
+import static org.junit.Assert.assertEquals;
+
+public class EthereumEventTest {
```
i-Alex, line 15: Seems good.
Let's add some negative tests:
1) too much topics found
2) invalid field/method type

Also more success tests for log0-log4 cases compared with some Eth retrieved RLP encoded data

Also a test for using it in Scala: normal class and case class.

i-Alex (changes requested):

PR #458, opened 2022-07-20
Mo/sdk 216 (Added BTs limit inside the certificate)

MarcoOl94:
## Description
This PR introduces the BTs limit inside the certificate.
For more info please look at: https://www.notion.so/horizen-labs/Certificate-backward-transfers-limit-fd8b23f56cb44362a271d795d982c22f

For what regards the Fork 1 changes, in the first place i was trying to use the function `SidechainState.getCurrentConsensusEpochInfo()` but it requires to have some ForgingStakeInfos and there are some cases where this is not always true.
For this reason I used another approach, see `SidechainState.backwardTransferLimitEnabled()`

## Jira Ticket
[216](https://horizenlabs.atlassian.net/jira/software/c/projects/SDK/boards/14?modal=detail&selectedIssue=SDK-216&assignee=615f31c32f6aed0068c67291)

## Changes

- Added new max WithdrawalRequestBoxes per certificate (3999)
- Added new Transaction validation rule that checks that the transaction doesn't create more than 3999 WBs.
- Added new rule inside BlockMutualityCheck.
- Added new logic in the Forger that takes in account the total amount of WBs when it retrieves the mempool transactions in order to forge a block
- New Unit Tests
- New python test 

## Breaking Changes
- New consensus rules = **non backward compatible changes!**

## Checks
- [✔] Project Builds
- [✔] Breaking changes have been correctly tagged and notified

### DIFF
--- a/qa/sc_bt_limit.py
+++ b/qa/sc_bt_limit.py
@@ -14,6 +14,7 @@ from test_framework.util import fail, assert_equal, assert_true, start_nodes, \
 from httpCalls.block.findBlockByID import http_block_findById
 from httpCalls.transaction.withdrawCoins import withdrawMultiCoins
 from httpCalls.block.forgingInfo import http_block_forging_info
+from httpCalls.transaction.sendCoinsToAddress import sendCointsToMultipleAddress
 
 """
 Configuration:
@@ -25,13 +26,14 @@ Configuration:
 
 Test:
     For the SC node:
-        - ############## EPOCH 0 #####################
+        - ############## WITHDRAWAL EPOCH 0 #####################
         - Send 1 FT to SC node 1 on different addresses.
         - Generate 1 MC and 1 SC block to see FT in the sidechain.
+        - Split the FT box into some ZenBoxes
         - Generate a transaction that has 999 WithdrawalRequestBoxes
         - Generate a SC block and verify that it includes this transaction since we didn't reach the SC fork 1
 
-        - ############## EPOCH 1 #####################
+        - ############## WITHDRAWAL EPOCH 1 #####################
         - Reach the SC fork 1
         - Generate a transaction that has 999 WithdrawalRequestBoxes
         - Generate a SC block and verify that it doesn't include this transaction (2 MC block reference included = 798 WBs slots)
@@ -42,24 +44,27 @@ Test:
         - Generate 2 transaction that create 80 WBs each one
         - Generate a SC block and verify that it contains only one of these 2 transctions
         - Generate 2 transactions that generate 999 WBs each one
-        - Generate MC blocks until the last block of the epoch 0.
-        - Generate a SC block that inlcudes all the previous transactions (we opened up enought slots)
-        - Create a transaction that generate the reamining WBs allowed in the epoch (3999 - alreadyMinedWbs = 824 WBs)
+        - Generate MC blocks to open up enough slots
+        - Generate a SC block and verify it contains the 2 transactions of 999 WBs and the one with 80 WBs
+        - Generate MC blocks until the last block of the epoch 1.
+        - Create a transaction that produce the reamining WBs allowed per epoch
+        - Generate a SC block that inlcudes and verify that it includes this transaction
         - Generate another MC block and SC block to reach the end of the epoch
+        - Create a transaction that has 999 WBs
 
-        - ############## EPOCH 2 #####################
+        - ############## WITHDRAWAL EPOCH 2 #####################
         - Generate the first MC block of the next epoch and a SC block
-        - Verify that this SC block contains the previously generated transaction containing the remaining WBs (824)
         - Wait for certificate submission.
         - Generate 1 MC and 1 SC block including the certificate.
         - Generate a transaction that creates 300 WBs
-        - Forge a SC block and verify that it doesn't contain this transaction. We already filled remaining_wbs (824) in this epoch that only added 2 MC block references (=798 empty slots)
+        - Forge a SC block and verify that it contains this transaction and not the one with 999 WBS (we didn't open enough slots yet)
+        - Generate MC blocks to open up slots
+        - Generate a SC block and verify it contains the 999 WBs transaction
         - Generate MC blocks to reach the end of the Withdrawal epoch 1.
         - Generate a SC block that contains all of these MC blocks.
         
-        - ############## EPOCH 3 #####################
+        - ############## WITHDRAWAL EPOCH 3 #####################
         - Generate the first MC block of the next epoch and a SC block
-        - Verify that we have the 300 WBs transaction included in this SC block. We couldn't add before because we opened all the needed slots in the last block of the epoch that can't include any transactions
         - Wait for certificate submission.
 
 """
@@ -98,25 +103,8 @@ class ScBtLimitTest(SidechainTestFramework):
         is_csw_enabled = sc_node.csw_isCSWEnabled()["result"]["cswEnabled"]
         assert_true(is_csw_enabled, "Ceased Sidechain Withdrawal expected to be enabled.")
 
-        # Checks that CSW is enabled in Sidechain creation transaction on Mainchain
-        mc_block_height = mc_node.getblockcount()
-        mc_block = mc_node.getblock(str(mc_block_height))
-        mc_sc_creation_tx_id = mc_block["tx"][1]
-        mc_sc_creation_tx = mc_node.getrawtransaction(mc_sc_creation_tx_id, 1)
-
-        vsc_ccout_ = mc_sc_creation_tx["vsc_ccout"][0]
-        assert_true(vsc_ccout_["vFieldElementCertificateFieldConfig"] == certificate_field_config_csw_enabled,
-                    "Custom Field Elements Configuration in MC are wrong. Expected: " + format(certificate_field_config_csw_enabled) +
-                    ", actual: " + format(vsc_ccout_["vFieldElementCertificateFieldConfig"]))
-        assert_true("wCeasedVk" in vsc_ccout_, "CSW verification key should be present")
-
-        # Checks that Sidechain creation transaction is in SC block
-        mc_block_hash = mc_sc_creation_tx["blockhash"]
-        sc_block_id = sc_node.block_best()["result"]["block"]["id"]
-        check_mcreference_presence(mc_block_hash, sc_block_id, sc_node)
-
-        # ******************** EPOCH 0 START ********************
-        print("******************** EPOCH 0 START ********************")
+        # ******************** WITHDRAWAL EPOCH 0 START ********************
+        print("******************** WITHDRAWAL EPOCH 0 START ********************")
 
         #Verify we didn't reach the SC fork1 that includes BT limit
         consensusEpochData = http_block_forging_info(sc_node)
@@ -126,7 +114,7 @@ class ScBtLimitTest(SidechainTestFramework):
 
         # create 1 FTs in the same MC block to SC
         sc_address_1 = sc_node.wallet_createPrivateKey25519()["result"]["proposition"]["publicKey"]
-        ft_amount_1 = 100
+        ft_amount_1 = 1000
         mc_return_address_1 = mc_node.getnewaddress()
 
         forward_transfer_to_sidechain(self.sidechain_id, mc_node,
@@ -137,6 +125,10 @@ class ScBtLimitTest(SidechainTestFramework):
 
         epoch_mc_blocks_left -= 1
         # Generate 1 SC block to include FTs.
+        generate_next_blocks(sc_node, "first node", 1)[0]
+
+        #Split the UTXO
+        sendCointsToMultipleAddress(sc_node, [sc_address_1 for _ in range (10)], [10 * 1e8 for _ in range(10)], 0)
         generate_next_blocks(sc_node, "first node", 1)[0]
 
         # Create a transaction that generates 999 WBs 
@@ -162,8 +154,8 @@ class ScBtLimitTest(SidechainTestFramework):
         block_json = http_block_findById(sc_node, sc_block_id)
         check_mcreferencedata_presence(we0_end_mcblock_hash, sc_block_id, sc_node)
 
-        # ******************** EPOCH 1 START ********************
-        print("******************** EPOCH 1 START ********************")
+        # ******************** WITHDRAWAL EPOCH 1 START ********************
+        print("******************** WITHDRAWAL EPOCH 1 START ********************")
 
         # Generate first mc block of the next epoch
         we1_1_mcblock_hash = mc_node.generate(1)[0]
@@ -188,7 +180,7 @@ class ScBtLimitTest(SidechainTestFramework):
         print(consensusEpochData)
         assert_equal(consensusEpochData["bestEpochNumber"], 3)
 
-        # Based on withdrawalEpochLength = 10, maxBTsAllowedPerCertificate = 3999, we open 399 slots for WithdrawalBoxes for each MC block reference
+        # Based on withdrawalEpochLength = 11, maxBTsAllowedPerCertificate = 3999, we open 399 slots for WithdrawalBoxes for each MC block reference
         # 2 MC block mined = 399 * 2 = 798 WthdrawalBoxes allowed
 
         # Create a transaction that generates 999 WBs 
@@ -232,6 +224,7 @@ class ScBtLimitTest(SidechainTestFramework):
         block_json = http_block_findById(sc_node, sc_block_id)
         assert_equal(len(block_json["block"]["sidechainTransactions"]), 1)    
         wbs_mined += 80
+        wbs_mined += 80
 
         # Create Some WBs
 
@@ -241,9 +234,26 @@ class ScBtLimitTest(SidechainTestFramework):
         wbs_mined += 999
 
         # Generate some MC blocks to open up slots
-        mc_node.generate(epoch_mc_blocks_left - 1)
+        mc_node.generate(epoch_mc_blocks_left - 2)
         sc_block_id = generate_next_block(sc_node, "first node")
         block_json = http_block_findById(sc_node, sc_block_id)
+
+        assert_equal(len(block_json["block"]["sidechainTransactions"]), 3)
+
+        #744
+        remaining_wbs = MAX_WBS_PER_EPOCH - wbs_mined
+        wbs_left_txid = withdrawMultiCoins(sc_node, bt_addresses[:remaining_wbs], amounts[:remaining_wbs])["result"]["transactionId"]
+
+        # Generate 1 MC block. Now all the slots should be opened
+        we0_end_mcblock_hash = mc_node.generate(1)[0]
+
+        #We should be able to include the remaining WBs inside the current epoch
+        sc_block_id = generate_next_block(sc_node, "first node")
+        block_json = http_block_findById(sc_node, sc_block_id)
+        assert_equal(block_json["block"]["sidechainTransactions"][0]["id"],wbs_left_txid)
+
+        #Create another transaction that generate some WBs.
+        not_included_tx_id = withdrawMultiCoins(sc_node, bt_addresses, amounts)["result"]["transactionId"]
 
         # Generate 1 MC block to reach the end of the epoch
         we0_end_mcblock_hash = mc_node.generate(1)[0]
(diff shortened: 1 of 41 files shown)


sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -177,6 +177,37 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+    var blockWBs = alreadyMinedWBs
+    mod.transactions.foreach(tx => {
+      blockWBs += tx.newBoxes().asScala.count(box => box.isInstanceOf[WithdrawalRequestBox])
+      if (blockWBs > allowedWBs) {
+        throw new IllegalStateException(s"Exceeded the maximum number of WithdrawalBoxes allowed!")
+      }
+    })
+  }
+
+  def getOrElseWithdrawalEpochInfo(): WithdrawalEpochInfo = {
```
paolocappelletti, line 197: check if can keep private any of these new methods if not called from outside

paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -61,6 +59,19 @@ class SidechainMemoryPool private(unconfirmed: MempoolMap, mempoolSettings: Memp
     unconfirmed.values.toSeq.sortWith(sortFunc).take(limit).map(tx => tx.getUnconfirmedTx())
   }
 
+  def takeWithWithdrawalBoxesLimit(limit: Int, allowedWithdrawalBoxes: Int): Iterable[SidechainTypes#SCBT] = {
```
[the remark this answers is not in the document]
MarcoOl94: Done

sdk/src/main/scala/com/horizen/SidechainState.scala
MarcoOl94: Done

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
```diff
@@ -313,7 +313,9 @@ class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
         Seq()
       } else { // SC block is in the middle of the epoch
         var txsCounter: Int = 0
-        nodeView.pool.take(nodeView.pool.size).filter(tx => {
+        val allowedWithdrawalRequestBoxes = nodeView.state.getAllowedWithdrawalRequestBoxes(mainchainBlockReferenceDataToRetrieve.size) - nodeView.state.getAlreadyMinedWithdrawalRequestBoxesInCurrentEpoch
+        val mempoolTx = if (nodeView.state.backwardTransferLimitEnabled)  nodeView.pool.takeWithWithdrawalBoxesLimit(allowedWithdrawalRequestBoxes) else nodeView.pool.take(nodeView.pool.size)
```
i-Alex, line 317: Be careful with the forger logic.
Let's assume that allowed WBs per MC block is 100, the state is before the Fork1, so no limit on BTs.
Mempool contains Tx with 200 WBs.
Then Forger creates the next block that reach the Fork1. So Forger will add Tx with 200 WBs, but state will reject this block, because 200 > 100 (BT limit).
Proposal:
* remove `nodeView.state.backwardTransferLimitEnabled` method, and use fork manager directly in the state and forger.
* for forger get the consensus epoch number of the candidate block -> pass it to the ForkManager to make a decision if `backwardTransferLimitEnabled` is enabled.
Let's add a separate python test to verify this tricky case.

sdk/src/main/scala/com/horizen/fork/SidechainFork1.scala
```diff
@@ -1,5 +1,5 @@
 package com.horizen.fork
 
 class SidechainFork1(override val epochNumber: ForkConsensusEpochNumber) extends BaseConsensusEpochFork(epochNumber) {
```
i-Alex, line 3: Let's add more description before the class, explaining the purpose and the changes of the fork. Can be very useful for future.

i-Alex (changes requested):
Looks good!
We need to discuss the slots opening formula. I think we can be more precise here

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+
+Test:
+    For the SC node:
+        - ############## EPOCH 0 #####################
+        - Send 1 FT to SC node 1 on different addresses.
+        - Generate 1 MC and 1 SC block to see FT in the sidechain.
+        - Generate a transaction that has 999 WithdrawalRequestBoxes
+        - Generate a SC block and verify that it includes this transaction since we didn't reach the SC fork 1
+
+        - ############## EPOCH 1 #####################
```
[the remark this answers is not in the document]
MarcoOl94: These are withdrawal epoch not consensus, but if you prefere i can change them

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+        mc_block_height = mc_node.getblockcount()
+        mc_block = mc_node.getblock(str(mc_block_height))
+        mc_sc_creation_tx_id = mc_block["tx"][1]
+        mc_sc_creation_tx = mc_node.getrawtransaction(mc_sc_creation_tx_id, 1)
+
+        vsc_ccout_ = mc_sc_creation_tx["vsc_ccout"][0]
+        assert_true(vsc_ccout_["vFieldElementCertificateFieldConfig"] == certificate_field_config_csw_enabled,
+                    "Custom Field Elements Configuration in MC are wrong. Expected: " + format(certificate_field_config_csw_enabled) +
+                    ", actual: " + format(vsc_ccout_["vFieldElementCertificateFieldConfig"]))
+        assert_true("wCeasedVk" in vsc_ccout_, "CSW verification key should be present")
```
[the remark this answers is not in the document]
MarcoOl94: Done

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+        we1_2_mcblock_hash = mc_node.generate(1)[0]
+        epoch_mc_blocks_left -= 1
+
+        # Reach the SC fork 1
+        generate_next_block(sc_node, "first node", force_switch_to_next_epoch = True)
+        consensusEpochData = http_block_forging_info(sc_node)
+        print(consensusEpochData)
+        assert_equal(consensusEpochData["bestEpochNumber"], 3)
+
+        # Based on withdrawalEpochLength = 10, maxBTsAllowedPerCertificate = 3999, we open 399 slots for WithdrawalBoxes for each MC block reference
```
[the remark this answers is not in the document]
MarcoOl94: Yes sorry

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -177,6 +178,47 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+    var blockWBs = alreadyMinedWBs
+    mod.transactions.foreach(tx => {
+      blockWBs += tx.newBoxes().asScala.count(box => box.isInstanceOf[WithdrawalRequestBox])
+      if (blockWBs > allowedWBs) {
+        throw new IllegalStateException(s"Exceeded the maximum number of WithdrawalBoxes allowed!")
+      }
+    })
+  }
+
+  private def getOrElseWithdrawalEpochInfo(): WithdrawalEpochInfo = {
```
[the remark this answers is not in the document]
MarcoOl94: Done

qa/sc_bt_limit.py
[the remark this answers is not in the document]
i-Alex: oh, ok.
Then it is fine. Let's just rename to `WITHDRAWAL EPOCH N`.
Since it is critical here to understand the difference

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+        - Generate a SC block and verify that it contains only one of these 2 transctions
+        - Generate 2 transactions that generate 999 WBs each one
+        - Generate MC blocks until the last block of the epoch 0.
+        - Generate a SC block that inlcudes all the previous transactions (we opened up enought slots)
+        - Create a transaction that generate the reamining WBs allowed in the epoch (3999 - alreadyMinedWbs = 824 WBs)
+        - Generate another MC block and SC block to reach the end of the epoch
+
+        - ############## EPOCH 2 #####################
+        - Generate the first MC block of the next epoch and a SC block
+        - Verify that this SC block contains the previously generated transaction containing the remaining WBs (824)
```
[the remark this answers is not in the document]
MarcoOl94: Isn't because the last block of the epoch can't contain any transaction?

qa/sc_bt_limit.py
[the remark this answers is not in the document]
MarcoOl94: Yes good point.

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+        # Create a transaction that generates 80 WBs
+        withdrawMultiCoins(sc_node, bt_addresses[:80], amounts[:80])
+        # Create a transaction that generates 80 WBs
+        withdrawMultiCoins(sc_node, bt_addresses[:80], amounts[:80])
+
+        # Generates 1 SC block and verify that it contains only 1 of the two transactions
+        sc_block_id = generate_next_blocks(sc_node, "first node", 1)[0]
+        block_json = http_block_findById(sc_node, sc_block_id)
+        assert_equal(len(block_json["block"]["sidechainTransactions"]), 1)    
+        wbs_mined += 80
```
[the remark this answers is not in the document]
MarcoOl94: Agree thanks.

qa/sc_bt_limit.py
```diff
@@ -0,0 +1,344 @@
...
+        # Generate 1 MC block to reach the end of the epoch
+        we0_end_mcblock_hash = mc_node.generate(1)[0]
+        print("End mc block hash in withdrawal epoch 0 = " + we0_end_mcblock_hash)
+        we0_end_mcblock_json = mc_node.getblock(we0_end_mcblock_hash)
+        we0_end_epoch_cum_sc_tx_comm_tree_root = we0_end_mcblock_json["scCumTreeHash"]
+        print("End cum sc tx cum comm tree root hash in withdrawal epoch 0 = " + we0_end_epoch_cum_sc_tx_comm_tree_root)
+
+        #824
+        remaining_wbs = MAX_WBS_PER_EPOCH - wbs_mined
+        wbs_left_txid = withdrawMultiCoins(sc_node, bt_addresses[:remaining_wbs], amounts[:remaining_wbs])["result"]["transactionId"]
```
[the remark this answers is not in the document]
MarcoOl94: Yes it does, but the purpose here was to test this https://github.com/HorizenOfficial/Sidechains-SDK/pull/458/files#diff-24698af97177eec65b9683f46b8ded574be0ce35d911cd64dbeff5a5e8d2e5f5R272

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -177,6 +178,47 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+
+  private def getAllowedWithdrawalRequestBoxesPerBlock: Int = {
+    params.maxWBsAllowed / (params.withdrawalEpochLength - 1)
+  }
+
+  def getAlreadyMinedWithdrawalRequestBoxesInCurrentEpoch: Int = {
+    stateStorage.getWithdrawalRequests(getOrElseWithdrawalEpochInfo().epoch).size
+  }
+
+  def getAllowedWithdrawalRequestBoxes(numberOfMainchainBlockReferenceInBlock: Int): Int = {
```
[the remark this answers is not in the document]
MarcoOl94: The formula shouldn't be:
 open slots number =`3999 * (current_block_of_the_epoch + num_of_mc_block_ref_data_in_current_block) / (we_length - 1)` ?

sdk/src/main/scala/com/horizen/fork/SidechainFork1.scala
MarcoOl94: Sure, should i add also the info related to the openStakeTransaction?

i-Alex: yes, please

sdk/src/main/scala/com/horizen/SidechainState.scala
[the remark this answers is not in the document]
i-Alex: yes, it seems fine.
And moreover let's keep do then Min(3999, open_slots_number), in case in the future the last block of the epoch will be able to keep txs (to prevent the result > 3999)

qa/sc_bt_limit.py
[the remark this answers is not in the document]
i-Alex: Let's test here the maximum WBS slots occupied. And do the check specified in you comment later in the code.

qa/sc_bt_limit.py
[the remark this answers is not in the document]
i-Alex: will be changed because of the testflow is changed.
In general it is ok

sdk/src/main/scala/com/horizen/fork/SidechainFork1.scala
MarcoOl94: Done

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
```diff
@@ -61,6 +59,19 @@ class SidechainMemoryPool private(unconfirmed: MempoolMap, mempoolSettings: Memp
     unconfirmed.values.toSeq.sortWith(sortFunc).take(limit).map(tx => tx.getUnconfirmedTx())
   }
 
+  def takeWithWithdrawalBoxesLimit(allowedWithdrawalBoxes: Int): Iterable[SidechainTypes#SCBT] = {
+    val filteredTxs: JArrayList[SidechainTypes#SCBT] = new JArrayList[SidechainTypes#SCBT]()
+    var newWithdrawalBoxes = 0
+    take(size).foreach( tx => {
```
paolocappelletti, line 65: take(size) is useless, we could do only unconfirmed.map or something similar

paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -196,29 +196,13 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
-    stateStorage.getWithdrawalEpochInfo.getOrElse(WithdrawalEpochInfo(0,0))
-  }
-
-  private def getAllowedWithdrawalRequestBoxesPerBlock: Int = {
-    params.maxWBsAllowed / (params.withdrawalEpochLength - 1)
-  }
-
   def getAlreadyMinedWithdrawalRequestBoxesInCurrentEpoch: Int = {
-    stateStorage.getWithdrawalRequests(getOrElseWithdrawalEpochInfo().epoch).size
+    stateStorage.getWithdrawalRequests(stateStorage.getWithdrawalEpochInfo.getOrElse(WithdrawalEpochInfo(0,0)).epoch).size
```
i-Alex, line 200: let's use `getWithdrawalEpochInfo()` instead of `stateStorage.getWithdrawalEpochInfo.getOrElse(WithdrawalEpochInfo(0,0)` to avoid code duplication.

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -196,29 +196,13 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
-
-  def backwardTransferLimitEnabled: Boolean = {
-    stateStorage.getConsensusEpochNumber match {
-      case Some(consensusEpochNumber) =>
-        ForkManager.getSidechainConsensusEpochFork(consensusEpochNumber).backwardTransferLimitEnabled()
-      case None =>
-        false
-    }
+    Math.min(params.maxWBsAllowed,
+            (params.maxWBsAllowed * (stateStorage.getWithdrawalEpochInfo.getOrElse(WithdrawalEpochInfo(0,0)).lastEpochIndex + numberOfMainchainBlockReferenceInBlock)) / (params.withdrawalEpochLength - 1))
```
i-Alex, line 205: same here: `getWithdrawalEpochInfo()`

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -320,6 +304,12 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
         })
 
       lazy val numberOfWithdrawalRequestBoxes = newBoxes.count(box => box.isInstanceOf[WithdrawalRequestBox])
+      val backwardTransferLimitEnabled = stateStorage.getConsensusEpochNumber match {
```
i-Alex, line 307: I think, we have an issue here.
Let's assumes we've just forged the first SC block after SCFork1 activation.
In this case during tx validation `backwardTransferLimitEnabled = false` because the `stateStorage.getConsensusEpochNumber` will return the actual tip epoch number that is before the fork 1.

It seems that we don't need to account the fork1 here at all. The only check is that `numberOfWithdrawalRequestBoxes > params.maxWBsAllowed`, which seems to be fork independent.
In other place we verify `checkWithdrawalBoxesAllowed` that seems enough.

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
```diff
@@ -314,7 +314,8 @@ class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
       } else { // SC block is in the middle of the epoch
         var txsCounter: Int = 0
         val allowedWithdrawalRequestBoxes = nodeView.state.getAllowedWithdrawalRequestBoxes(mainchainBlockReferenceDataToRetrieve.size) - nodeView.state.getAlreadyMinedWithdrawalRequestBoxesInCurrentEpoch
-        val mempoolTx = if (nodeView.state.backwardTransferLimitEnabled)  nodeView.pool.takeWithWithdrawalBoxesLimit(allowedWithdrawalRequestBoxes) else nodeView.pool.take(nodeView.pool.size)
+        //In case we reached the Sidechain Fork1 we filter the mempool txs considering also the WithdrawalBoxes allowed to be mined in the current block.
+        val mempoolTx = if (ForkManager.getSidechainConsensusEpochFork(TimeToEpochUtils.timeStampToEpochNumber(params, timestamp))backwardTransferLimitEnabled())  nodeView.pool.takeWithWithdrawalBoxesLimit(allowedWithdrawalRequestBoxes) else nodeView.pool.take(nodeView.pool.size)
```
i-Alex, line 318: `amp))ba...` -> `amp)).ba...`
dot is missed

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/SidechainState.scala
i-Alex: Seems, we can add this check to the semantic validity of `SidechainTransaction` to detect this case even earlier

sdk/src/main/scala/com/horizen/SidechainMemoryPool.scala
MarcoOl94: The **take** function is sorting the transactions by fee rate that it's something that we want to keep. Please correct me if i'm wrong.

sdk/src/main/scala/com/horizen/SidechainState.scala
MarcoOl94: Done

sdk/src/main/scala/com/horizen/SidechainState.scala
MarcoOl94: Done

sdk/src/main/scala/com/horizen/SidechainState.scala
MarcoOl94: Good point.
Maybe it is harder to test it because at the moment we have 1000 output boxes limit per transaction but for sure it is the best point where to put it.

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
MarcoOl94: Done

i-Alex (approved):

paolocappelletti (approved):

PR #459, opened 2022-07-21
Fix for wrong list of forgers deserialization.

saratnt:


alsala (approved):

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -633,6 +633,12 @@ class ForgerStakeMsgProcessorTest
 
     })
 
+    //Check getListOfForgers
+    val forgerList2 = forgerStakeMessageProcessor.getListOfForgers(stateView)
+    assertEquals("List of forgers from getListOfForgers has different size respect the one from the fake smart contract execution", forgerList.size, forgerList2.size)
```
i-Alex, line 638: Let's define equals and hashCode methods for `AccountForgingStakeInfo` class and all sub classes.
Then compare `forgerList` fully, instead of only seq length check.

i-Alex (changes requested):
merge actual dev_evm and resolve the conflicts.

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
saratnt: Done

i-Alex (approved):

PR #460, opened 2022-07-21
wip: eth wallet related rpc

akrckn:


akrckn:
@i-Alex 
Could you help with these? Maybe after our standup?

1. I moved it to EthService, but getting the current view didn't work there, so I moved it back to finish the other tasks before. Now I would like to solve it, because I'll need it for the other part mentioned below.

```
  def ethRpc: Route = (post) {
    entity(as[JsonNode]) { body =>
      applyOnAccountView { view =>
        // TODO: optimize usage of rpcHandler (no need to create an object from scratch every time)
        // TODO: improve the usage of node and state views. Possibly put the getters into the RpcHandler
        val currentView: AccountStateView = view.state.getView
        val rpcHandler = new RpcHandler(new EthService(currentView, view, params, sidechainSettings, sidechainTransactionActorRef));
        val res = rpcHandler.apply(new RpcRequest(body))
        currentView.close()
        ApiResponseUtil.toResponseWithoutResultWrapper(res);

      }
    }
  }
```

2. How can I get the stateView of a specific block?

libevm/interop/invoke_test.go
```diff
@@ -70,6 +70,15 @@ func TestInvoke(t *testing.T) {
 	if result.EvmError != "" {
 		t.Fatalf("vm error: %v", result.EvmError)
 	}
+	/* TODO: check test
```
i-Alex, line 73: need to restore this test

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlock.java
```diff
@@ -0,0 +1,157 @@
...
+    private List<String> sealFields;
+    private String baseFeePerGas;
+
+    public EthereumBlock() {
+    }
+
+    public EthereumBlock(String number, String hash, List<String> transactions, List<EthereumTransactionView> transactionViews, Boolean hydrated, AccountBlock block) {
+        this.number = number;
+        this.hash = hash;
+        this.parentHash = "0x0";
```
i-Alex, line 43: use parent hash from `AccountBlock`

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlock.java
```diff
@@ -0,0 +1,157 @@
...
+        this.nonce = "0x0"; // no nonce
+        this.sha3Uncles = "0x0"; // no uncles
+        this.logsBloom = "0x0"; // not included in block now
+        this.transactionsRoot = "0x0";
+        this.stateRoot = Numeric.toHexString(block.header().stateRoot());
+        this.receiptsRoot = Numeric.toHexString(block.header().receiptsRoot());
+        this.miner = "0x0";
+        this.mixHash = "0x0";
+        this.extraData = "0x0";
+        this.size = String.valueOf(block.header().bytes().length);
```
i-Alex, line 53: I think it must be the size of the whole block.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumReceiptView.java
```diff
@@ -0,0 +1,62 @@
...
+    private String root;
+    private String status;
+    private String effectiveGasPrice;
+
+    public EthereumReceiptView() {
+    }
+
+    public EthereumReceiptView(EthereumReceipt receipt, EthereumTransaction tx) {
+        transactionHash = Numeric.toHexString(receipt.transactionHash());
+        transactionIndex = String.valueOf(receipt.transactionIndex());
```
i-Alex, line 34: do we expect decimal or hexadecimal format here?
I think it should be `Numeric.toHexString...` or `toHexStringWithPrefix`?
same for `blockHash`

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumReceiptView.java
```diff
@@ -0,0 +1,62 @@
...
+        blockHash = Numeric.toHexString(receipt.blockHash());
+        blockNumber = String.valueOf(receipt.blockNumber());
+        from = (tx.getFrom() != null) ? Numeric.toHexString(tx.getFrom().address()) : "";
```
i-Alex, line 37: who `getFrom` can be null?

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumReceiptView.java
```diff
@@ -0,0 +1,62 @@
...
+        to = (tx.getTo() != null) ? Numeric.toHexString(tx.getTo().address()) : "";
+        cumulativeGasUsed = Numeric.toHexStringWithPrefix(receipt.consensusDataReceipt().cumulativeGasUsed());
+        gasUsed = Numeric.toHexStringWithPrefix(receipt.gasUsed());
+        contractAddress = Numeric.toHexString(receipt.contractAddress());
+        logs = receipt.deriveFullLogs().seq().toList();
```
i-Alex, line 42: just a moment for the documentation: we don't have `removed` fields, in our case it is always `false`.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -0,0 +1,66 @@
...
+    private String yParity;
+    private String r;
+    private String s;
+
+
+    public EthereumTransactionView() {
+    }
+
+    public EthereumTransactionView(EthereumTransaction ethTx) {
+        type = String.valueOf(ethTx.transactionTypeId());
```
i-Alex, line 32: in decimal or in like `0x...` ?

sdk/src/main/scala/com/horizen/api/http/ApiResponseUtil.scala
```diff
@@ -18,28 +18,16 @@ object ApiResponseUtil {
...
-            val msg = thr.getMessage
-            if (msg != null && !msg.isEmpty)
-              SidechainApiResponse(SerializationUtil.serialize(e.code, e.description, msg))
-            else SidechainApiResponse(SerializationUtil.serialize(e.code, e.description, ""))
-          case None => SidechainApiResponse(SerializationUtil.serialize(e.code, e.description, ""))
-        }
-    }
-  }
+  def toResponseWithoutResultWrapper(response: ApiResponse): Route =
+    SidechainApiResponse(SerializationUtil.serialize(response))
```
i-Alex, line 30: why do we have this change?
I can harm the original behavior of API responses.
pls, doublecheck

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
-      var result = processor.process(msg, stateView)
-      if (!result.isFailed)
-        return new Data(result.asInstanceOf[ExecutionSucceeded].returnData())
+  //function which describes default transaction representation for answer after adding the transaction to a memory pool
+  val defaultTransactionResponseRepresentation: (Transaction => SuccessResponse) = {
+    transaction => TransactionIdDTO(transaction.id)
+  }
+
+  @RpcMethod("eth_getBlockByNumber") def getBlockByNumber(tag: Quantity, hydratedTx: Boolean): EthereumBlock = {
+    val blockId = nodeView.history.getBlockIdByHeight(Integer.parseInt(Numeric.cleanHexPrefix(tag.getValue), 16))
```
i-Alex, line 47: use `getBlockIdByTag` here to support tags like "latest"

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
+    val txViews: ListBuffer[EthereumTransactionView] = ListBuffer()
+
+    if (blockId.isEmpty) return new EthereumBlock()
+
+    val txs = getTransactionsFromBlock(getBlockById(blockId.get()))
+    txs.foreach(transaction => {
+      txViews.append(new EthereumTransactionView(transaction))
+      txHashes.append(Numeric.prependHexPrefix(transaction.id))
+    })
+    new EthereumBlock(tag.getValue, Numeric.prependHexPrefix(blockId.get()), txHashes.toList.asJava, txViews.toList.asJava, hydratedTx, nodeView.history.getBlockById(blockId.get()).get())
```
i-Alex, line 58: `nodeView.history.getBlockById(blockId.get()).get()`
don't need to get block again from the storage - it costs.
we already have a block in line 53: `val txs = getTransactionsFromBlock(getBlockById(blockId.get()))`
let's assign it to some val and reuse here.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
-    if (processor.canProcess(msg, stateView)) {
-      var result = processor.process(msg, stateView)
-      if (!result.isFailed)
-        return new Data(result.asInstanceOf[ExecutionSucceeded].returnData())
+  //function which describes default transaction representation for answer after adding the transaction to a memory pool
+  val defaultTransactionResponseRepresentation: (Transaction => SuccessResponse) = {
+    transaction => TransactionIdDTO(transaction.id)
+  }
+
+  @RpcMethod("eth_getBlockByNumber") def getBlockByNumber(tag: Quantity, hydratedTx: Boolean): EthereumBlock = {
```
i-Alex, line 46: to avoid code duplication, after retrieving `blockId` you may call `getBlockByHash`

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
+        )
+      )
     }
-    return new Data(null)
+    new Data(result.returnData)
   }
 
   @RpcMethod("eth_blockNumber") def blockNumber = new Quantity(Numeric.toHexStringWithPrefix(BigInteger.valueOf(getBlockHeight)))
 
+  private def getBlockHeight = nodeView.history.getCurrentHeight
```
i-Alex, line 109: this method is used only in the `def blockNumber`.
let's remove it at all. the name `getBlockHeight` confuse me a bit.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
 
   @RpcMethod("eth_getBalance") def GetBalance(address: String, blockNumberOrTag: Quantity) = new Quantity(getBalance(address, blockNumberOrTag))
 
+  private def getBlockIdByTag(tag: String): String = {
+    var blockId: java.util.Optional[String] = null
+    val history = nodeView.history
+    tag match {
+      case "earliest" => blockId = nodeView.history.getBlockIdByHeight(1)
+      //case "finalized" => //-39001: Unkown block
+      //case "safe" => //-39001: Unkown block
```
i-Alex, line 127: do we need to support it?
consider that we use this function for blocks/balances/tx count.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
+    val nonce = stateDbRoot.getNonce(Numeric.hexStringToByteArray(address))
     new Quantity(Numeric.toHexStringWithPrefix(nonce))
   }
 
-  @RpcMethod("net_version") def version: String = Numeric.toHexStringWithPrefix(BigInteger.TEN) // TODO: sidechainSettings.genesisData.mcNetwork or just a number????
+  @RpcMethod("net_version") def version: String = String.valueOf(networkParams.chainId)
 
-  @RpcMethod("eth_estimateGas") def estimateGas(transaction: util.LinkedHashMap[String, String]/*, tag: Quantity*/) = { // TODO: We need an estimateGas function to execute EVM and get the gasUsed
-    if(transaction.containsKey("data") && transaction.get("data") != null && transaction.get("data").length > 1)
+  @RpcMethod("eth_estimateGas") def estimateGas(transaction: util.LinkedHashMap[String, String] /*, tag: Quantity*/) = { // TODO: We need an estimateGas function to execute EVM and get the gasUsed
```
i-Alex, line 152: I think this method is important to have now, should call `eth_call` inside and get total gas used from it.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
+    new Quantity("0x1")
   }
 
-  @RpcMethod("eth_getTransactionByHash") def getTransactionByHash(transactionHash: Quantity): ResponseObject = {
-    val tx = getTransaction(transactionHash)
-    if (tx.isEmpty) return null
-    val ethTx = getEthereumTransaction(tx)
-    new ResponseObject(getTxObject(transactionHash, ethTx, tx))
+  @RpcMethod("eth_getTransactionByHash") def getTransactionByHash(transactionHash: String): EthereumTransactionView = {
+    val tx = nodeView.history.searchTransactionInsideBlockchain(Numeric.cleanHexPrefix(transactionHash))
```
i-Alex, line 164: this method is too costly.
Instead we should:
1) for given tx hash get the receipt from the state
2) extract from receipt block id and tx index
3) get block by id from history and extract tx for given index.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1,112 +1,178 @@
...
+    if (tx.isEmpty) {
+      null
```
i-Alex, line 166: if `null` is expected result in case of tx was not found?

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -123,60 +189,24 @@ class EthService(val stateView: AccountStateView, val nodeView: CurrentView[Acco
...
-  private def getBalance(address: String, tag: Quantity) = { // TODO: Add blockNumberOrTag handling
-    val balance = nodeView.state.getBalance(Numeric.hexStringToByteArray(address))
-    Numeric.toHexStringWithPrefix(balance)
+  def getTransactionReceipt(transactionHash: String): EthereumReceiptView = {
+    val receipt = stateView.getTransactionReceipt(Numeric.hexStringToByteArray(Numeric.cleanHexPrefix(transactionHash)))
+    val tx = nodeView.history.searchTransactionInsideBlockchain(Numeric.cleanHexPrefix(transactionHash))
+    if (receipt.isEmpty || tx.isEmpty) {
+      new EthereumReceiptView()
+    } else {
+      new EthereumReceiptView(receipt.get, tx.get().asInstanceOf[EthereumTransaction])
```
i-Alex, line 199: I think, we don't need to pass `tx` to `EthereumReceiptView`. All we need is stored in the `EthereumReceipt`

sdk/src/main/scala/com/horizen/account/history/AccountHistory.scala
```diff
@@ -31,9 +31,37 @@ extends com.horizen.AbstractHistory[SidechainTypes#SCAT, AccountBlockHeader, Acc
...
-  private def findTransactionInsideBlock(transactionId : String, block : AccountBlock) : JOptional[SidechainTypes#SCAT] = ???
+  private def findTransactionInsideBlock(transactionId : String, block : AccountBlock) : JOptional[SidechainTypes#SCAT] = {
+    block.transactions.find(tx => tx.id.equals(ModifierId(transactionId))) match {
+      case Some(tx) => JOptional.ofNullable(tx)
+      case None => JOptional.empty()
+    }
+  }
 
-  override def searchTransactionInsideBlockchain(transactionId: String): JOptional[SidechainTypes#SCAT] = ???
+  override def getBestBlock: AccountBlock = bestBlock
```
i-Alex, line 41: we don't need this method. Was defined in the parent class.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
akrckn: yes

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
akrckn: No, the to and from address is not or am I missing something?

libevm/interop/invoke_test.go
akrckn: That test won't be possible in this way, because the output depends on the compiler version and some more things - more about that later.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlock.java
akrckn: Where do we get it?

sdk/src/main/scala/com/horizen/api/http/ApiResponseUtil.scala
akrckn: I assume that it was changed here, because it is not used anywhere, but we should move it to AccountEthRpcRoute

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
akrckn: I just added it

i-Alex (approved):

PR #462, opened 2022-07-21
libevm build on maven test

JcBernack:
- add libevm `build` goal to mavens `generate-resources` phase
- this will make sure the libevm binary is built not just on `package` but also on `compile`, `test`, etc.


i-Alex (approved):
Good

PR #463, opened 2022-07-22
Fix forger logic in case of ommers: skip txs inclusion.

i-Alex:


sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
```diff
@@ -320,7 +327,7 @@ class ForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
 
     val feePayments = if(isWithdrawalEpochLastBlock) {
       // Current block is expect to be the continuation of the current tip, so there are no ommers.
-      require(nodeView.history.bestBlockId == branchPointInfo.branchPointId, "Last block of the withdrawal epoch expect to be a continuation of the tip.")
+      require(nodeView.history.bestBlockId == parentBlockId, "Last block of the withdrawal epoch expect to be a continuation of the tip.")
```
alsala, line 330: Now the require() should not be necessary since we have checked the condition at line 309 above

alsala (approved):

sdk/src/main/scala/com/horizen/forge/ForgeMessageBuilder.scala
i-Alex: yes, it was not necessary even before. Just a double check

VolodymyrBabaiev (approved):

PR #465, opened 2022-07-22
As/mempool tx validation

alsala:


i-Alex (approved):

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -242,7 +242,10 @@ public AddressProposition getTo() {
         throw new RuntimeException(String.format("Invalid to address length %d", to.length));
     }
 
+    @JsonIgnore
     public String getToAddress() {
+        // TODO currently we have a bug that removes 0x prefix during the serialization
```
akrckn, line 247: I assume that the `Numeric.toHexString(encodedMessage)` causes this issue, but I'll check while working on RPC.

akrckn (approved):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -380,6 +383,30 @@ class AccountState(val params: NetworkParams,
     res
   }
 
+  override def validate(tx: SidechainTypes#SCAT): Try[Unit] = Try {
+    tx.semanticValidity()
+
+    if (!tx.isInstanceOf[MC2SCAggregatedTransaction]) {
```
saratnt, line 389: Why not checking directly if it is a EthereumTransaction instead of checking if it is not an MC2SCAggregatedTransaction?

saratnt (approved):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
i-Alex: good point,
must be `EthereumTransaction` here

i-Alex (changes requested):

i-Alex (approved):

PR #468, opened 2022-07-26
St/evm 196

saratnt:


sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -382,4 +385,6 @@ class AccountStateView(private val metadataStorageView: AccountStateMetadataStor
 
   override def getStateDbHandle: ResourceHandle = stateDb
 
+  override def getTransactionBlockNumber(txId: scorex.util.ModifierId): Int = metadataStorageView.getTransactionBlockNumber(txId).get
```
i-Alex, line 388: I think, the result type must be as in Metadatastorage: `Option[Int]`,
otherwise we will throw an exception in case tx is not known

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -330,6 +365,11 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
     calculateKey(Bytes.concat("receipt".getBytes, txHash))
   }
 
+  private[horizen] def getTransactionBlockNumberKey(txId : scorex.util.ModifierId): ByteArrayWrapper = {
+    calculateKey(Bytes.concat("txblock".getBytes, txId.getBytes))
```
i-Alex, line 369: Let's do `idToBytes(txId)` instead of `txId.getBytes`

i-Alex (changes requested):
Seems good.

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
saratnt: Done

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
saratnt: My reasoning was that using an Option would have been like saying it is possible that a transaction doesn't have a block number, while in my understanding that shouldn't be possible. But in the end it is probably better using an Option and avoiding the exception.
Done

i-Alex (approved):

i-Alex (approved):

i-Alex:
Closed due to feature duplication, we have all the data needed in the TranscationReceipt.

PR #471, opened 2022-07-27
As/forging test

alsala:


sdk/src/main/scala/com/horizen/account/state/ForgerStakeMsgProcessor.scala
```diff
@@ -306,7 +306,15 @@ case class ForgerStakeMsgProcessor(params: NetworkParams) extends AbstractFakeSm
 
     if (isGenesisScCreation) {
       // no gas paid here
+
+      // increase the balance of the "forger stake smart contract” account
+      view.addBalance(fakeSmartContractAddress.address(), stakedAmount).get
+
+      // TODO
+      //view.addLog(new EvmLog concrete instance) // EvmLog will be used internally
```
i-Alex, line 314: no logs needed here. since the balance was updated by McBlockRefData not but Tx in the block

i-Alex (approved):

qa/sc_evm_forger.py
```diff
@@ -0,0 +1,426 @@
...
+        # check both nodes see the same stake list and same contract amount
+        assert_equal(
+            sc_node_1.transaction_allForgingStakes()["result"],
+            sc_node_2.transaction_allForgingStakes()["result"])
+        assert_equal(
+            get_account_balance(sc_node_1, ForgerStakeSmartContractAddress),
+            get_account_balance(sc_node_2, ForgerStakeSmartContractAddress))
+
+        # get owner pub key from the node stake list (we have only 1 item)
+        stakeList = sc_node_1.transaction_allForgingStakes()["result"]['listOStakes']
```
i-Alex, line 135: listOStakes -> listOfStakes or even better just `stakes`

qa/sc_evm_forger.py
```diff
@@ -0,0 +1,426 @@
...
+            print("Forger stake created: " + json.dumps(makeForgerStakeJsonRes))
+        self.sc_sync_all()
+
+        # Generate SC block on SC node (keep epoch)
+        generate_next_block(sc_node_1, "first node", force_switch_to_next_epoch=False)
+        self.sc_sync_all()
+        print_current_epoch_and_slot(sc_node_1)
+
+        # we now have 3 stakes
+        stakeList = sc_node_1.transaction_allForgingStakes()["result"]['listOStakes']
```
i-Alex, line 249: I think about adding the api endpoint `allActiveForgingStakes`, that will return only those who can participate in forging. Can be useful even for this test.
Can be done as a separate task in the next sprints.

i-Alex (changes requested):

qa/sc_evm_forger.py
alsala: Done

qa/sc_evm_forger.py
alsala: Ok

i-Alex (approved):

PR #480, opened 2022-08-01
Handling of AccountStateView closing.

saratnt:


sdk/src/main/scala/com/horizen/utils/ClosableResourceHandler.scala
```diff
@@ -0,0 +1,16 @@
+package com.horizen.utils
+
+trait ClosableResourceHandler {
+  //TODO This implementation is not perfect: in case of an exception calling close(), the latter exception will be
+  //propagated and it will hide the original one. A better solution will be using scala.util.Using available
+  //starting from 2.13 version.
```
JcBernack, line 6: I assume upgrading to Scala 2.13 is not simple enough to just do it?

JcBernack: Regarding possible exceptions: `statedb.close()` will never throw on the GO side, even if called multiple times for the same handle for example. So I think this is acceptable.

JcBernack (approved):
Looks good to me!

i-Alex (approved):

alsala (approved):

PR #482, opened 2022-08-02
As/fix eip155

alsala:


sdk/src/main/java/com/horizen/account/transaction/AccountTransaction.java
```diff
@@ -24,5 +24,7 @@
 
     public abstract byte[] getData();
 
+    public abstract PR getRealSignature();
```
i-Alex, line 27: It is a bit strange to see both `getRealSignature` and `getSignature` methods in the `AccountTransaction`. Let's have only `getSignature()`.

In EthTransaction we may rename `getRealSignature` to `getSignature`. And don't use the current `getSignature` at all. Instead we may define a method like `getSignatureData` that will return `Sign.SignatureData` and used in all the places needed: `toString`/RPC/serialization

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -221,7 +221,7 @@ else if (this.isSigned()) {
             var sigData = signedTx.getSignatureData();
             if (sigData.getS()[0] == 0 && sigData.getR()[0] == 0)
                 return convertToLong(sigData.getV());
-            else return ((SignedRawTransaction) this.transaction).getChainId(); // TODO check necessity
+            else return ((SignedRawTransaction) this.transaction).getChainId(); // useful for EIP155
```
i-Alex, line 224: do we really need if/else here? Can't we use only `((SignedRawTransaction) this.transaction).getChainId()` ?

sdk/src/main/java/com/horizen/account/utils/Secp256k1.java
```diff
@@ -5,9 +5,10 @@ public final class Secp256k1 {
     public static final int PRIVATE_KEY_SIZE = COORD_SIZE;
     public static final int PUBLIC_KEY_SIZE = COORD_SIZE * 2;
 
-    public static final int SIGNATURE_V_SIZE = 1;
+    // signature's v might be set to {0,1} + CHAIN_ID * 2 + 35 (see EIP155)
+    public static final int SIGNATURE_V_MAXSIZE = 4;
     public static final int SIGNATURE_RS_SIZE = 32;
-    public static final int SIGNATURE_SIZE = SIGNATURE_V_SIZE + SIGNATURE_RS_SIZE * 2;
+    public static final int SIGNATURE_MAXSIZE = SIGNATURE_V_MAXSIZE + SIGNATURE_RS_SIZE * 2;
```
i-Alex, line 11: actually never used. Can we removed it?

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -196,6 +211,27 @@ public void ethereumLegacyEIP155TransactionTest() {
...
+        byte[] toAddress = decodedTx.getTo().address();
+
+        assertEquals("892278d9f50a1da5b2e98e5056f165b1b2486d97", BytesUtils.toHexString(fromAddress));
+        assertEquals("d830603264bd3118cf95f1fc623749337342f9e9", BytesUtils.toHexString(toAddress));
+        assertEquals(1997, chainId);
+        assertEquals("3000000000000000000", decodedTx.getValue().toString());
+
+        // re-encode and check it is the same
+        byte[] encodedTx = EthereumTransactionSerializer.getSerializer().toBytes(decodedTx);
+        assertEquals(metamaskHexStr, BytesUtils.toHexString(encodedTx));
```
i-Alex, line 234: let's check the signature before and after encoding

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/transaction/AccountTransaction.java
alsala: Ok, done

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
alsala: I added some comment

sdk/src/main/java/com/horizen/account/utils/Secp256k1.java
alsala: Right, done

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
alsala: Ok, done

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -310,18 +304,21 @@ public SignatureSecp256k1 getRealSignature() {
...
                     stx.getSignatureData().getV(),
                     stx.getSignatureData().getR(),
                     stx.getSignatureData().getS());
         }
         return null;
     }
 
+    public byte[] getV() { return (getSignatureData() != null) ? getSignatureData().getV() : null; }
+    public byte[] getR() { return (getSignatureData() != null) ? getSignatureData().getR() : null; }
+    public byte[] getS() { return (getSignatureData() != null) ? getSignatureData().getS() : null; }
```
i-Alex, line 321: I think it is never used.
let's remove this methods, of?

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -349,22 +346,30 @@ public String toString() {
...
+        ByteBuffer buffer = ByteBuffer.allocate(Long.BYTES);
+        buffer.putLong(x);
+        return buffer.array();
+    }
+
+    // w3j way for converting bytes, it works also with generic byte contents (not only Long.BYTES byte arrays)
+    public static long convertToLong(byte[] bytes) {
+        BigInteger bi = Numeric.toBigInt(bytes);
+        return bi.longValue();
+    }
```
i-Alex, line 374: I would suggest to move it a separate utility class into `com.horizen.account.utils` package.

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
alsala: They are used in json representation, I added a comment

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
alsala: Ok, moved in a new EthereumTransactionUtils class in the package

JcBernack (approved):

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -124,16 +137,33 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+            val response = if (isEIP155) {
+                val tmpTx = new EthereumTransaction(
+                  destAddress,
+                  nonce,
+                  gasPrice,
+                  gasLimit,
+                  valueInWei,
+                  "",
+                  new SignatureData(
+                    EthereumTransaction.convertToBytes(params.chainId),
```
JcBernack, line 151: moved to `EthereumTransactionUtils`

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
alsala: Added, thanks!

i-Alex (approved):

JcBernack (approved):

PR #483, opened 2022-08-02
RPC calls to fake contracts

JcBernack:
- use message processors instead of `Evm.apply` to allow calls to fake contracts
- some refactoring


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -59,57 +54,75 @@ class EthService(val stateView: AccountStateView, val nodeView: CurrentView[Acco
...
-    transactions.foreach(transaction => {
-      if (hydratedTx) {
-        val receipt = stateView.getTransactionReceipt(Numeric.hexStringToByteArray(Numeric.cleanHexPrefix(transaction.id)))
-        if (!receipt.isEmpty) txList.append(new EthereumTransactionView(receipt.get, transaction))
-      } else txList.append(Numeric.prependHexPrefix(transaction.id))
-    })
-    new EthereumBlock(Numeric.prependHexPrefix(Integer.toHexString(nodeView.history.getBlockHeight(blockId).get())), Numeric.prependHexPrefix(blockId), txList.toList.asJava, block.get())
-  }
+    // TODO: why does this not compile?
+    //val transactions = block.get().transactions.collect { case tx: EthereumTransaction => tx }
```
JcBernack, line 58: I don't understand why this does not compile - shouldn't it be equivalent to the code below..?

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -59,57 +54,75 @@ class EthService(val stateView: AccountStateView, val nodeView: CurrentView[Acco
...
+
+  private def doCall(params: TransactionArgs, tag: Quantity) = {
+    using(getStateViewFromBlockById(getBlockIdByTag(tag.getValue))) { stateDbRoot =>
+      if (stateDbRoot == null) {
+        throw new IllegalArgumentException(s"Unable to get state for given tag: ${tag.getValue}")
+      }
+      val message = params.toMessage
+
+      // TODO: refactor to reuse parts of AccountStateView.applyTransaction?
+      val messageProcessors = Seq(
```
JcBernack, line 86: I would prefer to directly use the sequence of message processors of the current stateView, but that is not accessible currently. Suggestions?

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: Scala patterning matching is not available here, because the AccountTx and EthTx are not covariant.
It is expected.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: Let's implement `staticCall` method in the AccountStateView class, which has all needed elements inside.
EthService should just decide at which point to open the view and pass the message

JcBernack: AccountStateView now has `applyMessage` to be closer to geth naming.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -59,170 +56,130 @@ class EthService(val stateView: AccountStateView, val nodeView: CurrentView[Acco
...
-              result.evmError,
-              Numeric.toHexString(result.returnData)
-            )
-          )
-        }
-        if (result.returnData == null)
-          null
-        else
-          Numeric.toHexString(result.returnData)
+    val result = doCall(params, tag)
```
i-Alex, line 101: why we don't catch the potential exceptions here?
I guess it can lead to the internal server error (500) result.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -59,170 +56,130 @@ class EthService(val stateView: AccountStateView, val nodeView: CurrentView[Acco
...
+    if (result.hasReturnData)
+      Numeric.toHexString(result.returnData)
+    else
+      null
+  }
 
-      }
-    }
+  @RpcMethod("eth_estimateGas") def estimateGas(params: TransactionArgs /*, tag: Quantity*/): String = {
+    Numeric.toHexStringWithPrefix(doCall(params, new Quantity("latest")).gasUsed())
```
i-Alex, line 109: same here. Need to process the case when `doCall` throws exception and return the error result compatible to EthRpc.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -179,6 +179,16 @@ class AccountStateView(metadataStorageView: AccountStateMetadataStorageView,
     BigInteger.ZERO
   }
 
+  def applyMessage(message: Message): Option[ExecutionResult] = {
+    // TODO: some refactoring is required here, applyMessage should include:
+    //  - buying gas
+    //  - validations that the caller has enough funds for everything (gas and value transfer)
+    //  - processing message
+    //  - returning unused gas and refunds
```
i-Alex, line 187: This point requires a lot of attention.
Consider that `eth_estimateGas` uses staticCalll to predict the gasLimit, it expects the result before `refund`.
Otherwise, as for me, the gasLimit is going to be too small, so the real Tx will fail.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
JcBernack: The calling `RpcHandler` will catch any exceptions and handle them accordingly:
- If it is an `RpcException` it will unwrap and return the contained `RpcError` response.
- Any other exception will be returned as an `RpcError` with the code of "internal error" and the message of the exception.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
JcBernack: Interesting point. I had a look at the geth implementation and they do not return the gas used before refund.

They do a binary search to find a gasLimit that allows the transaction to succeed (with starting bounds `low=21000` and `high=args.gasLimit`).

Wallets will also not use the result of `eth_estimateGas` directly as the `gasLimit`, but rather add a buffer on top. In the source of MetaMask I found a [multiplier of `1.5`](https://github.com/MetaMask/metamask-extension/blob/c72199a1a6e4151c40c22f79d0f3b6ed7a2d59a7/app/scripts/controllers/transactions/tx-gas-utils.js#L86).

akrckn (approved):

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
i-Alex: Strange but ok.
Let's consider it in the current tasks

i-Alex (approved):

PR #484, opened 2022-08-03
Sdk 306 add run xml configuration for simple app

sgonta:
This configuration allows developer to run SimpleApp in the IntelliJ IDEA without need to configure manually


i-Alex (approved):
Good! Works for me

saratnt (approved):

PR #488, opened 2022-08-04
SDK-300: config param to customize forging frequency

paologalligit:


sdk/src/main/scala/com/horizen/SidechainSettings.scala
```diff
@@ -39,7 +39,9 @@ case class WithdrawalEpochCertificateSettings(submitterIsEnabled: Boolean,
 
 case class ForgerSettings(automaticForging: Boolean = false,
                           restrictForgers: Boolean = false,
-                          allowedForgersList: Seq[ForgerKeysData] = Seq())
+                          allowedForgersList: Seq[ForgerKeysData] = Seq(),
+                          forgingFrequencyInSeconds: Int = 120
```
i-Alex, line 43: fine for test. But not applicable for real environment, because must be set to some value for the whole network. If someone node will have a different value - will lead to divergence

i-Alex (approved):

sdk/src/main/scala/com/horizen/SidechainSettings.scala
paologalligit: yes, this is something just for testing, the `tps-tests` branch will never be merged into dev

PR #489, opened 2022-08-05
SDK-302: remove check on tx size per block

paologalligit:


sdk/src/main/scala/com/horizen/block/SidechainBlock.scala
```diff
@@ -158,8 +158,8 @@ class SidechainBlock(override val header: SidechainBlockHeader,
       case Failure(e) => throw e
     }
 
-    if(sidechainTransactions.size > SidechainBlock.MAX_SIDECHAIN_TXS_NUMBER)
-      throw new InvalidSidechainBlockDataException(s"SidechainBlock $id sidechain transactions amount exceeds the limit.")
+//    if(sidechainTransactions.size > SidechainBlock.MAX_SIDECHAIN_TXS_NUMBER)
```
i-Alex, line 161: the same can be done also for forger, otherwise Forger still has a limit of 1000000

PR #494, opened 2022-08-12
[SDK-221] Fixed Node termination in case of an unexpected shutdown

MarcoOl94:
## Description
This PR fixes the issue that causes the Node to not terminate correctly in case of an unexpected shutdown.

## Jira Ticket
[221](https://horizenlabs.atlassian.net/jira/software/c/projects/SDK/boards/14?modal=detail&selectedIssue=SDK-221&assignee=615f31c32f6aed0068c67291)

## Changes

- Removed the preloaded Log shutdown hook and added its termination inside our shutdown thread.
- Fixed the akka actors termination.

## Checks
- [✔] Project Builds

### DIFF
--- a/sdk/src/main/resources/log4j2.xml
+++ b/sdk/src/main/resources/log4j2.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="WARN" monitorInterval="30">
+<Configuration status="WARN" monitorInterval="30" shutdownHook="disable">
 
     <Properties>
         <Property name="LOG_PATTERN">[%-5level] %d{yyyy-MM-dd HH:mm:ss:SSS Z} [%28F:%-4L] [%t] %c - %msg%n</Property>
--- a/sdk/src/main/resources/log4j2.xml
+++ b/sdk/src/main/resources/log4j2.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="WARN" monitorInterval="30" shutdownHook="disable">
+<Configuration status="WARN" monitorInterval="30">
 
     <Properties>
         <Property name="LOG_PATTERN">[%-5level] %d{yyyy-MM-dd HH:mm:ss:SSS Z} [%28F:%-4L] [%t] %c - %msg%n</Property>
--- a/sdk/src/main/scala/com/horizen/SidechainApp.scala
+++ b/sdk/src/main/scala/com/horizen/SidechainApp.scala
@@ -58,7 +58,9 @@ import java.util.concurrent.TimeUnit
 import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 import scala.util.{Failure, Success, Try}
-
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.core.impl.Log4jContextFactory
+import org.apache.logging.log4j.core.util.DefaultShutdownCallbackRegistry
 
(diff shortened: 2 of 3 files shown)


sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -434,15 +423,28 @@ class SidechainApp @Inject()
...
+    field.setAccessible(true)
+    val hooks = field.get(null)
+
+    //Remove the Logger shutdown hook
+    val threadMap = hooks.asInstanceOf[JMap[Thread, Thread]].values()
+    threadMap.asScala.toArray.find(thread => thread.getName.equals("Logging-Cleaner")) match {
+      case Some(thread) =>
+        Runtime.getRuntime.removeShutdownHook(thread)
+      case None =>
+    }
```
i-Alex, line 438: Is it not enough just to set `shutdownHook="disable"` in the logger xml file?
I am a bit worried that we point to the name `Logging-Cleaner`, that can be changed inside the library.
What do you think about this comment https://stackoverflow.com/a/48063926 ?

sdk/src/main/resources/log4j2.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Configuration status="WARN" monitorInterval="30">
+<Configuration status="WARN" monitorInterval="30" shutdownHook="disable">
```
i-Alex, line 2: Is "disable" a valid value? I say "true|false" in the docs

MarcoOl94: I've tried both but this is the only value that works in my case

sdk/src/main/scala/com/horizen/SidechainApp.scala
MarcoOl94: It's not enought to use the config file.
With the approach described here  https://stackoverflow.com/a/48063926 the Logging-Cleaner Thread is not removed from the list of the ShutdowHook thread.
Beside that i agree with you that pointing to a specific Thread name is not a good practice.

I would like to add that also removing the `shutdownHook="disable"` breaks the correct behaviour.

saratnt (approved):

sdk/src/main/scala/com/horizen/SidechainApp.scala
i-Alex: within the current approach on the node startup I can see the batch of new warnings:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.horizen.SidechainApp (file:/home/oleksandr/Dev/Sidechains-SDK/sdk/target/classes/) to field java.lang.ApplicationShutdownHooks.hooks
WARNING: Please consider reporting this to the maintainers of com.horizen.SidechainApp
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
it doesn't seems good to me.

sdk/src/main/resources/log4j2.xml
i-Alex: right, it is specified to be set as disabled.

i-Alex (approved):

sdk/src/main/scala/com/horizen/SidechainApp.scala
i-Alex: Looks good with the new solution

saratnt (approved):

PR #495, opened 2022-08-12
Sdk 156 test script updates

rushby:
## Description
This PR introduces a new shell script for running the python tests named "run_sc_tests.sh".
The python Logger has replaced all occurrences of Print() in the tests, with logging messages now displaying in the console and the log file depending on the log level passed in.

## Jira Ticket
[156](https://horizenlabs.atlassian.net/browse/SDK-156)

## Changes

- Added run_sc_tests.sh script to run tests and output logs to file and console depend on log level
- Added setup_logger function to sc_test_framework
- Refactored some relative paths to use absolute
- Updated ./qa/README.md

## Checks
- [✔] Project Builds
- [✔] Updated documentation accordingly


qa/SidechainTestFramework/sc_test_framework.py
```diff
@@ -144,7 +158,7 @@ def main(self):
             if not os.path.isdir(self.options.tmpdir):
                 os.makedirs(self.options.tmpdir)
 
-            print("Initializing test directory "+self.options.tmpdir)
+            logging.info("Initializing test directory "+self.options.tmpdir)
```
i-Alex, line 161: I'm curious how it was working on Zen side with regular `print` method?
It is quite useful to see all the prints + node loges of level `ERROR` on console when running a single python test without `run_sc_tests.sh`.
Can we do something similar? So, to log to file only within the script.

qa/run_sc_tests.sh
```diff
@@ -0,0 +1,153 @@
...
+    *)
+      # unknown option/passOn
+      passOn+="${i} "
+      ;;
+  esac
+done
+
+#Run the tests
+testScripts=(
+    'websocket_server_fee_payments.py'
```
i-Alex, line 46: Pls, check the list of tests to be executed.
I saw the changes in the last PR, so no it is only one test in the list.

i-Alex, line 46: Another point, seems `websocket_server_fee_payments.py` is broken, the same for `websocket_server.py`

i-Alex (changes requested):
Script looks good!
Let's discuss a bit more the logger.

qa/run_sc_tests.sh
rushby: These tests appear to be broken prior to this change, I have only changed Print() to logging.info() in both these files.  I thought some tests may be failing and that was part of the reason why we wanted all the tests to continue even if one failed

qa/SidechainTestFramework/sc_test_framework.py
rushby: The parameters for setting log level already existed.  You can print to console by adding the --logconsolelevel=info parameter

qa/run_sc_tests.sh
i-Alex: seems they were broken for a while. I've tested with the current dev branch.
Will fix them as a separate task.

qa/SidechainTestFramework/sc_test_framework.py
i-Alex: Thanks, it works as expected. I missed that point

i-Alex (approved):

qa/run_sc_tests.sh
```diff
@@ -0,0 +1,196 @@
...
+      ;;
+    *)
+      # unknown option/passOn
+      passOn+="${i} "
+      ;;
+  esac
+done
+
+#Run the tests
+testScripts=(
```
paologalligit, line 45: What do you think about moving all the test scripts into a subfolder so we can iterate over all the files inside it instead of listing them all manually?

paologalligit (approved):
Looks very good, just left a minor comment

qa/run_sc_tests.sh
rushby: I agree with this Paolo, perhaps this can be changed in a future PR

VolodymyrBabaiev (approved):

PR #500, opened 2022-08-18
SDK-282 - consensus improvements

i-skrypnyk:
https://horizenlabs.atlassian.net/browse/SDK-282


sdk/src/test/scala/com/horizen/validation/ConsensusValidatorTest.scala
```diff
@@ -111,10 +116,11 @@ class ConsensusValidatorTest extends JUnitSuite with HistoryConsensusChecker {
       case nonExpected => assert(false, s"Got incorrect exception: ${nonExpected}")
     }
 
-    println("Test blockInFuture")
     // TODO: Current test is wrong, timestamp is hardcoded to have persistent tests, so now instead of "Block in future" we have "Whole epoch had been skipped".
     // TODO: restore and fix
-    /*val blockInFuture = generateBlockInTheFuture(lastGenerator)
+    /*
+    println("Test blockInFuture")
+    val blockInFuture = generateBlockInTheFuture(lastGenerator)
```
i-Alex, line 123: Let's try to fix this test.
Even if it is out of scope of the task.

sdk/src/test/scala/com/horizen/forge/ForgerTest.scala
```diff
@@ -0,0 +1,84 @@
...
+    when(params.sidechainGenesisBlockTimestamp).thenReturn(11)
+
+    val (forger, viewHolder) = prepareTestData(params, timeProvider)
+
+    forger ! StartForging
+    viewHolder.expectMsgType[LocallyGeneratedModifier[SidechainBlock]] //first right away
+    viewHolder.expectNoMessage(900.millis) //then ~1s pause
+    viewHolder.expectMsgType[LocallyGeneratedModifier[SidechainBlock]](1100.millis) //next in a second
+    viewHolder.expectNoMessage(4900.millis) //then ~10s pause
+    viewHolder.expectMsgType[LocallyGeneratedModifier[SidechainBlock]](5100.millis) //next in 10 seconds
```
i-Alex, line 53: Test idea seems good!
Can we reduce the time to make it pass faster? At the moment it is the longest test > 6s.

i-Alex, line 53: Why next in 10 seconds, but not in 5 seconds? Do we skip slot?

sdk/src/main/scala/com/horizen/consensus/package.scala
```diff
@@ -1,18 +1,21 @@
...
 import scorex.util.ModifierId
 import supertagged.TaggedType
 
+import java.math.{BigDecimal, BigInteger, MathContext}
+
 package object consensus {
   val merkleTreeHashLen: Int = 32
   val sha256HashLen: Int = 32
-  val consensusNonceLength: Int = Longs.BYTES
+  val consensusNonceLength: Seq[Int] = Seq(8, 32)
```
i-Alex, line 15: I think if we really need this.
Possibly, just for double check.
But in anyway, it is not very clear now from the parameter type how it is used. Let's change the name to something like `consensusNonceAllowedLengths`

i-Alex (changes requested):
Looks good!
Have you checked that automatic forging works in a real environment? At least a manual check.

sdk/src/main/scala/com/horizen/consensus/package.scala
```diff
@@ -53,15 +56,24 @@ package object consensus {
     val nonceBytes = nonce.consensusNonce
 
     val resBytes = Bytes.concat(slotNumberBytes, nonceBytes, consensusHardcodedSaltString)
-    VrfMessage @@ resBytes
+    if (resBytes.length > consensusPreForkLength) {
+      nonceBytes.update(nonceBytes.length - 1, (nonceBytes.last.abs % 64).toByte) //only 0 to 63 is allowed for last byte
```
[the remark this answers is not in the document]
i-skrypnyk: That's a great solution, thanks

sdk/src/main/scala/com/horizen/consensus/package.scala
```diff
@@ -53,15 +56,24 @@ package object consensus {
...
+      val digest = PoseidonHash.getInstanceConstantLength(3)
+      digest.update(FieldElementUtils.messageToFieldElement(slotNumberBytes))
+      digest.update(FieldElementUtils.messageToFieldElement(nonceBytes))
+      digest.update(FieldElementUtils.messageToFieldElement(consensusHardcodedSaltString))
+      VrfMessage @@ digest.finalizeHash().serializeFieldElement()
```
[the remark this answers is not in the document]
i-skrypnyk: done, thanks

sdk/src/test/scala/com/horizen/fork/ForkManagerTest.scala
```diff
@@ -1,25 +1,34 @@
...
 import org.scalatestplus.junit.JUnitSuite
 
 class ForkManagerTest extends JUnitSuite {
+
+  @Before
+  def init(): Unit = {
+    //this is to reset ForkManager from other tests
+    val c = ForkManager.getClass.getDeclaredConstructor()
+    c.setAccessible(true)
+    c.newInstance()
```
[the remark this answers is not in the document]
i-skrypnyk: good idea, thanks!

sdk/src/test/scala/com/horizen/forge/ForgerTest.scala
i-skrypnyk: reduced consensusSecondsInSlot to 2 seconds, which I think is the minimum value to make the test possible, cause we are working in seconds

i-skrypnyk: it was actually 5, I just started the test with 10 seconds consensusSecondsInSlot, then reduced it to 5 and forgot to fix the comment

sdk/src/main/scala/com/horizen/consensus/NonceConsensusEpochInfo.scala
```diff
@@ -29,5 +29,5 @@ case class NonceConsensusEpochInfo(consensusNonce: ConsensusNonce) extends Bytes
 object NonceConsensusEpochInfoSerializer extends ScorexSerializer[NonceConsensusEpochInfo]{
   override def serialize(obj: NonceConsensusEpochInfo, w: Writer): Unit = w.putBytes(obj.consensusNonce)
 
-  override def parse(r: Reader): NonceConsensusEpochInfo = NonceConsensusEpochInfo(ConsensusNonce @@ r.getBytes(consensusNonceLength))
+  override def parse(r: Reader): NonceConsensusEpochInfo = NonceConsensusEpochInfo(ConsensusNonce @@ r.getBytes(r.remaining))
```
VolodymyrBabaiev, line 32: Why don't we get fixed size of consensusNonce based on consensus epoch?

i-skrypnyk: There's not enough context in this place to determine current consensus epoch

i-skrypnyk: Also let me share my comment from discord:

> The reason for that is that nonce length is going to change after the fork(8 -> 32), and we have too many places where this class is used. To get current epoch and thus actual consensus nonce length, we'll have to add code to determine current epoch and fork to all of them. It's not looking good and greatly complicates the code. Also I can't also make this change in one place, cause around the fork node may encounter both new and old nonce values.

sdk/src/main/scala/com/horizen/consensus/package.scala
```diff
@@ -53,15 +56,37 @@ package object consensus {
...
+  }
+
+  private def generateHashAndCleanUp(elements: Array[Byte]*): Array[Byte] = {
+    val digest = PoseidonHash.getInstanceConstantLength(elements.length)
+    elements.foreach { message =>
+      val fieldElement = FieldElementUtils.messageToFieldElement(message)
+      digest.update(fieldElement)
+      fieldElement.freeFieldElement()
+    }
+    val result = digest.finalizeHash().serializeFieldElement()
```
i-Alex, line 79: here we need to free `digest.finalizeHash()` field element as well.

i-Alex (changes requested):
A minor thing left.

sdk/src/main/scala/com/horizen/consensus/NonceConsensusEpochInfo.scala
VolodymyrBabaiev: Double checked usage of NonceConsensusEpochInfoSerializer.

Yes, we can keep this implementation.

sdk/src/test/scala/com/horizen/forge/ForgerTest.scala
```diff
@@ -0,0 +1,83 @@
...
+    when(settings.websocket).thenReturn(webSocketSettings)
+    when(webSocketSettings.allowNoConnectionInRegtest).thenReturn(true)
+    when(settings.sparkzSettings).thenReturn(scorexSettings)
+    when(scorexSettings.restApi).thenReturn(restApiSettings)
+    when(restApiSettings.timeout).thenReturn(1.seconds)
+    val viewHolder = TestProbe()
+    val mainchainSynchronizer = mock[MainchainSynchronizer]
+    val companion = mock[SidechainTransactionsCompanion]
+
+    class TestForget(settings: SidechainSettings,
```
VolodymyrBabaiev, line 69: Please fix typo TestForge**r**

i-Alex (approved):

PR #502, opened 2022-08-19
Web socket settings and enabling of connections

sgonta:
Signed-off-by: sgonta <sgonta@horizenlabs.io>

### DIFF
--- a/examples/simpleapp/src/main/resources/sc_settings.conf
+++ b/examples/simpleapp/src/main/resources/sc_settings.conf
@@ -21,11 +21,15 @@ sparkz {
         agentName = "2-Hop"
     }
 
-    websocket {
+    websocketClient {
+        enabled = true
         address = "ws://localhost:8888"
         connectionTimeout = 100 milliseconds
         reconnectionDelay = 1 seconds
         reconnectionMaxAttempts = 1
+    }
+
+    websocketServer {
         wsServer = true
         wsServerPort = 0
     }
--- a/examples/simpleapp/src/main/resources/sc_settings.conf
+++ b/examples/simpleapp/src/main/resources/sc_settings.conf
@@ -22,7 +22,7 @@ sparkz {
     }
 
     websocketClient {
-        enabled = true
+        enabled = false
         address = "ws://localhost:8888"
         connectionTimeout = 100 milliseconds
         reconnectionDelay = 1 seconds
--- a/sdk/src/main/scala/com/horizen/SidechainApp.scala
+++ b/sdk/src/main/scala/com/horizen/SidechainApp.scala
@@ -26,6 +26,7 @@ import com.horizen.storage._
 import com.horizen.transaction._
 import com.horizen.utils.{BytesUtils, Pair}
 import com.horizen.wallet.ApplicationWallet
+import com.horizen.websocket.client.{DefaultWebSocketReconnectionHandler, WebSocketChannel, WebSocketCommunicationClient, WebSocketConnector, WebSocketConnectorImpl, WebSocketReconnectionHandler}
 import com.horizen.websocket.server.WebSocketServerRef
 import sparkz.core.api.http.ApiRoute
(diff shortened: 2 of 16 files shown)


sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -333,24 +333,29 @@ class SidechainApp @Inject()
...
+    if (connectorStarted.isSuccess)
+      communicationClient.setWebSocketChannel(webSocketConnector)
+    else if (sidechainSettings.withdrawalEpochCertificateSettings.submitterIsEnabled)
+      throw new RuntimeException("Unable to connect to websocket. Certificate submitter needs connection to Mainchain.")
+
+  } else {
+    log.info("Due to the settings, node is not enabled for connections.")
+  }
 
   // If the web socket connector can be started, maybe we would to associate a client to the web socket channel created by the connector
```
i-Alex, line 358: The comment is now detached from the code.
Must be put before the line `if (connectorStarted.isSuccess)`

sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -333,24 +333,29 @@ class SidechainApp @Inject()
...
-  val connectorStarted : Try[Unit] = webSocketConnector.start()
+    // Start the web socket connector
+    val connectorStarted : Try[Unit] = webSocketConnector.start()
+    if (connectorStarted.isSuccess)
+      communicationClient.setWebSocketChannel(webSocketConnector)
+    else if (sidechainSettings.withdrawalEpochCertificateSettings.submitterIsEnabled)
+      throw new RuntimeException("Unable to connect to websocket. Certificate submitter needs connection to Mainchain.")
+
+  } else {
+    log.info("Due to the settings, node is not enabled for connections.")
```
i-Alex, line 355: Honestly, it is not very clear for which exactly connections. We have a multiple different one.
Let's say something like "Websocket client is disabled."

sdk/src/main/scala/com/horizen/SidechainSettings.scala
```diff
@@ -7,11 +7,15 @@ import scala.concurrent.duration.FiniteDuration
...
-case class WebSocketSettings(address: String,
+case class WebSocketClientSettings(address: String,
                              connectionTimeout: FiniteDuration,
                              reconnectionDelay: FiniteDuration,
                              reconnectionMaxAttempts: Int,
                              allowNoConnectionInRegtest: Boolean = true, // In Regtest allow to forge new blocks without connection to MC node, for example.
+                             enabled: Boolean
+                             )
+
+case class WebSocketServerSettings(address: String,
```
i-Alex, line 18: `address: String,` should be removed. It is a `client` related

examples/simpleapp/src/main/resources/sc_settings.conf
```diff
@@ -22,11 +22,15 @@ scorex {
         agentName = "2-Hop"
     }
 
-    websocket {
+    websocketClient {
+        enabled = true
```
i-Alex, line 26: Here we may setup it to `false` because it is the config file with already predefined genesi info and no expected MC connection.

i-Alex (changes requested):
Looks good!

There are several UTs failing in SidechainNodeViewHolderActorTest.scala
Need to update some mocks.
As an additional check you may run SimpleApp with the conf file `examples/simpleapp/src/main/resources/sc_settings.conf`. Also check the Python tests results.

examples/simpleapp/src/main/resources/sc_settings.conf
sgonta: image

i-Alex (approved):

Anon79:
> Signed-off-by: sgonta [sgonta@horizenlabs.io](mailto:sgonta@horizenlabs.io)

SDK-40-websocket-client-improvements

PR #514, opened 2022-08-29
Bloom filter implementation

GitMark0:
Functions: 

- getBloomFilterValues - takes in an array of bytes, eg. a topic and returns byte indexes and values that the bytes should be set to in the bloom filter.
- addDataToBloomFilter - takes in a bloom filter instance and data, eg. a topic and returns a modified bloom filter array of bytes
- createBloomFilter - initializes an empty 256 size byte array and populates it based on the addresses from topics and topic data

### DIFF
--- a/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
+++ b/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
@@ -271,6 +271,7 @@ object EthereumConsensusDataReceipt {
     var bloomFilter = Array.fill[Byte](256)(0)
 
     evmLog.foreach(log => {
+      bloomFilter = addDataToBloomFilter(bloomFilter, log.address.toBytes)
       log.topics.foreach(topic => {
         bloomFilter = addDataToBloomFilter(bloomFilter, topic.toBytes)
       })
--- a/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
+++ b/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
@@ -101,8 +101,7 @@ case class EthereumConsensusDataReceipt(
     logsString = logsString.concat("}")
 
     var logsBloomStr = "null"
-    if (logsBloom != null)
-      logsBloomStr = BytesUtils.toHexString(logsBloom.getBloomFilter())
+    logsBloomStr = BytesUtils.toHexString(logsBloom.getBloomFilter())
 
     String.format(
       s"EthereumReceipt (receipt consensus data) { txType=$getTxType, status=$getStatus, cumGasUsed=$cumulativeGasUsed, logs=$logsString, logsBloom=$logsBloomStr}"
--- a/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
+++ b/sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
@@ -1,7 +1,10 @@
 package com.horizen.account.receipt
 
 import com.horizen.account.receipt.EthereumConsensusDataReceipt.ReceiptStatus.ReceiptStatus
-import com.horizen.account.receipt.EthereumConsensusDataReceipt.{ReceiptStatus, createLogsBloom}
+import com.horizen.account.receipt.EthereumConsensusDataReceipt.{
+  ReceiptStatus,
+  createBloomFilter
+}
 import com.horizen.account.receipt.ReceiptTxType.ReceiptTxType
 import com.horizen.evm.interop.EvmLog
 import com.horizen.utils.BytesUtils
@@ -14,10 +17,12 @@ import java.util
 import scala.collection.mutable.ListBuffer
 
 case class EthereumConsensusDataReceipt(
-                                         transactionType: Int,
-                                         status: Int,
-                                         cumulativeGasUsed: BigInteger,
-                                         logs: Seq[EvmLog], logsBloom: Array[Byte]) {
+    transactionType: Int,
+    status: Int,
+    cumulativeGasUsed: BigInteger,
+    logs: Seq[EvmLog],
+    logsBloom: Array[Byte]
+) {
 
   /*  From yellow paper
       ----------------------
@@ -28,16 +33,28 @@ case class EthereumConsensusDataReceipt(
       and the Bloom filter composed from information in those logs, Rb
    */
 
-  def this(transactionType: Int,
-           status: Int,
-           cumulativeGasUsed: BigInteger,
-           logs: Seq[EvmLog]) {
+  def this(
+      transactionType: Int,
+      status: Int,
+      cumulativeGasUsed: BigInteger,
+      logs: Seq[EvmLog]
+  ) {
 
-    this(transactionType, status, cumulativeGasUsed, logs, createLogsBloom(logs))
+    this(
+      transactionType,
+      status,
+      cumulativeGasUsed,
+      logs,
+      createBloomFilter(logs)
+    )
   }
 
-  require(transactionType >= 0 && transactionType <= ReceiptTxType.DynamicFeeTxType.id)
-  require(status == ReceiptStatus.SUCCESSFUL.id || status == ReceiptStatus.FAILED.id)
+  require(
+    transactionType >= 0 && transactionType <= ReceiptTxType.DynamicFeeTxType.id
+  )
+  require(
+    status == ReceiptStatus.SUCCESSFUL.id || status == ReceiptStatus.FAILED.id
+  )
 
   def getTxType: ReceiptTxType = {
     transactionType match {
@@ -91,7 +108,9 @@ case class EthereumConsensusDataReceipt(
     if (logsBloom != null)
       logsBloomStr = BytesUtils.toHexString(logsBloom)
 
-    String.format(s"EthereumReceipt (receipt consensus data) { txType=$getTxType, status=$getStatus, cumGasUsed=$cumulativeGasUsed, logs=$logsString, logsBloom=$logsBloomStr}")
+    String.format(
+      s"EthereumReceipt (receipt consensus data) { txType=$getTxType, status=$getStatus, cumGasUsed=$cumulativeGasUsed, logs=$logsString, logsBloom=$logsBloomStr}"
+    )
   }
 }
 
@@ -111,7 +130,6 @@ object EthereumConsensusDataReceipt {
     val FAILED, SUCCESSFUL = Value
   }
 
-
   def decodeLegacy(rlpData: Array[Byte]): EthereumConsensusDataReceipt = {
     val rlpList = RlpDecoder.decode(rlpData)
     val values = rlpList.getValues.get(0).asInstanceOf[RlpList]
@@ -124,7 +142,8 @@ object EthereumConsensusDataReceipt {
       }
       1
     }
-    val cumulativeGasUsed = values.getValues.get(1).asInstanceOf[RlpString].asPositiveBigInteger
+    val cumulativeGasUsed =
+      values.getValues.get(1).asInstanceOf[RlpString].asPositiveBigInteger
     val logsBloom = values.getValues.get(2).asInstanceOf[RlpString].getBytes
     val logList = values.getValues.get(3).asInstanceOf[RlpList]
     val logs = new ListBuffer[EvmLog]
@@ -132,17 +151,32 @@ object EthereumConsensusDataReceipt {
     if (logsListSize > 0) {
       // loop on list and decode all logs
       for (i <- 0 until logsListSize) {
-        val log = EvmLogUtils.rlpDecode(logList.getValues.get(i).asInstanceOf[RlpList])
+        val log =
+          EvmLogUtils.rlpDecode(logList.getValues.get(i).asInstanceOf[RlpList])
         logs += log
       }
     }
-    EthereumConsensusDataReceipt(ReceiptTxType.LegacyTxType.id, status, cumulativeGasUsed, logs, logsBloom)
+    EthereumConsensusDataReceipt(
+      ReceiptTxType.LegacyTxType.id,
+      status,
+      cumulativeGasUsed,
+      logs,
+      logsBloom
+    )
   }
 
-
-  def decodeTyped(rt: Int, rlpData: Array[Byte]): EthereumConsensusDataReceipt = {
+  def decodeTyped(
+      rt: Int,
+      rlpData: Array[Byte]
+  ): EthereumConsensusDataReceipt = {
     val r = decodeLegacy(rlpData)
-    EthereumConsensusDataReceipt(rt, r.status, r.cumulativeGasUsed, r.logs, r.logsBloom)
+    EthereumConsensusDataReceipt(
+      rt,
+      r.status,
+      r.cumulativeGasUsed,
+      r.logs,
+      r.logsBloom
+    )
   }
 
   def rlpDecode(rlpData: Array[Byte]): EthereumConsensusDataReceipt = {
@@ -154,10 +188,8 @@ object EthereumConsensusDataReceipt {
     if (b0 == 1 || b0 == 2) {
       val rt = ReceiptTxType(b0).id
       decodeTyped(rt, util.Arrays.copyOfRange(rlpData, 1, rlpData.length))
-    }
-    else decodeLegacy(rlpData)
+    } else decodeLegacy(rlpData)
   }
-
 
   def rlpEncode(r: EthereumConsensusDataReceipt): Array[Byte] = {
     val values = asRlpValues(r)
@@ -165,7 +197,11 @@ object EthereumConsensusDataReceipt {
     val encoded = RlpEncoder.encode(rlpList)
     if (!(r.getTxType == ReceiptTxType.LegacyTxType)) {
       // add byte for versioned type support
-      ByteBuffer.allocate(encoded.length + 1).put(r.getTxType.id.toByte).put(encoded).array
+      ByteBuffer
+        .allocate(encoded.length + 1)
+        .put(r.getTxType.id.toByte)
+        .put(encoded)
+        .array
     } else {
       encoded
     }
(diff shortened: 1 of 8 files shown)


sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+  }
+
+  def createBloomFilter(evmLog: Seq[EvmLog]): Array[Byte] = {
+    var bloomFilter = Array.fill[Byte](256)(0)
+
+    evmLog.foreach(log => {
+      log.topics.foreach(topic => {
+        bloomFilter = addDataToBloomFilter(bloomFilter, topic.toBytes)
+      })
+    })
```
i-Alex, line 277: here we miss `log.address` part to be added to the bloom filter.
In the original code for every log we first add address then all the topics one by one.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+  def bytesToShort(byte1: Byte, byte2: Byte): Short = {
+    ByteBuffer
+      .allocate(2)
+      .order(ByteOrder.BIG_ENDIAN)
+      .put(byte1)
+      .put(byte2)
+      .getShort(0)
+  }
+
+  def getBloomFilterValues(data: Array[Byte]) = {
```
i-Alex, line 237: please, specify the return type value for every method.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+    // Copy first 3 pairs of bytes to hashBuffer
+    val hashBuffer = Keccak256.hash(data)
+
+    // Calculate byte values
+    val v1 = 1 << (hashBuffer(1) & 0x7)
+    val v2 = 1 << (hashBuffer(3) & 0x7)
+    val v3 = 1 << (hashBuffer(5) & 0x7)
```
i-Alex, line 244: For every value here and in other places please setup explicitly the types.
It seems very critical in the code like this. Otherwise, whatever change in the algorithm may lead to the  different type assigned to the value, like int\long\short\etc. And we will not be able to detect it properly on compilation time. So may lead to unpredictable result.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+    val (i1, v1, i2, v2, i3, v3) = getBloomFilterValues(data)
+
+    bloomFilter(i1) = (bloomFilter(i1) | v1).toByte
+    bloomFilter(i2) = (bloomFilter(i2) | v2).toByte
+    bloomFilter(i3) = (bloomFilter(i3) | v3).toByte
+
+    bloomFilter
+  }
+
+  def createBloomFilter(evmLog: Seq[EvmLog]): Array[Byte] = {
```
i-Alex, line 270: I suggest to create a separate class `BloomFilter` that will encapsulate all the logic of bloom filter inside:
```
class BloomFilter {
  private val bloomFilter: Array[Byte] = new Array[Byte](BLOOM_FILTER_LENGTH)

  def add(data: Array[Byte]): Unit = {
    ///... like addDataToBloomFilter
  }
  def getBloomFilter: Array[Byte] = { /* copy of bloomFilter array*/ }
  private def bytesToShort(byte1: Byte, byte2: Byte): Short = { ... }

  private def getBloomFilterValues(data: Array[Byte]) = { ... }
}
object BloomFilter {
  def apply(evmLog: Seq[EvmLog]): BloomFilter =  { /* create filter and add all the logs */ }
  def apply(evmLog: Seq[EthereumReceipt]): BloomFilter = { /* create filter and add all the receipts' logs */ }
  
  val BLOOM_FILTER_LENGTH: Int = 256
}
```

sdk/src/test/scala/com/horizen/account/receipt/EthereumConsensusDataReceiptTest.scala
```diff
@@ -0,0 +1,32 @@
...
+import org.junit._
+import org.scalatestplus.junit.JUnitSuite
+import org.scalatestplus.mockito._
+
+class EthereumConsensusDataReceiptTest
+    extends JUnitSuite
+    with MockitoSugar
+    with ReceiptFixture {
+
+  @Test def bloomFilterTest(): Unit = {
```
i-Alex, line 15: We need more tests, more regression data.
I believe that Geth have enough tests for bloom filters we need to reuse all of them.

i-Alex (changes requested):
Next step is to start using bloom filters in `EthereumConsensusDataReceipt`, see `logsBloom` field. After we may modify AccountBlockHeader to keep `bloomFilter` as a field. That field should have the same name as in Geth, must be calculated by the Forger and validated by the State in the end of applyModifier() method.
After this, we should check that bloom filter is properly displayed in RPC methods and all the related methods are implemented in general.
Last step is to modify STF tests and verify bloom filter related part.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,59 @@ object EthereumConsensusDataReceipt{
...
-                    i2 := BloomByteLength - uint((binary.BigEndian.Uint16(hashbuf[2:])&0x7ff)>>3) - 1
-                    i3 := BloomByteLength - uint((binary.BigEndian.Uint16(hashbuf[4:])&0x7ff)>>3) - 1
-
-                    return i1, v1, i2, v2, i3, v3
-                }
-
-             */
-    // TODO shall we use libevm implementation?
-    new Array[Byte](256)
+  def bytesToShort(byte1: Byte, byte2: Byte): Short = {
```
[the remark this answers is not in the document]
JcBernack: could also be replaced by our existing `BytesUtils.getShort()`

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+      bloomFilter: Array[Byte],
+      data: Array[Byte]
+  ): Array[Byte] = {
+    val (i1, v1, i2, v2, i3, v3) = getBloomFilterValues(data)
+
+    bloomFilter(i1) = (bloomFilter(i1) | v1).toByte
+    bloomFilter(i2) = (bloomFilter(i2) | v2).toByte
+    bloomFilter(i3) = (bloomFilter(i3) | v3).toByte
+
+    bloomFilter
```
JcBernack, line 267: Just a general comment:
From the outside this method looks like it doesn't modify `bloomFilter` as it returns another byte array, while it does in fact mutate the array that was passed in. Imho it should be one or the other, either return a mutated copy, or mutate and don't return anything.

In this case we really want to mutate, so I would suggest to just remove the return value. See also the BloomFilter class suggestion by @i-Alex where `add()` does not return anything.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
JcBernack: Yes, this please. I just have two things to add:
- Also add `Test(data: Array[Byte]): Boolean` to check presence of data in the bloom filter. See geth implementation below.
- We should also be able to reconstruct a BloomFilter instance from the resulting data like `new BloomFilter(bloomFilterArray)` to be able to use the Test method.
```
// Test checks if the given topic is present in the bloom filter
func (b Bloom) Test(topic []byte) bool {
	i1, v1, i2, v2, i3, v3 := bloomValues(topic, make([]byte, 6))
	return v1 == v1&b[i1] &&
		v2 == v2&b[i2] &&
		v3 == v3&b[i3]
}
```

sdk/src/test/scala/com/horizen/account/receipt/EthereumConsensusDataReceiptTest.scala
JcBernack: There are not too many tests in geth regarding bloom filter, but I guess they are enough as a slight variance would immediately fail the test. I mean, their main test just adds 100 elements to the bloom filter, gets the resulting data and compares its hash to a precalculated value. This would fail if a single bit of the bloom filter is off.

We should just do the same, preferably exactly the same, because then we can also reuse their precalculated hash value and make sure it matches.

@GitMark0 see `TestBloom` and `TestBloomExtensively` in geth source: `core/types/bloom9_test.go`.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
```diff
@@ -188,54 +225,57 @@ object EthereumConsensusDataReceipt{
...
+
+  def addDataToBloomFilter(
+      bloomFilter: Array[Byte],
+      data: Array[Byte]
+  ): Array[Byte] = {
+    val (i1, v1, i2, v2, i3, v3) = getBloomFilterValues(data)
+
+    bloomFilter(i1) = (bloomFilter(i1) | v1).toByte
+    bloomFilter(i2) = (bloomFilter(i2) | v2).toByte
+    bloomFilter(i3) = (bloomFilter(i3) | v3).toByte
```
JcBernack, line 265: You should be able to shorten this: `bloomFilter(i1) |= v1`

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
GitMark0: > I suggest to create a separate class `BloomFilter` that will encapsulate all the logic of bloom filter inside:
> 
> ```
> class BloomFilter {
>   private val bloomFilter: Array[Byte] = new Array[Byte](BLOOM_FILTER_LENGTH)
> 
>   def add(data: Array[Byte]): Unit = {
>     ///... like addDataToBloomFilter
>   }
>   def getBloomFilter: Array[Byte] = { /* copy of bloomFilter array*/ }
>   private def bytesToShort(byte1: Byte, byte2: Byte): Short = { ... }
> 
>   private def getBloomFilterValues(data: Array[Byte]) = { ... }
> }
> object BloomFilter {
>   def apply(evmLog: Seq[EvmLog]): BloomFilter =  { /* create filter and add all the logs */ }
>   def apply(evmLog: Seq[EthereumReceipt]): BloomFilter = { /* create filter and add all the receipts' logs */ }
>   
>   val BLOOM_FILTER_LENGTH: Int = 256
> }
> ```

I've tried creating two different apply methods as you've suggested, but Scala doesn't allow me to create them because both arguments are of type Seq (I guess that Scala only checks if it is Seq, not Seq[Type]) so it doesn't differentiate them. I also tried creating two auxiliary constructors based on your approach, I get the same problem.

JcBernack: Ah yes, that doesn't work, your suggestion is ok.

sdk/src/main/scala/com/horizen/account/receipt/EthereumConsensusDataReceipt.scala
GitMark0: This doesn't work unfortunately... https://stackoverflow.com/a/24780501

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  private def getBloomFilterValues(
+      data: Array[Byte]
+  ): (Array[Int], Array[Int]) = {
+    val hashBuffer = Keccak256.hash(data)
+
+    val bloomFilterValues: Array[Int] =
+      Array(1, 3, 5).map(e => 1 << (hashBuffer(e) & 0x7))
+
+    val bloomFilterIndexes: Array[Int] = Array((0, 1), (2, 3), (4, 5)).map(e =>
```
JcBernack, line 64: Personally, I'm not a fan of using these interator-map-constructs for something that will only ever be executed exactly three times, It doesn't improve readability. Not sure if there is any performance difference, so I guess it comes down to personal preference :wink:

sdk/src/test/scala/com/horizen/account/receipt/EthereumConsensusDataReceiptTest.scala
JcBernack: Geth tests are included now, I think this is ok :+1:

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
GitMark0: In my opinion the first iterator looks cleaner, second one is kinda more confusing because of the tuples. I actually think that there is performance degradation regarding the second one because the array needs to be created just to retrieve the indexes of hashBuffer that need to be set. Second one was created just to continue the coding style based on the first one. I'll revert the changes to make the code more readable and optimized 😄

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
```diff
@@ -0,0 +1,105 @@
...
+          (bloomFilter(bloomFilterIndex) | bloomFilterValues(i)).toByte
+      })
+  }
+
+  def setBytes(data: Array[Byte]): Unit = {
+    require(bloomFilter.length == BLOOM_FILTER_LENGTH)
+
+    bloomFilter.zipWithIndex
+      .foreach({ case (bloomByte, i) =>
+        bloomFilter(i) = (bloomByte | data(i)).toByte
```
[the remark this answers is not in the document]
GitMark0: I see that the GETH implementation of the BloomFilter has this `setBloom` function where you can explicitly `OR` the existing bloomFilter with the provided 256 byte array. That is why I assumed that they maybe use it somewhere else in the code so I decided to implement it also. If I delete it, we can only do that operation once, in the constructor.

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
```diff
@@ -0,0 +1,105 @@
...
+    log.topics.foreach(topic => addBytesToBloomFilter(topic.toBytes))
+  }
+
+  def addBytesToBloomFilter(
+      data: Array[Byte]
+  ): Unit = {
+    val (bloomFilterIndexes, bloomFilterValues) = getBloomFilterValues(data)
+
+    bloomFilterIndexes.zipWithIndex
+      .foreach({ case (bloomFilterIndex, i) =>
```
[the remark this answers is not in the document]
GitMark0: I'm not sure that Scala can do the destructuring in that way, that is why the `case` is there. From code snippets I've seen that this the way to "destructure" in Scala.

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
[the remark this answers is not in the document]
GitMark0: I don't know either 😄 , yes it is possible to combine them.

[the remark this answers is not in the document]
JcBernack: Ok then this operation at least is meaningful, whether we use or not. All good then :+1:

JcBernack (approved):

akrckn (approved):
If the tasks mentioned by oleksandr are not part of this PR and task, it seems fine for me - else there would be some more work left.

sdk/src/test/scala/com/horizen/account/receipt/EthereumReceiptTest.scala
```diff
@@ -187,7 +187,7 @@ class EthereumReceiptTest
       //println("cumGas =" + cumGas.toString())
       val logsBloom = new Array[Byte](256)
       val logs = new ListBuffer[EvmLog]
-      receipts(i) = new EthereumConsensusDataReceipt(txType, status, cumGas, logs, logsBloom)
+      receipts(i) = new EthereumConsensusDataReceipt(txType, status, cumGas, logs,new LogsBloom(logsBloom))
```
i-Alex, line 190: We can remove `val logsBloom = new Array[Byte](256)`
Then use another constructor `receipts(i) = new EthereumConsensusDataReceipt(txType, status, cumGas, logs)` so to be sure that bloom filter is consistent to `logs`

sdk/src/test/scala/com/horizen/account/receipt/ReceiptFixture.scala
```diff
@@ -42,7 +42,7 @@ trait ReceiptFixture {
     } else {
       new Array[Byte](0)
     }
-    val consensusDataReceipt = new EthereumConsensusDataReceipt(txType, 1, BigInteger.valueOf(1000), logs, new Array[Byte](256))
+    val consensusDataReceipt = new EthereumConsensusDataReceipt(txType, 1, BigInteger.valueOf(1000), logs, new LogsBloom())
```
i-Alex, line 45: here `new LogsBloom()` is not consistent to `logs`. Use the same constructor without `...new LogsBloom()...`

sdk/src/test/scala/com/horizen/account/receipt/ReceiptFixture.scala
```diff
@@ -57,7 +57,7 @@ trait ReceiptFixture {
     val logs = new ListBuffer[EvmLog]
     for (_ <- 1 to num_logs)
       logs += createTestEvmLog
-    new EthereumConsensusDataReceipt(txType, 1, BigInteger.valueOf(1000), logs, new Array[Byte](256))
+    new EthereumConsensusDataReceipt(txType, 1, BigInteger.valueOf(1000), logs, new LogsBloom())
```
i-Alex, line 60: same here

i-Alex (approved):

PR #516, opened 2022-08-30
add base fee calculation

akrckn:


sdk/src/main/java/com/horizen/account/utils/Account.java
```diff
@@ -2,10 +2,13 @@
 
 import org.web3j.crypto.Keys;
 
+import java.math.BigInteger;
+
 public final class Account {
 
     public static final int ADDRESS_SIZE = Keys.ADDRESS_LENGTH_IN_HEX / 2;
     public static final int GAS_LIMIT = 30000000;
+    public static final BigInteger INITIAL_BASE_FEE = BigInteger.valueOf(1000000000);
```
i-Alex, line 11: We need to choose the better place for the GAS_LIMIT and INITIAL_BASE_FEE constants.
They are related to block, not to account.
Let's create a scala object in utils package, for example.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -312,7 +322,7 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
   @RpcMethod("eth_gasPrice")
   def gasPrice: Quantity = {
     applyOnAccountView { nodeView =>
-      getStateViewAtTag(nodeView, "latest") { tagStateView => new Quantity(tagStateView.getBaseFee) }
+      getBlockAtTag(nodeView, "latest") { block => new Quantity(block.header.baseFee)}
```
i-Alex, line 325: So here we want to predict the min gasPrice for given point.
For example, "latest", it seems we need to return not the block baseFee, but the next potential block baseFee.
We may use `def calculateBaseFee(nodeView: View, parentId: BlockId)`

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,6 +78,41 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
     }
   }
 
+  def calculateBaseFee(nodeView: View, parentId: BlockId): BigInteger = {
```
i-Alex, line 81: 1) we may use `history` instead of `nodeView`
2) let's put this method as a separate utility method. Since it can be used in other places: here, RPC, API, State

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,6 +78,41 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+    // If the current block is the first block, return the InitialBaseFee.
+    if (nodeView.history.getCurrentHeight < 2) {
```
i-Alex, line 83: we may use `parentId != params.sidechainGenesisBlockParentId` to check if the new block is genesis block.
we can't rely on the `nodeView.history.getCurrentHeight` results here.
But in general, it is even not a case, since during the verification phase in the validator you may check if `history.isGenesisBlock(...)` before running `calculateBaseFee`

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -214,7 +250,7 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
       new Array[Byte](MerkleTree.ROOT_HASH_LENGTH),
       new Array[Byte](MerkleTree.ROOT_HASH_LENGTH), // stateRoot TODO add constant
       new AddressProposition(new Array[Byte](Account.ADDRESS_SIZE)), // forgerAddress: PublicKeySecp256k1Proposition TODO add constant,
-      Long.MaxValue,
+      Account.INITIAL_BASE_FEE,
```
i-Alex, line 253: need to pass `calculateBaseFee(nodeView, parentId)` to guarantee that block size is correct (at least not smaller). Or we can reserve again `BigInteger(Long.MaxValue).getBytes().size()`

i-Alex (changes requested):

i-Alex (changes requested):
At the moment we do not verify if the incoming AccountBlock has a valid baseFee compared to the parent.
For this we need to define a new `HistoryBlockValidator` (see `ConsensusValidator` and `ChainIdBlockSemanticValidator` as an example). And then register it in the `AccountSidechainNodeViewHolder`.
Addition to that, we may add UTs coverage for the newly created class.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
akrckn: I chose the first suggestion, because as you wrote above, we can use this function in several places

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -112,8 +113,10 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
   private def doCall[A](nodeView: NV, params: TransactionArgs, tag: String)(fun: (Array[Byte], AccountStateView) ⇒ A): A = {
     getStateViewAtTag(nodeView, tag) { tagStateView =>
       try {
-        val msg = params.toMessage(tagStateView.getBaseFee)
-        fun(tagStateView.applyMessage(msg, new GasPool(msg.getGasLimit)), tagStateView)
+        getBlockAtTag(nodeView, tag) { block =>
+          val msg = params.toMessage(block.header.baseFee)
```
JcBernack, line 117: We will have conflicts here. The base fee is in the BlockContext now.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -295,6 +298,13 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
     }
   }
 
+  private def getBlockAtTag[A](nodeView: NV, tag: String)(fun: AccountBlock ⇒ A): A = {
+    nodeView.history.getBlockById(getBlockIdByTag(nodeView, tag)).asScala match {
+      case None => throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, "Invalid block tag parameter."))
+      case Some(block) => fun(block)
```
JcBernack, line 304: This function style structure is useful in the other method because want to wrap the `fun` execution in a `using` to make sure we properly close the view. Here you could also just return the block.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,6 +78,41 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+      val blockHeader = block.header
+      val gasTarget = blockHeader.gasLimit / 2
+
+        // If the parent gasUsed is the same as the target, the baseFee remains unchanged
+      if (blockHeader.gasUsed == gasTarget) {
+        return blockHeader.baseFee
+      }
+
+      if (blockHeader.gasUsed > gasTarget) {
+        // If the parent block used more gas than its target, the baseFee should increase by 12.5%
```
JcBernack: The comment is a bit misleading, the 12.5% is the maximum change.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,6 +78,41 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+        baseFeeInc = baseFeeInc.divide(BigInteger.valueOf(gasTarget))
+        baseFeeInc = baseFeeInc.divide(BigInteger.valueOf(8))
+        blockHeader.baseFee.add(baseFeeInc.max(BigInteger.ONE))
+      } else {
+        // Otherwise if the parent block used less gas than its target, the baseFee should decrease by 12.5%
+        var baseFeeDec: BigInteger = BigInteger.valueOf(gasTarget - blockHeader.gasUsed)
+        baseFeeDec = baseFeeDec.multiply(blockHeader.baseFee)
+        baseFeeDec = baseFeeDec.divide(BigInteger.valueOf(gasTarget))
+        baseFeeDec = baseFeeDec.divide(BigInteger.valueOf(8))
+        blockHeader.baseFee.subtract(baseFeeDec).max(BigInteger.ONE)
```
JcBernack: Should be `BigInteger.ZERO` here, according to GETH implementation.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
akrckn: afaik we will need this + the result of SuggestTipCap - so we should be fine within this PR, no?

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
akrckn: BigInteger(Long.MaxValue) would not be the max value, I changed it to a BigInteger with 256-bit

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
akrckn: Yes, I already saw that and changed

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
akrckn: I know, we discussed to change this later to another value, but I just wanted to prevent it to become 0 for now.

sdk/src/main/scala/com/horizen/account/validation/BaseFeeBlockValidator.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.compat.java8.OptionConverters.RichOptionalGeneric
+import scala.util.Try
+
+case class BaseFeeBlockValidator() extends HistoryBlockValidator[SidechainTypes#SCAT, AccountBlockHeader, AccountBlock, AccountHistoryStorage, AccountHistory] {
+
+  override def validate(block: AccountBlock, history: AccountHistory): Try[Unit] = Try {
+    val baseFee = block.header.baseFee
+    history.getBlockById(block.header.parentId).asScala match {
+      case None =>
+        if (history.isGenesisBlock(block.id) && baseFee.compareTo(FeeUtils.INITIAL_BASE_FEE) != 0)
```
JcBernack, line 20: If we do not find a block with given `parentId` the current block *must* be the genesis block, right?
Does it make sense to change this to:
- check if block is not genesis block => exception
- check if base fee differs from `INITIAL_BASEFEE` => exception

sdk/src/test/scala/com/horizen/account/validation/BaseFeeBlockValidatorTest.scala
```diff
@@ -0,0 +1,49 @@
...
+    // Test 2: Validation exception expected, block is genesis block and does not have initial base fee
+    assertThrows[InvalidBaseFeeException] {
+      BaseFeeBlockValidator().validate(mockedBlock, mockedHistory).get
+    }
+
+    mockHelper = AccountMockDataHelper(false)
+    mockedHistory = mockHelper.getMockedAccountHistory(Option.apply(mockedGenesisBlock).asJava)
+    mockedBlock = mockHelper.getMockedBlock(FeeUtils.INITIAL_BASE_FEE)
+
+    // Test 3: Successful validation, block is empty, therefore base fee did not change
```
JcBernack, line 35: This comment is a bit misleading. You construct two blocks with the same base fee, so validation should succeed.
If these were real blocks the base should decrease by 12.5% after an empty block, right?

JcBernack (changes requested):

sdk/src/test/scala/com/horizen/account/validation/BaseFeeBlockValidatorTest.scala
akrckn: Yes, you are right - I mean gasUsed is the same, therefore same baseFee. Just saw that the other comment below is false, too. 🙈

JcBernack (approved):

sdk/src/main/scala/com/horizen/account/validation/BaseFeeBlockValidator.scala
```diff
@@ -0,0 +1,32 @@
...
+        if (!history.isGenesisBlock(block.id)) throw new InternalError(s"No block with id ${block.header.parentId} found")
+        if (baseFee.compareTo(FeeUtils.INITIAL_BASE_FEE) != 0)
+          throw new InvalidBaseFeeException(s"Initial base fee ${baseFee} invalid")
+      case Some(parentBlock) =>
+        val parentBaseFee = parentBlock.header.baseFee
+        val range = parentBaseFee.multiply(BigInteger.valueOf(125)).divide(BigInteger.valueOf(1000))
+        val lowerBound = parentBaseFee.subtract(range).max(BigInteger.ONE)
+        val upperBound = parentBaseFee.add(range).max(BigInteger.ONE)
+        if (baseFee.compareTo(lowerBound) < 0 || baseFee.compareTo(upperBound) > 0)
+          throw new InvalidBaseFeeException(s"Calculated base fee ${baseFee} of block with id ${block.id} is out of 12.5% adjustment range")
```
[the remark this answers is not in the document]
JcBernack: True! A point of confusion here might be how gas limit is changed from block to block compared to the base fee.
When validating a block in Ethereum, the gas limit just has to be in the accepted bounds, but the base fee is recalculated and has to match exactly just like you said. [See geth source](https://github.com/ethereum/go-ethereum/blob/490c45c70fd1be8f9bade21abebd0cc9c989b562/consensus/misc/eip1559.go#L46).

sdk/src/test/scala/com/horizen/account/utils/AccountMockDataHelper.scala
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  def getMockedBlock(baseFee: BigInteger, gasUsed: Long, blockId: scorex.util.ModifierId, parentBlockId: scorex.util.ModifierId): AccountBlock = {
+    val block: AccountBlock = mock[AccountBlock]
+    Mockito.when(block.header).thenReturn(mock[AccountBlockHeader])
+    Mockito.when(block.header.parentId).thenReturn(parentBlockId)
+    Mockito.when(block.id).thenReturn(blockId)
+    Mockito.when(block.header.baseFee).thenReturn(baseFee)
+    Mockito.when(block.header.gasUsed).thenReturn(gasUsed)
+    Mockito.when(block.header.gasLimit).thenReturn(FeeUtils.GAS_LIMIT)
```
i-Alex, line 34: looks good.
As was agree offline let's add a couple of test case with the modified `header.gasLimit`.
To be sure that algorithm works as expected in case the `gasLimit` may change as well.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -203,7 +194,7 @@ class AccountState(
 
       // update current base fee
       // TODO: should use updated base fee based on consumed gas in the new block
-      stateView.updateBaseFee(BigInteger.valueOf(mod.header.baseFee))
+      stateView.updateBaseFee(mod.header.baseFee)
```
JcBernack, line 197: We could update this to the new base fee, considering gas consumption of the block that was just applied, see TODO comment.

JcBernack (approved):
looks good to me

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
i-Alex: right.
`stateView.baseFee` is used in the validate(tx) method, which is used during tx apply to mempool.
So, if `block1` with baseFee `BF1` is the chain tip, then every tx must be validated against `BF2 = nextBaseFee(block1)`

sdk/src/main/scala/com/horizen/account/block/AccountBlockHeader.scala
```diff
@@ -35,7 +36,7 @@ case class AccountBlockHeader(
                                stateRoot: Array[Byte],
                                receiptsRoot: Array[Byte],
                                forgerAddress: AddressProposition,
-                               baseFee: Long,
+                               baseFee: BigInteger,
                                gasUsed: Long,
                                gasLimit: Long,
```
i-Alex, line 41: We need to verify the value of `gasLimit`.
At the moment malicious forger may setup any value and we will accept it.
Let's update `semanticValidity` method below and check:
0) baseFee > 0
1) gasLimit == FeeUtils.GAS_LIMIT
2) gasUsed >= 0 && gasUsed  <= gasUsed

i-Alex (changes requested):

i-Alex (approved):

PR #518, opened 2022-08-31
Block Context

JcBernack:
- add `BlockContext` for contextual information during message processing
- remove block height, base fee and block gas limit from `AccountStateReader`
- update usages to get the information from `BlockContext`
- update tests


sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -122,13 +123,19 @@ class AccountState(val params: NetworkParams,
 
       // get also list of receipts, useful for computing the receiptRoot hash
       val receiptList = new ListBuffer[EthereumReceipt]()
-      val blockNumber = stateView.getHeight + 1
+      val blockNumber = stateMetadataStorage.getHeight + 1
```
JcBernack, line 126: Is this the correct way getting the new block number?

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +350,31 @@ class AccountState(val params: NetworkParams,
...
+        0,
+        // TODO: what baseFee do we use here? can we get the baseFee of the last block?
+        0,
+        // TODO: can we get the gas limit of the last block?
+        blockGasLimit,
+        stateMetadataStorage.getHeight,
+        // TODO: can the consensus epoch number be None here?
+        stateMetadataStorage.getConsensusEpochNumber.get,
+        stateMetadataStorage.getWithdrawalEpochInfo.epoch
+      )
```
JcBernack, line 374: I could use some feedback regarding the TODOs here.

sdk/src/main/java/com/horizen/utils/TimeToEpochUtils.scala
```diff
@@ -1,49 +1,56 @@
 package com.horizen.utils
```
JcBernack, line 1: This was just an auto-reformat with scalafmt. If pure whitespace changes like these are undesirable I can undo this.

sdk/src/main/scala/com/horizen/utils/WithdrawalEpochUtils.scala
```diff
@@ -1,37 +1,61 @@
 package com.horizen.utils
```
JcBernack, line 1: Same here, just an auto-reformat to reduce line length and increase readability.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
akrckn: I think so - @i-Alex?

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
akrckn: 1-3.) If we pass the block here instead of or in addition to tx we can surely do that, because you then get the parent block id, too.
Do we want to do that @i-Alex?

akrckn: 4.) afaik no at this point

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,106 +61,76 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+    val gasLimit = Account.GAS_LIMIT
 
-    // 5. Get cumulativeGasUsed from last receipt in list if available
-    val gasUsed = if (receiptList.size > 0) receiptList.last.cumulativeGasUsed.longValue() else 0
+    val blockContext = new BlockContext(
+      forgerAddress.address(),
+      timestamp,
+      baseFee,
+      gasLimit,
+      nodeView.history.getBlockHeightById(parentId).asScala.map(height => height + 1).getOrElse(0),
```
i-Alex, line 99: I would say that in case there is no parent, out block has height `1`, not `0`.
So can be `nodeView.history.getBlockHeightById(parentId).asScala.getOrElse(0) + 1`

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -77,106 +61,76 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+      nextConsensusEpochNumber,
+      nodeView.state.getWithdrawalEpochInfo.epoch
```
i-Alex, line 101: here we should use same as in the state:
`val modWithdrawalEpochInfo = WithdrawalEpochUtils.getWithdrawalEpochInfo(mod, parentWithdrawalEpochInfo, params)`
So, get the info of the Parent and then apply utility method.
Otherwise, at least for WithdrawalMsgProcessor forger may get different stateDbRoot result than State will have.

Another hint, `nodeView.history.getBlockHeightById(parentId)` above can be changed to get block info  -> height, just to avoid looking in the database twice.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -68,22 +69,18 @@ class AccountState(val params: NetworkParams,
       // Validate top quality certificate in the end of the submission window:
       // Reject block if it refers to the chain that conflicts with the top quality certificate content
       // Mark sidechain as ceased in case there is no certificate appeared within the submission window.
-      val currentWithdrawalEpochInfo = stateView.getWithdrawalEpochInfo
-      val modWithdrawalEpochInfo: WithdrawalEpochInfo = WithdrawalEpochUtils.getWithdrawalEpochInfo(mod, currentWithdrawalEpochInfo, params)
+      val parentWithdrawalEpochInfo = stateMetadataStorage.getWithdrawalEpochInfo
```
i-Alex, line 72: can be just `val parentWithdrawalEpochInfo = getWithdrawalEpochInfo`

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -68,22 +69,18 @@ class AccountState(val params: NetworkParams,
...
+      val modWithdrawalEpochInfo = WithdrawalEpochUtils.getWithdrawalEpochInfo(mod, parentWithdrawalEpochInfo, params)
 
       // If SC block has reached the certificate submission window end -> check the top quality certificate
       // Note: even if mod contains multiple McBlockRefData entries, we are sure they belongs to the same withdrawal epoch.
-      if (WithdrawalEpochUtils.hasReachedCertificateSubmissionWindowEnd(mod, currentWithdrawalEpochInfo, params)) {
+      if (WithdrawalEpochUtils.hasReachedCertificateSubmissionWindowEnd(mod, parentWithdrawalEpochInfo, params)) {
```
i-Alex, line 77: we need to use `currentWithdrawalEpochInfo` not `parentWithdrawalEpochInfo` as it was before.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +338,31 @@ class AccountState(val params: NetworkParams,
...
     if (tx.isInstanceOf[EthereumTransaction]) {
 
       val ethTx = tx.asInstanceOf[EthereumTransaction]
-      val blockGasPool = new GasPool(getBlockGasLimit)
+      val blockGasLimit = Account.GAS_LIMIT
+      val blockGasPool = new GasPool(BigInteger.valueOf(blockGasLimit))
+      val blockContext = new BlockContext(
+        // use the null address as forger
+        new Array[Byte](32),
+        // TODO: what timestamp do we use here? just use "now()" or can we get the timestamp of the last block?
```
i-Alex, line 352: i think we need to use `timeProvider: NetworkTimeProvider` method `time()`
So the tx that is out of block is always validated against the current time.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +338,31 @@ class AccountState(val params: NetworkParams,
...
+        0,
+        // TODO: what baseFee do we use here? can we get the baseFee of the last block?
```
i-Alex, line 354: yes, we may use the base fee of the block candidate after current tip.
For this we may store this data in the metadata storage, for example.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +338,31 @@ class AccountState(val params: NetworkParams,
...
+        0,
+        // TODO: can we get the gas limit of the last block?
```
i-Alex, line 356: so far we rely on the fact the Gas limit is a constant.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +338,31 @@ class AccountState(val params: NetworkParams,
...
+        blockGasLimit,
+        stateMetadataStorage.getHeight,
+        // TODO: can the consensus epoch number be None here?
```
i-Alex, line 359: no, it can't

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -347,20 +338,31 @@ class AccountState(val params: NetworkParams,
...
+      val blockContext = new BlockContext(
+        // use the null address as forger
+        new Array[Byte](32),
+        // TODO: what timestamp do we use here? just use "now()" or can we get the timestamp of the last block?
+        0,
+        // TODO: what baseFee do we use here? can we get the baseFee of the last block?
+        0,
+        // TODO: can we get the gas limit of the last block?
+        blockGasLimit,
+        stateMetadataStorage.getHeight,
```
i-Alex, line 358: `getHeight+1`

sdk/src/main/scala/com/horizen/forge/AbstractForgeMessageBuilder.scala
```diff
@@ -97,7 +97,7 @@ abstract class AbstractForgeMessageBuilder[
 
       val forgingResult = eligibleForgerOpt
         .map { case (forgingStakeMerklePathInfo, privateKey25519, vrfProof, _) =>
-          forgeBlock(nodeView, nextBlockTimestamp, branchPointInfo, forgingStakeMerklePathInfo, privateKey25519, vrfProof, timeout)
+          forgeBlock(nodeView, nextConsensusEpochNumber, nextBlockTimestamp, branchPointInfo, forgingStakeMerklePathInfo, privateKey25519, vrfProof, timeout)
```
i-Alex, line 100: let's keep it as it was before.
we can reconstruct `nextConsensusEpochNumber` from the `nextBlockTimestamp`

sdk/src/main/scala/com/horizen/state/BaseStateReader.scala
```diff
@@ -2,12 +2,9 @@ package com.horizen.state
 
 import com.horizen.account.state.WithdrawalRequest
 import com.horizen.block.WithdrawalEpochCertificate
-import com.horizen.consensus.ConsensusEpochNumber
-import com.horizen.utils.{BlockFeeInfo, WithdrawalEpochInfo}
+import com.horizen.utils.BlockFeeInfo
 
 trait BaseStateReader {
-  def getWithdrawalEpochInfo: WithdrawalEpochInfo
```
i-Alex, line 9: Note: here we actually left available for the MsgProcessors to get access to all other methods of the `BaseStateReader` for `BaseAccountStateView`, but from other side removed an access to the `getWithdrawalEpochInfo` and `getConsensusEpochNumber` from `AccountStateView`.
Let's keep this class as is.
But change the overall hierarchy of thew classes:
It was `BaseStateReader -> AccountStateReader -> BaseAccountStateView -> AccountStateView`
can be
```
BaseAccountStateReader -> BaseAccountStateView (empty)
                               \                                                                      
BaseStateReader ------------>  AccountStateReader   -> AccountStateView
```
`BaseAccountStateReader` contains the actual `AccountStateReader` methods but doesn't extend `BaseStateReader`.
Moreover `BaseAccountStateView` seems to be empty now. So we even can remove it and use `BaseAccountStateReader`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
JcBernack: And will the genesis block ever go through this function?
- if no, there will be no block that does not have a parent so it doesnt matter..?
- if yes, what is the height of the genesis block, `0` or `1`?

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
JcBernack: done

sdk/src/main/scala/com/horizen/forge/AbstractForgeMessageBuilder.scala
JcBernack: done

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
JcBernack: What about calculating the timestamp from consensus epoch and slot+1, is that possible? Epoch is already known here, we are just missing the slot.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
JcBernack: done

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -131,7 +130,12 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
       applyOnNodeView { sidechainNodeView =>
         val valueInWei = ZenWeiConverter.convertZenniesToWei(body.value)
         val destAddress = body.to
-        val gasPrice = sidechainNodeView.getNodeState.getBaseFee // TODO actual gas implementation
+        // TODO actual gas implementation
+        val gasPrice = {
+          val bestBlock = sidechainNodeView.getNodeHistory.getBestBlock
+          if (bestBlock == null) BigInteger.ZERO
```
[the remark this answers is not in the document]
JcBernack: During tests it is not, which is why I added the default here. `AccountTransactionApiRouteTest` failed otherwise.

sdk/src/main/scala/com/horizen/state/BaseStateReader.scala
JcBernack: I don't understand what you want to achieve yet. The way I see it we currently have three traits roughly corresponding to this:
- `BaseStateReader` base sidechain stuff, consensus epoch, withdrawal epoch, etc.
- `AccountStateReader` read-only account state access, like getBalance, getNonce, etc.
- `BaseAccountStateView` write access to state, like addBalance, updateAccountStorage, etc.

These traits extend each other like you said:
`BaseStateReader -> AccountStateReader -> BaseAccountStateView`

Usages are:
- `AccountState` implements `AccountStateReader` (and thus `BaseStateReader`)
- `AccountStateView` implements `BaseAccountStateView` and therefore all three
- `MessageProcessor`s  have access to `BaseAccountStateView`

Do you want to prevent `MessageProcessors` from having access to `BaseStateReader`?

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
i-Alex: genesis block does not use it.
but the height of the genesis block is 1.
I even think, that in case Forger get `nodeView.history.getBlockHeightById(parentId)` result as None -> it is a bug an we need to throw an exception.

sdk/src/main/scala/com/horizen/state/BaseStateReader.scala
i-Alex: right, `MessageProcessors` should not be able to use/see `BaseStateReader` methods. In case using of the StateView in the past -> the methods output may be irrelevant.

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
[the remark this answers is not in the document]
i-Alex: ok, i got it.
Then in the test we need to mock `sidechainNodeView.getNodeHistory.getBestBlock` method to return non-null mocked `block.header.gasPrice`

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
i-Alex: the point with consensus epoch&slot that we don't know the forger block values. It can be `+1` or `+many`.
The consensus protocol we use has a time based slots who may remain empty. From other side, considering again that slots are time based, we may assume that current time is the mostly likely to be used when the block is going to be produced.
Note: we can't use system.time, but timeProvider that considers time shift of the node.

i-Alex (commented):
Changes seems good.
Let's do the fix for the test and timestamp, then merge actual dev_evm into current branch.

i-Alex (approved):

PR #519, opened 2022-08-31
Fix tests gas usage

GitMark0:
Changes:
- Removed `gasPrice` and `gasLimit` from static rpc calls
- Tests that check native currency balance also take gas used into account


qa/sc_evm_test_metamask_related.py
```diff
@@ -640,6 +644,5 @@ def run_test(self):
                                     initial_balance - transfer_amount)
         res = compare_erc20_balance(sc_node, smart_contract, smart_contract_address, other_address, transfer_amount)
 
-
-if __name__ == "__main__":
-    SCEvmMetamaskTest().main()
+        if __name__ == "__main__":
```
JcBernack, line 647: Accidental indentation?

qa/sc_evm_test_metamask_related.py
```diff
@@ -409,19 +401,31 @@ def run_test(self):
         eoa_transfer(sc_node, evm_address, other_address, transfer_amount, static_call=True, generate_block=False)
         eoa_transfer(sc_node, evm_address, other_address, transfer_amount, call_method=CallMethod.RPC_LEGACY)
 
-        eoa_assert_native_balance(sc_node, evm_address, initial_balance - transfer_amount)
+        current_balance = int(sc_node.rpc_eth_getBalance(str(evm_address), "latest")['result'], 16)
+        total_gas_used = initial_balance - (current_balance + transfer_amount)
+
+        eoa_assert_native_balance(sc_node, evm_address, current_balance - (transfer_amount + total_gas_used))
```
JcBernack, line 407: This should be `initial_balance - ...` right?

Additionally, I don't think we can test like this. By fetching the balance to calculate the fees and then verifying the balance this becomes a self-fulfilling test. An alternative would be to fetch the receipt for the transaction, get the `usedGas` and multiply it with the `gasPrice` we set in the transaction. This would work at least for legacy transactions. I don't know if we can fetch receipts yet, though.

JcBernack (changes requested):

qa/SidechainTestFramework/account/eoa_util.py
```diff
@@ -130,5 +130,9 @@ def eoa_transaction(node, *,
                                               max_priority_fee_per_gas=max_priority_fee_per_gas, chain_id=chain_id)
 
     response = node.rpc_eth_signTransaction(payload)
-    response = node.rpc_eth_sendRawTransaction(response['result'])
-    return response["result"]
+
+    if "result" in response:
+        response = node.rpc_eth_sendRawTransaction(response['result'])
+        return response["result"]
```
JcBernack, line 136: Missing one last check for `"result" in response` after `rpc_eth_sendRawTransaction`.

qa/httpCalls/block/findBlockByID.py
```diff
@@ -1,13 +1,15 @@
...
-      j = {\
-            "blockId": blockId
-      }
-      request = json.dumps(j)
-      response = sidechainNode.block_findById(request)
-      return response["result"]
 
+# executes a  block/findById call
+def http_block_findById(sidechainNode, blockId):
+    j = { \
```
JcBernack, line 6: unnecessary backslash

qa/httpCalls/transaction/findTransactionByID.py
```diff
@@ -1,14 +1,16 @@
...
-      }
-      request = json.dumps(j)
-      response = sidechainNode.transaction_findById(request)
-      return response["result"]
 
+# executes a  transaction/findById call
+def http_transaction_findById(sidechainNode, txid):
+    j = { \
+        "transactionId": txid, \
+        "format": True \
```
JcBernack, line 8: unnecessary backslashes, why do these exist? :thinking:

qa/SidechainTestFramework/account/ac_use_smart_contract.py
```diff
@@ -120,8 +120,11 @@ def call_function(self, node, functionName: str, *args,
                                                  max_fee_per_gas=maxFeePerGas,
                                                  max_priority_fee_per_gas=maxPriorityFeePerGas, gas=gasLimit, data=data,
                                                  value=value))
-        response = node.rpc_eth_sendRawTransaction(response['result'])
-        return response["result"]
+        if "result" in response:
+            response = node.rpc_eth_sendRawTransaction(response['result'])
+            if "result" in response:
+                return response["result"]
+        raise RuntimeError("Something went wrong, see {}".format(str(response)))
```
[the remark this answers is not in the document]
JcBernack: good

JcBernack (changes requested):

qa/httpCalls/transaction/findTransactionByID.py
GitMark0: Either that was there before, or the formatter added them. I don't recall writing them 😺 .

JcBernack: They were there before yes. Always leave code cleaner than it was before :stuck_out_tongue_winking_eye:

JcBernack (approved):

i-Alex (approved):

PR #521, opened 2022-09-02
add test for estimateGas

akrckn:
This test will work if EVM-299 is fixed - therefore it is a draft pr - fix should be built on top of this one.


qa/sc_evm_bootstrap.py
```diff
@@ -214,6 +214,16 @@ def run_test(self):
...
...65206f776e6572600082015250565b7f45524332303a206275726e2066726f6d20746865207a65726f2061646472657360008201527f7300000000000000000000000000000000000000000000000000000000000000602082015250565b7f45524332303a207472616e736665722066726f6d20746865207a65726f20616460008201527f6472657373000000000000000000000000000000000000000000000000000000602082015250565b7f45524332303a20617070726f76652066726f6d20746865207a65726f2061646460008201527f7265737300000000000000000000000000000000000000000000000000000000602082015250565b7f45524332303a2064656372656173656420616c6c6f77616e63652062656c6f7760008201527f207a65726f000000000000000000000000000000000000000000000000000000602082015250565b7f45524332303a206d696e7420746f20746865207a65726f206164647265737300600082015250565b611ba78161172e565b8114611bb257600080fd5b50565b611bbe8161176c565b8114611bc957600080fd5b5056fea264697066735822122087b660d81a0b4c64fea037a87fa782fb475cd227094a04112ee89714ca9325a364736f6c634300080700330000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
```
JcBernack, line 220: Where is this bytecode coming from? A short comment would be nice :wink:

JcBernack (changes requested):

JcBernack (approved):
good :+1:

qa/sc_evm_bootstrap.py
```diff
@@ -214,6 +214,16 @@ def run_test(self):
         # check if header contains correct gasUsed (2 * eoa to eoa transfer gas costs)
         assert_equal(21000*2, sc_best_block['block']['header']['gasUsed'])
 
+        # test eth_estimateGas with ERC20 (TestERC20.sol with additional mint function) contract bytecode
+        request = {
```
i-Alex, line 218: looks to be a good candidate to be put into the separate py test.
We need to extend it and check the gas estimation for:
1) EOA to EOA
2) EOA to EOA with NON empty data
3) Forger stake delegation/withdrawal
4) SC to MC withdrawal
5) other common smart contracts (any regression data from geth?)
6) eth

sdk/src/main/java/com/horizen/account/state/EvmMessageProcessor.java
```diff
@@ -59,6 +59,7 @@ public byte[] process(Message msg, BaseAccountStateView view, GasPool gas) throw
                     // returnData here will include the revert reason
                     throw new ExecutionRevertedException(returnData);
                 case "out of gas":
+                case "contract creation code storage out of gas":
```
i-Alex, line 62: 1) Are we sure now if we cover all the "out of gas" errors of EVM?
2) Relying on error text description seems to be very-very unstable.

There is a list of EVM errors in `core/vm/errors.go`.
We need to introduce mapping between that go types and java wrapper types.
Then here we may do a patter matching per type instead of per text.
@JcBernack, want to know your opinion as well.

i-Alex (changes requested):
Looks good in general.
I can see a lack of tests (added a comment). Let's add them

JcBernack (changes requested):

sdk/src/main/java/com/horizen/account/state/EvmMessageProcessor.java
i-Alex: Agree with you.
This way of error handling is much better, and `reverted` flag is much clear for external user (like EvmMsgProcessor).
@akrckn let's proceed in this way in current PR.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -209,6 +208,8 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
       val lowBound = GasUtil.TxGas.subtract(BigInteger.ONE)
       // Determine the highest gas limit can be used during the estimation.
       var highBound = params.gas
+      // we need to set nonce to one, because we substract by one in the evm call
+      params.nonce = BigInteger.ONE
```
JcBernack, line 212: This will not fix our issue, we need to discuss this issue further.

akrckn: As discussed this might be an issue related to the docker setup and not related to the PR here, so I would suggest closing this PR without any changes regarding the ticket 319.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -61,12 +60,12 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
     Await.result(res, nvtimeout) match {
       case Failure(exception) => exception match {
         case reverted: ExecutionRevertedException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, exception.getMessage, Numeric.toHexString(reverted.revertReason)))
+          RpcCode.ExecutionError.getCode, reverted.getMessage, Numeric.toHexString(reverted.revertReason)))
         case err: ExecutionFailedException => throw new RpcException(new RpcError(
           RpcCode.ExecutionError.getCode, err.getMessage, null))
-        case err: TransactionSemanticValidityException => throw new RpcException(new RpcError(
+        case err: IntrinsicGasException => throw new RpcException(new RpcError(
```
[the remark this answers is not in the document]
akrckn: In geth the default error code `-32000` is used for these kind of errors. Should we change that here and where did our `3` come from?

i-Alex (changes requested):

qa/sc_evm_estimateGas.py
```diff
@@ -0,0 +1,205 @@
...
+        # data from AccountTransactionApiRouteTest
+        request = {
+            "from": "0x" + evm_address,
+            "data": "0x5ca748ff1122334455669988112233445566778811223344556677881122334455667788aabbddddeeff0099aabbccddeeff0099aabbccddeeff0099aabbccddeeff00123400000000000000000000000000000000000000000000000000000000000000000000000000000000000000b876fa045fcd3d15a3b6ea51497e06244af17978",
+            "to": "0x0000000000000000000022222222222222222222",
+            "value": "0x1",
+            "gasPrice": "0x4B9ACA00"
+        }
+        response = sc_node_1.rpc_eth_estimateGas(request)
+        assert_equal('0x1c9c380', response['result'])
```
i-Alex, line 190: `0x1c9c380` equals to 30M - block gas limit.
It cant be like this...
We for sure have a bug.

qa/sc_evm_estimateGas.py
```diff
@@ -0,0 +1,205 @@
...
+        # data from AccountTransactionApiRouteTest
+        request = {
+            "from": "0x" + evm_address,
+            "data": "0x9950a60f9988b9e8e729c83ecccb1c9549c46ed226f14de7000000000000000000000000",
+            "to": "0x0000000000000000000011111111111111111111",
+            "value": "0x1",
+            "gasPrice": "0x4B9ACA00"
+        }
+        response = sc_node_1.rpc_eth_estimateGas(request)
+        assert_equal('0x1c9c380', response['result'])
```
i-Alex, line 202: same here

i-Alex (changes requested):

qa/sc_evm_estimateGas.py
akrckn: Oh, I did not notice that! I'll directly check why that occurs

i-Alex (approved):

JcBernack (approved):

PR #524, opened 2022-09-02
Gas refunds for fake contracts

JcBernack:
The EVM tracks gas refunds within the implementation of the SSTORE op-code, our fake smart contracts access StateDB directly and therefore bypass this logic.

This PR replicates the gas refund mechanic of SSTORE in our methods that write to account storage.

### DIFF
--- a/libevm/lib/service_state_refund.go
+++ b/libevm/lib/service_state_refund.go
@@ -1,0 +1,61 @@
+package lib
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
+	gethParams "github.com/ethereum/go-ethereum/params"
+)
+
+// setStateRefunds replicates the refund part of the original SSTORE gas consumption logic:
+//
+// 0. If *gasleft* is less than or equal to 2300, fail the current call.
+// 1. If current value equals new value (this is a no-op), SLOAD_GAS is deducted.
+// 2. If current value does not equal new value:
+//   2.1. If original value equals current value (this storage slot has not been changed by the current execution context):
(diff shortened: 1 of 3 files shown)


akrckn (approved):
Seems fine to me!

libevm/lib/service_state_refund.go
```diff
@@ -0,0 +1,61 @@
+package lib
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/state"
+	gethParams "github.com/ethereum/go-ethereum/params"
+)
+
+// setStateRefunds replicates the refund part of the original SSTORE gas consumption logic:
```
i-Alex, line 9: It seems that we implement `EIP-1283` here and it is the most actual implementation, right?
Let's add the regression test coverage using the test data from it: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1283.md#test-cases

I can see similar for the eip2200 in Geth code - `eip2200Tests`.
It is strange that they have only legacy tests there...

i-Alex (changes requested):
Looks good.
Let's add some test data to check the consistency (see the comment).

libevm/lib/service_state_refund.go
JcBernack: Not quite, 1283 was the first modification to SSTORE refunds, then came 2200, 2929 and then 3529. This implements the latest one, 3529. I will add their test cases from [EIP-3529](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-3529.md#test-cases).

i-Alex: Ok, thanks. than 3529

JcBernack: done

i-Alex (approved):

PR #525, opened 2022-09-05
Added top quality cert in state when processing a mc ref while applying block; added a py …

alsala:
…script for testing that


i-Alex (approved):

lander86 (approved):
LGTM

PR #526, opened 2022-09-05
bugfixes for estimateGas

akrckn:
add optional nonce validation to msg precheck, add optional forwarding of errors in doCall function, add missing case in msg processor

### DIFF
--- a/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
+++ b/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
@@ -109,19 +109,23 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
     }
   }
 
-  private def doCall[A](nodeView: NV, params: TransactionArgs, tag: String)(fun: (Array[Byte], AccountStateView) ⇒ A): A = {
+  private def doCall[A](nodeView: NV, params: TransactionArgs, tag: String, forwardErrors: Boolean)(fun: (Array[Byte], AccountStateView) ⇒ A): A = {
     getStateViewAtTag(nodeView, tag) { tagStateView =>
-      try {
-        val msg = params.toMessage(tagStateView.getBaseFee)
-        fun(tagStateView.applyMessage(msg, new GasPool(msg.getGasLimit)), tagStateView)
-      } catch {
-        // throw on execution errors, also include evm revert reason if possible
-        case reverted: ExecutionRevertedException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, reverted.getMessage, Numeric.toHexString(reverted.revertReason)))
-        case err: ExecutionFailedException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, err.getMessage, null))
-        case err: TransactionSemanticValidityException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, err.getMessage, null))
+      val msg = params.toMessage(tagStateView.getBaseFee)
+      if (forwardErrors) {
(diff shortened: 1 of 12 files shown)


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -242,7 +246,7 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
       // other exceptions are not caught as the call would not succeed with any amount of gas
       val check = (gas: BigInteger) => try {
         params.gas = gas
-        doCall(nodeView, params, tag) { (_, _) => true }
+        doCall(nodeView, params, tag, true) { (_, _) => true }
```
JcBernack, line 249: I think we actually want the exception mapping to happen here too. We just want it outside the binary search, not inside where it would mask the `OutOfGas` exception.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -109,19 +109,23 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
...
-        // throw on execution errors, also include evm revert reason if possible
-        case reverted: ExecutionRevertedException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, reverted.getMessage, Numeric.toHexString(reverted.revertReason)))
-        case err: ExecutionFailedException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, err.getMessage, null))
-        case err: TransactionSemanticValidityException => throw new RpcException(new RpcError(
-          RpcCode.ExecutionError.getCode, err.getMessage, null))
+      val msg = params.toMessage(tagStateView.getBaseFee)
+      if (forwardErrors) {
+        fun(tagStateView.applyMessage(msg, new GasPool(msg.getGasLimit), true), tagStateView)
```
JcBernack, line 116: With `forwardErrors == true` this whole function doesn't really do anything anmore :grinning: 
Lets move the exception mapping to `applyOnAccountStateView`. That way it would be wrapped around every relevant call on the outside. [See line 60](https://github.com/HorizenOfficial/Sidechains-SDK/pull/526/files#diff-9f3a238171c2b2e81d34f037a29411b0fd262d527fcbb24883ed13db1676aedbR60)

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
JcBernack: See the other comment about moving the exception mapping.

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
```diff
@@ -46,22 +46,23 @@ class StateTransition(view: AccountStateView, messageProcessors: Seq[MessageProc
     }
   }
 
-  private def preCheck(msg: Message): Unit = {
+  private def preCheck(msg: Message, fakeMsg: Boolean): Unit = {
```
JcBernack, line 49: I would prefer to have `isFakeMessage` as a property on the `Message` which gets set to `true` in `TransactionArgs.toMessage()` and `false` in `EthereumTransaction.asMessage()`. What do you think?

JcBernack (changes requested):
My comment are minor and if this bug is blocking people we could merge and cleanup afterwards.

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
akrckn: Seems fine to me, will do

JcBernack (approved):
good :+1:

PR #527, opened 2022-09-06
As/forger fee payments

alsala:
Implemented pool approach, base fee goes to the forgers pool (forgers that created at least one block in the epoch) and tips directly to each block forger.
In py STF added a util method for getting the fees contributions of a forged transaction.


sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -340,7 +355,25 @@ class AccountStateView(
...
+      // if the Base Fee plus the Max Priority Fee exceeds the Max Fee, the Max Priority Fee will be reduced
+      // in order to maintain the upper bound of the Max Fee.
+      val forgerTipPerGas = if (baseFeePerGas.add(maxPriorityFeePerGas).compareTo(maxFeePerGas) > 0) {
+        maxFeePerGas.subtract(baseFeePerGas)
+      } else {
+        maxPriorityFeePerGas
+      }
+      (baseFeePerGas, forgerTipPerGas)
+    } else {
+      (BigInteger.ZERO, tx.getGasPrice)
```
JcBernack, line 374: The type of transaction that a user submits has an impact on the way we distribute fees? Is this how we want it to be?
Even in legacy transactions the gasPrice has to be greater or equal to the base fee, why don't we just divide the gasPrice into base fee and tip and distribute fees accordingly?
`(baseFeePerGas, tx.getGasPrice.subtract(baseFeePerGas))`

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
```diff
@@ -105,10 +108,20 @@ class StateTransition(view: AccountStateView, messageProcessors: Seq[MessageProc
...
     val usedGas = msg.getGasLimit.subtract(gas.getGas)
     // cap gas refund to a quotient of the used gas
-    gas.addGas(view.getRefund.min(usedGas.divide(GasUtil.RefundQuotientEIP3529)))
+    val refGas = view.getRefund.min(usedGas.divide(GasUtil.RefundQuotientEIP3529))
+    gas.addGas(refGas)
     // return funds for remaining gas, exchanged at the original rate.
     val remaining = gas.getGas.multiply(msg.getGasPrice)
-    view.addBalance(msg.getFrom.address(), remaining)
+    val sender = msg.getFrom.address()
+    log.debug(s"gas used: $usedGas, remaining gas: ${gas.getGas}, refGas: $refGas, gasPrice: ${msg.getGasPrice}")
```
JcBernack, line 116: `usedGas` will be outdated here because we added some gas back to the pool in line 112.

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
```diff
@@ -105,10 +108,20 @@ class StateTransition(view: AccountStateView, messageProcessors: Seq[MessageProc
...
+    log.debug(s"now refunding $remaining to sender ${BytesUtils.toHexString(sender)}")
+    if (remaining.compareTo(BigInteger.ZERO) > 0) {
+      view.addBalance(sender, remaining)
+      // the addBalance op subtracts gas, we have to restore it here otherwise remaining gas balance is wrong
```
JcBernack, line 120: Good point! The gas consumption should be disabled before we even call `refundGas`. Except for the intrinsic gas, gas should only be consumed from within the message processors. Here we already arrived at a point where "out of gas" should not happen anymore. Currently we make sure gas tracking is disabled in `applyTransaction`, lets move it to the finally block in `transition` before we call `refundGas`. What do you think?

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import java.math.BigInteger
+
+object AccountFeePaymentsUtils {
+  val DEFAULT_ACCOUNT_FEE_PAYMENTS_HASH: Array[Byte] = new Array[Byte](MerkleTree.ROOT_HASH_LENGTH)
+
+  def calculateFeePaymentsHash(feePayments: Seq[AccountBlockFeeInfo]): Array[Byte] = {
+    if(feePayments.isEmpty) {
+      // No fees for the whole epoch, so no fee payments for the Forgers.
+      DEFAULT_ACCOUNT_FEE_PAYMENTS_HASH
```
JcBernack, line 15: This would all zeros, correct? The hash of an empty trie is `EMPTY_ROOT_HASH`. Does it make sense to use that instead?

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -0,0 +1,57 @@
...
+    val forgerKeys: Seq[AddressProposition] = allForgersRewards.map(_._1).distinct
+
+    val forgersRewards: Seq[(AddressProposition, BigInteger)] = forgerKeys.map {
+      forgerKey => { // consider this forger
+        var forgerTotalFee = BigInteger.ZERO
+        allForgersRewards.withFilter(
+          pair => forgerKey.equals(pair._1) // if the address is the one of this forger
+        ).foreach(
+          pair => forgerTotalFee = forgerTotalFee.add(pair._2) // increment the amount for this address
+        )
```
JcBernack, line 51: Could be done without mutation
```
// sum all rewards for this forger address
val forgerTotalFee = allForgersRewards
  .withFilter(pair => forgerKey.equals(pair._1))
  .foldLeft(BigInteger.ZERO)((sum, pair) => sum.add(pair._2))
```

JcBernack (changes requested):

qa/SidechainTestFramework/scutil.py
```diff
@@ -1089,4 +1090,52 @@ def get_account_balance(sc_node, address):
     return sc_node.wallet_getBalance(
         json.dumps({"address": str(address)}))["result"]["balance"]
 
+def computeForgedTxFee(sc_node, tx_hash, tracing_on=False):
+    transactionJson = sc_node.rpc_eth_getTransactionByHash(tx_hash)['result']
```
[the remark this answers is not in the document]
alsala: Actually that line does not throw an error, but `transactionJson` will be `None` , and the condition is checked in the next line.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
alsala: I took inspiration from what I saw in geth/etherscan, where there is no burnt quota for legacy txes. I think both alternatives are viable, and I personally like best the approach you are proposing. Lets discuss it further.

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
alsala: Agree, that could be a better option, I chose this because the UTXO model has the analogous constant value defined as:
`  val DEFAULT_FEE_PAYMENTS_HASH: Array[Byte] = new ArrayBytes(32)`

qa/SidechainTestFramework/account/httpCalls/createEIP1559Transaction.py
```diff
@@ -19,6 +20,10 @@ def createEIP1559Transaction(sidechainNode, *, chainId, fromAddress=None, toAddr
         "signature_r": signature_r,
         "signature_s": signature_s
     }
+    if nonce:
+        j["nonce"] = nonce
     request = json.dumps(j)
     response = sidechainNode.transaction_createEIP1559Transaction(request)
+    if not 'result' in response:
+      pprint.pprint(response)
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -687,11 +700,11 @@ object AccountTransactionRestScheme {
   private[api] case class ReqEIP1559Transaction(
                                                  from: Option[String],
                                                  to: Option[String],
-                                                 nonce: BigInteger,
+                                                 nonce: Option[BigInteger],
                                                  gasLimit: BigInteger,
                                                  maxPriorityFeePerGas: BigInteger,
                                                  maxFeePerGas: BigInteger,
-                                                 value: BigInteger,
+                                                 @JsonDeserialize(contentAs = classOf[lang.Long]) value: Long,
```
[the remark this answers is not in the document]
alsala: Actually no reason, you are right. Reverted back to BigInteger

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -298,17 +303,21 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
 
-        // TODO actual gas implementation
-        var maxFeePerGas = BigInteger.ONE
-        var maxPriorityFeePerGas = BigInteger.ONE
-        var gasLimit = BigInteger.ONE
+        // default gas related params
+        val baseFee = sidechainNodeView.getNodeState.getBaseFeePerGas
+        var maxPriorityFeePerGas = BigInteger.valueOf(120)
+        var maxFeePerGas = BigInteger.TWO.multiply(baseFee).add(maxPriorityFeePerGas)
+        var gasLimit = GasUtil.TxGasContractCreation
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
alsala: Agree

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
alsala: Agree, the modification you suggest is much cleaner. Done.

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
alsala: Thanks! Done

qa/SidechainTestFramework/scutil.py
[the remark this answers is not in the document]
JcBernack: That is true if the transaction was not found but the call itself was valid and executed successfully. If for some reason the call itself fails, e.g. because of "invalid params", "internal error" or any bug on the RPC handler we get an error response like
```
{
  "jsonrpc" : "2.0",
  "id" : "1",
  "error" : {
    "code" : -32602,
    "message" : "Invalid params"
  }
}
```
In that case `response["result"]` will throw and we don't even see the actual error but the generic python error for accessing a non-existent key.

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
JcBernack: Ok, so `EMPTY_ROOT_HASH` is a bit more consistent with ethereum, while all-zeroes is more consistent within the SDK? :wink:

JcBernack (approved):

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
JcBernack: Interesting. I had a look too and on Etherscan it really looks like legacy TXs do not burn any of the fees, but that is not true. The geth implementation does not differentiate between TX types [here](https://github.com/ethereum/go-ethereum/blob/20356e57b119b4e70ce47665a71964434e15200d/core/state_transition.go#L335).
```
effectiveTip := st.gasPrice
if london {
    effectiveTip = cmath.BigMin(st.gasTipCap, new(big.Int).Sub(st.gasFeeCap, st.evm.Context.BaseFee))
}
st.state.AddBalance(st.evm.Context.Coinbase, new(big.Int).Mul(new(big.Int).SetUint64(st.gasUsed()), effectiveTip))
```
At this point, for any type of TX there are usable values for all the fields (gasPrice, feeCap and tip). So this works on any kind of TX as expected: burn the base fee, everything else goes to coinbase. Same thing is applicable to us because we do the same when creating a message from a TX: if it is a legacy TX set both feeCap and tip to gasPrice.

Here are some very helpful illustrations I just found:
- [EIP-1559 TX](https://uploads-ssl.webflow.com/5fe512bf3e57baa22a4812f0/611fec11baefcfd52512b216_0*R_f7IHC0jAQBJBEC.png)
- [Legaxy TX](https://uploads-ssl.webflow.com/5fe512bf3e57baa22a4812f0/611fec114bafd083ee330450_0*nu522iCGRurPooVF.png)
- [Source article](https://www.liquity.org/blog/how-eip-1559-changes-the-transaction-fees-of-ethereum)

qa/SidechainTestFramework/scutil.py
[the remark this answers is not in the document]
alsala: Ok you are right, I added error handling

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
alsala: Ok, I added fee splitting pool/forger also for legacy txes

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
alsala: I set the default to EMPTY_ROOT_HASH as suggested, this is consistent with other empty tries in the block

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
JcBernack: this one is still open

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
alsala: Ok, done

JcBernack (approved):

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -298,9 +304,16 @@ class AccountStateView(
...
 
-  override def getFeePayments(withdrawalEpoch: Int): Seq[BlockFeeInfo] =
-    metadataStorageView.getFeePayments(withdrawalEpoch)
+  override def getFeePayments(withdrawalEpoch: Int, blockToAppendFeeInfo: Option[AccountBlockFeeInfo] = None): Seq[AccountBlockFeeInfo] = {
+    var blockFeeInfoSeq = metadataStorageView.getFeePayments(withdrawalEpoch)
+    blockToAppendFeeInfo.foreach(blockFeeInfo => blockFeeInfoSeq = blockFeeInfoSeq :+ blockFeeInfo)
+
+    if(blockFeeInfoSeq.isEmpty) {
+      return Seq()
+    }
```
i-Alex, line 313: why do we need this? `blockFeeInfoSeq` is an empty Seq

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -298,9 +304,16 @@ class AccountStateView(
 
   override def getConsensusEpochNumber: Option[ConsensusEpochNumber] = metadataStorageView.getConsensusEpochNumber
 
-  override def getFeePayments(withdrawalEpoch: Int): Seq[BlockFeeInfo] =
-    metadataStorageView.getFeePayments(withdrawalEpoch)
+  override def getFeePayments(withdrawalEpoch: Int, blockToAppendFeeInfo: Option[AccountBlockFeeInfo] = None): Seq[AccountBlockFeeInfo] = {
```
i-Alex, line 307: Why do we return the seq of `sources` instead of `AccountFeePaymentsUtils.getForgersRewards`, so the resulting fee payments?
I assume, no one except the state inside is interested in the list of `AccountBlockFeeInfo`?

The root of fee payments expected to be the root of pairs `(AddressProposition, BigInteger)`. Otherwise, to understand the changes and verify the AccountBlock.feePaymentsHash a third-party need to know how to calculate `getForgersRewards`.

sdk/src/main/scala/com/horizen/account/state/GasPool.scala
```diff
@@ -17,15 +19,19 @@ class GasPool(initialGas: BigInteger) {
     if (currentGas.compareTo(gas) < 0) {
       throw new OutOfGasException()
     }
+    log.debug(s"subtracting $gas from currentGas=$currentGas")
```
i-Alex, line 22: don't forget to remove the debug calls from this place. Or at least keep it as `trace` logs.
otherwise they will bring a lot of noise to us.

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
```diff
@@ -41,6 +43,8 @@ class StateTransition(view: AccountStateView, messageProcessors: Seq[MessageProc
             throw err
           // any other exception will bubble up and invalidate the block
         } finally {
+          // make sure we disable automatic gas consumption in case a message processor enabled it
+          view.disableGasTracking()
```
i-Alex, line 47: so we move it here to avoid `refundGas` consuming another gas, right?

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -265,7 +266,7 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
     // If withdrawal epoch switched to the next one, then perform some database clean-up:
     // 1) remove outdated topQualityCertificate retrieved 3 epochs before and referenced to the 4 epochs before.
     //    Note: we should keep last 2 epoch certificates, so in case SC has ceased we have an access to the last active cert.
-    // 1) remove outdated BlockFeeInfo records
+    // 1) remove outdated AccountBlockFeeInfo records
```
i-Alex, line 269: Now i have a doubt if is it safe to remove outdated `AccountBlockFeeInfo`.
Can it have an impact in case we need to debug_block  (eth rpc call) with fee payments in the past?

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -0,0 +1,59 @@
...
+    if(feePaymentsInfo.isEmpty) {
+      // No fees for the whole epoch, so no fee payments for the Forgers.
+      DEFAULT_ACCOUNT_FEE_PAYMENTS_HASH
+    } else {
+      // turn seq elements into leaves and compute merkel root hash
+      TrieHasher.Root(feePaymentsInfo.map(info => info.bytes).toArray)
+    }
+  }
+
+  def getForgersRewards(blockFeeInfoSeq : Seq[AccountBlockFeeInfo]): Seq[(AddressProposition, BigInteger)] = {
```
i-Alex, line 22: let's add the batch of Unit tests similar to the `src/test/scala/com/horizen/SidechainStateTest.scala -> feePayments()`

qa/sc_evm_forging_fee_payments.py
```diff
@@ -0,0 +1,323 @@
...
+      assert_equal(node_1_fees + node_2_fees, total_fee)
+
+      # Check forger fee payments
+      assert_equal(sc_node_1_balance_after_payments, sc_node_1_balance_before_payments + node_1_fees,
+                   "Wrong fee payment amount for SC node 1")
+      assert_equal(sc_node_2_balance_after_payments, sc_node_2_balance_before_payments + node_2_fees,
+                   "Wrong fee payment amount for SC node 2")
+
+      '''
+      TODO enable when API will be ready
```
i-Alex, line 300: do you plan to implement API in this PR or as a part of the separate task after?

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -170,8 +170,13 @@ public void semanticValidity() throws TransactionSemanticValidityException {
...
         //TODO: add this again later or remove, because these checks are already made in some other place
-        if (this.getFrom().address().length != Account.ADDRESS_SIZE)
-            throw new TransactionSemanticValidityException("Cannot create signed transaction without valid from address");
+        try {
+            if (this.getFrom().address().length != Account.ADDRESS_SIZE)
+                throw new TransactionSemanticValidityException("Cannot create signed transaction without valid from address");
+        } catch (org.web3j.crypto.exception.CryptoWeb3jException e) {
+            // for instance badly constructed tx having wrong type
+            throw new TransactionSemanticValidityException("Cannot create signed transaction without valid from address: " + e.getMessage());
+        }
```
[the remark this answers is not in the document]
JcBernack: Probably in the `getFrom()` implementation which tries to rebuild the address from the signature.

sdk/src/main/scala/com/horizen/account/state/StateTransition.scala
JcBernack: This was moved here to make sure automatic gas consumption is only active while the message processor is executing. Afterwards we don't want it anymore. Before this change it could cause an `OutOfGas` exception while doing `addBalance` in `refundGas`. Doesn't matter really, because for [EVM-274](https://horizenlabs.atlassian.net/browse/EVM-274?atlOrigin=eyJpIjoiZmVhZjdkZjYwYWU3NDcwOWFhYzdiMDg2NDFjOGNhNGIiLCJwIjoiaiJ9) this will be removed.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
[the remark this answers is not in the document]
alsala: Correct, it is in the `EthereumTransaction::getFromAddress()` method

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -127,11 +128,10 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
   def sendCoinsToAddress: Route = (post & path("sendCoinsToAddress")) {
     entity(as[ReqSendCoinsToAddress]) { body =>
       // lock the view and try to create EvmTransaction
-      // TODO also account for gas fees
       applyOnNodeView { sidechainNodeView =>
         val valueInWei = ZenWeiConverter.convertZenniesToWei(body.value)
         val destAddress = body.to
-        val gasPrice = sidechainNodeView.getNodeState.getBaseFee // TODO actual gas implementation
+        val gasPrice = sidechainNodeView.getNodeState.getBaseFeePerGas.add(BigInteger.valueOf(200))
```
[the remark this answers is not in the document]
alsala: No special reason, it is a default value I choose just to be above the base fee. This code will be changed as soon as a it will be meged with Sara's branch which adds specific parameters to this command:
```
        if (body.gasInfo.isDefined) {
          gasPrice = body.gasInfo.get.maxFeePerGas
          gasLimit = body.gasInfo.get.gasLimit
        }
```

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -298,17 +302,21 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
       applyOnNodeView { sidechainNodeView =>
         val valueInWei = ZenWeiConverter.convertZenniesToWei(body.forgerStakeInfo.value)
 
-        // TODO actual gas implementation
-        var maxFeePerGas = BigInteger.ONE
-        var maxPriorityFeePerGas = BigInteger.ONE
-        var gasLimit = BigInteger.ONE
+        // default gas related params
+        val baseFee = sidechainNodeView.getNodeState.getBaseFeePerGas
+        var maxPriorityFeePerGas = BigInteger.valueOf(120)
```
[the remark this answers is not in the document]
alsala: Ok, let's discuss it

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -110,17 +118,27 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
           receiptList += consensusDataReceipt
           txHashList += txHash
 
+          val txGasUsed = consensusDataReceipt.cumulativeGasUsed.subtract(cumGasUsed)
+          // update cumulative gas used so far
+          cumGasUsed = consensusDataReceipt.cumulativeGasUsed
+
+          // legacy TXes will have just the second contribution, given by user set gasPrice.
+          // EIP1559 TXes have both.
+          val (txBaseFeePerGas, txForgerTipPerGas) = stateView.getTxFeesPerGas(ethTx)
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -141,36 +159,61 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
...
+    // 1. As forger address take first address from the wallet
     val addressList = nodeView.vault.secretsOfType(classOf[PrivateKeySecp256k1])
     if (addressList.size() == 0)
       throw new IllegalArgumentException("No addresses in wallet!")
 
     val forgerAddress = addressList.get(0).publicImage().asInstanceOf[AddressProposition]
 
+    // 2. create a disposable view and try to apply all transactions in the list, collecting all data needed for
+    //    going on with the forging of the block
+    val (stateRoot, receiptList, appliedTxList, feePayments)
```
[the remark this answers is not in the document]
alsala: You are right, I added the payment in this code so to update the state root

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -187,7 +230,7 @@ class AccountForgeMessageBuilder(mainchainSynchronizer: MainchainSynchronizer,
       stateRoot,
       receiptsRoot,
       forgerAddress,
-      baseFee,
+      baseFee.longValue(), // TODO to be changed into biginteger?
```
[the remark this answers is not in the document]
alsala: Ok

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -167,16 +163,23 @@ class AccountState(val params: NetworkParams,
...
 
       // check stateRoot and receiptRoot against block header
       mod.verifyReceiptDataConsistency(receiptList.map(_.consensusDataReceipt))
 
       val stateRoot = stateView.getIntermediateRoot
       mod.verifyStateRootDataConsistency(stateRoot)
 
+      // If SC block has reached the end of the withdrawal epoch reward the forgers.
+      evalForgersReward(mod, modWithdrawalEpochInfo, stateView)
+
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/state/AccountStateReader.scala
```diff
@@ -21,6 +22,8 @@ trait AccountStateReader extends BaseStateReader {
 
   def getHeight: Int
 
-  def getBaseFee: BigInteger
+  def getBaseFeePerGas: BigInteger
   def getBlockGasLimit: BigInteger
+  def getTxFeesPerGas(tx: EthereumTransaction) : (BigInteger, BigInteger)
```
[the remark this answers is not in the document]
alsala: Ok

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
alsala: You are right, we do not need it

sdk/src/main/scala/com/horizen/account/state/GasPool.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
alsala: Let's discuss it

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
alsala: Ok added a new `AccountStateTest.scala` file

qa/sc_evm_forging_fee_payments.py
alsala: Yes correct, I added it in PR #531

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
i-Alex: let's add todo

alsala: Done

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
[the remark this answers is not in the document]
alsala: We agreed to start a new task. It would be better to have mandatory arguments `body.gasInfo.*` for this API

i-Alex (approved):

PR #530, opened 2022-09-08
Fix/crypto lib inconsistency fix

i-skrypnyk:


sdk/src/main/scala/com/horizen/block/SidechainCommitmentTree.scala
```diff
@@ -78,8 +78,8 @@ class SidechainCommitmentTree {
       btrList.toArray,
       certificate.customFieldsOpt(version).asJava,
       certificate.endCumulativeScTxCommitmentTreeRoot,
+      certificate.ftMinAmount, //TODO: temporal dirty porkaround until SC zendoo_commitment_tree_add_cert is fixed
```
i-Alex, line 81: let's add a bit more info about the porkaround:
that we swapped the order of the `ftMinAmount` and `btrFee` and now setting them in the wrong order to be consistent with the wrong order on MC side

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/block/SidechainCommitmentTree.scala
i-skrypnyk: improved todo comment

VolodymyrBabaiev (approved):

PR #531, opened 2022-09-08
As/fee payments api

alsala:
Added `block/getFeePayments` API http method and modified py test making use of it. In the implementation a new template parameter has been added to some class hierarchies for generalizing some code between utxo and account models.
Moreover the check on block size limit and number of txes in a block has been removed.


sdk/src/main/scala/com/horizen/chain/AbstractFeePaymentsInfo.scala
```diff
@@ -0,0 +1,8 @@
+package com.horizen.chain
+
+import scorex.core.serialization.{BytesSerializable, ScorexSerializer}
+
+abstract class AbstractFeePaymentsInfo() extends BytesSerializable {
```
i-Alex, line 5: We can keep it as trait: `trait FeePaymentsInfo extends BytesSerializable`

sdk/src/main/scala/com/horizen/storage/AbstractHistoryStorage.scala
```diff
@@ -271,9 +274,7 @@ class AbstractHistoryStorage[PM <: SidechainBlockBase[_ <: Transaction, _ <: Sid
     this
   }
 
-  def getFeePaymentsInfo(blockId: ModifierId): Option[FeePaymentsInfo] = {
-    storage.get(feePaymentsInfoKey(blockId)).asScala.flatMap(baw => FeePaymentsInfoSerializer.parseBytesTry(baw.data).toOption)
-  }
+  def getFeePaymentsInfo(blockId: ModifierId): Option[FPI]
```
i-Alex, line 277: Probably we may pass `PaymentsInfoSerializer` to the constructor, similar to how we passed `blockSerializer: ScorexSerializer[PM],`
then will have a generic implementation as it was before.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/AccountSidechainNodeViewHolder.scala
```diff
@@ -96,16 +100,35 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
...
+
+    val epochInfo = state.getWithdrawalEpochInfo
+    val isWithdrawalEpochLastIndex: Boolean = WithdrawalEpochUtils.isEpochLastIndex(epochInfo, params)
+    if (isWithdrawalEpochLastIndex) {
+      val epochNumber: Int = epochInfo.epoch
+      val feePayments = state.getFeePayments(epochNumber)
+      if (feePayments.nonEmpty) {
+        val forgersPoolRewardsSeq: Seq[AccountPayment] = AccountFeePaymentsUtils.getForgersRewards(feePayments)
+        log.debug(s"updating history with ${feePayments.size} fee payments to ${forgersPoolRewardsSeq.size} forgers for withdrawal epoch $epochNumber")
+        val historyAfterUpdateFee = history.updateFeePaymentsInfo(modToApply.id, AccountFeePaymentsInfo(forgersPoolRewardsSeq))
```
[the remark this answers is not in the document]
alsala: I moved the implementation in the base abstract class by adding some helper methods to be override by the concrete classes

sdk/src/main/scala/com/horizen/storage/AbstractHistoryStorage.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
```diff
@@ -157,23 +158,22 @@ class SidechainNodeViewHolder(sidechainSettings: SidechainSettings,
...
+  override def getFeePaymentsInfo(state: MS, epochNumber: Int) : FPI = {
+    val feePayments = state.getFeePayments(epochNumber)
+    SidechainFeePaymentsInfo(feePayments)
+  }
 
-      (historyAfterUpdateFee, walletAfterApply)
-    } else {
-      val walletAfterApply: VL = wallet.scanPersistent(modToApply, stateWithdrawalEpochNumber, Seq(), None)
-      (history, walletAfterApply)
+  def getScanPersistentWallet(modToApply: SidechainBlockBase[_ <: Transaction, _ <: SidechainBlockHeaderBase], stateOp: Option[MS], epochNumber: Int, wallet: VL) : VL = {
```
i-Alex, line 166: 1) add `override`
2)`SidechainBlockBase[_ <: Transaction, _ <: SidechainBlockHeaderBase]` -> `SidechainBlock` . So, no need to case `.asInstanceOf[SidechainBlock]`.
same for account model

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
alsala: Done

i-Alex (approved):

PR #532, opened 2022-09-09
Fix for reverting the state db when unexpected exceptions occur.

saratnt:


sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -151,25 +151,43 @@ class AccountStateView(
...
+        // make sure we disable automatic gas consumption in case a message processor enabled it
+        disableGasTracking()
+      }
+      val consensusDataReceipt = new EthereumConsensusDataReceipt(
+        ethTx.version(), status.id, blockGasPool.getUsedGas, getLogs(txHash))
+      log.debug(s"Returning consensus data receipt: ${consensusDataReceipt.toString()}")
+      consensusDataReceipt
+    }
+    catch {
+      case e : Exception =>
```
JcBernack, line 177: This exception case could be moved next to the other. All the exception will be coming from `applyMessage` anyway.

JcBernack (changes requested):
I don't know if I really understand why we need this. Why do we need to be able to salvage the situation after applying a transactions that was so invalid, that it failed with an exception that was not an `ExecutionFailedException`? A transaction like that should have failed validation before being applied.

If we do need it the changes are good.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -151,25 +151,43 @@ class AccountStateView(
...
+          ReceiptStatus.FAILED
+      } finally {
+        // finalize pending changes, clear the journal and reset refund counter
+        // make sure we disable automatic gas consumption in case a message processor enabled it
+        disableGasTracking()
+      }
+      val consensusDataReceipt = new EthereumConsensusDataReceipt(
+        ethTx.version(), status.id, blockGasPool.getUsedGas, getLogs(txHash))
+      log.debug(s"Returning consensus data receipt: ${consensusDataReceipt.toString()}")
+      consensusDataReceipt
```
i-Alex, line 174: what about the `null` logs case here? I think we need to solve this issue first of all.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
i-Alex: Instead of wrapping the whole code base, I propose to add one more parameter to the `applyTransaction` methods: `finalizeChanges: Boolean` that has default value `true`.
Forger will:
1) create a snapshot
2) execute`applyTransaction` with `finalizeChanges=false`
3) in case applied tx doesn't feet the block -> do revertToSnapshot. Otherwise, do `stateDb.finalizeChanges()`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
saratnt: Do you mean the null value returned from the StateDb? If it is, I have fixed it in the orphan txs branch.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
saratnt: @JcBernack: This fix was made because there was a bug in the EthereumConsensusDataReceipt.toString method that caused a NullPointerException. Because of that exception, the tx was not included in the block by the forger but the state wasn't reverted, so the rootstatehash in the block was wrong etc. etc. Now, the root cause of the NullPointerException is fixed and I agree with you that, except for bugs, there are no real possible causes of unexpected exceptions here. Still I think that it is better to be safe than sorry and to protect all the code. Consider also that maybe in the future new, more "dangerous" line of codes could be added in the same place and maybe one could forget to protect them.

saratnt: @i-Alex: I thought that it was a responsibility of AccountStateView class to guarantee that the state remained consistent, regardless of what the caller would have done next. But if you think it is better to leave the forger to call finalizeChanges, it's ok for me.

i-Alex (approved):

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -102,44 +102,55 @@ class AccountForgeMessageBuilder(
...
+            log.trace(s"Could not apply tx, reason: ${e.getMessage}")
+            return Success(receiptList, txHashList, cumBaseFee, cumForgerTips)
+          case Failure(e: NonceTooLowException) =>
+            //SHOULD NEVER HAPPEN, but in case just skip this tx
+            log.error(s"******** Could not apply tx for NonceTooLowException ******* : ${e.getMessage}")
+          case Failure(e) =>
+            // skip all txs from the same account but remove any changes caused by the rejected tx
+            log.warn(s"Could not forge tx, reason: ${e.getMessage}", e)
+            accountsToSkip += tx.getFrom
+            stateView.revertToSnapshot(revisionId)
```
JcBernack, line 144: This revert is the only reason we need the `finalizeChanges` switch on `applyTransaction`, correct?
Is there a reason we can't make `applyTransaction` revert everything on its own if this occurs?

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -25,19 +36,27 @@ import scala.collection.JavaConverters.collectionAsScalaIterableConverter
...
 
 class AccountStateView(
-  metadataStorageView: AccountStateMetadataStorageView,
-  stateDb: StateDB,
-  messageProcessors: Seq[MessageProcessor])
-  extends StateView[SidechainTypes#SCAT]
+    metadataStorageView: AccountStateMetadataStorageView,
+    stateDb: StateDB,
+    messageProcessors: Seq[MessageProcessor]
+) extends StateView[SidechainTypes#SCAT]
```
JcBernack, line 42: This is a file reformat, but not with the Scalafmt settings, if I'm not mistaken. If we do reformatting then with common settings please.

JcBernack (changes requested):

sdk/src/test/scala/com/horizen/account/fixtures/EthereumTransactionFixture.scala
```diff
@@ -12,9 +12,11 @@ trait EthereumTransactionFixture {
   def createLegacyTransaction(value: BigInteger,
                               nonce: BigInteger = BigInteger.ZERO,
                               pairOpt: Option[ECKeyPair] = None,
-                              gasPrice: Option[BigInteger] = None): EthereumTransaction = {
+                              gasPrice: Option[BigInteger] = None,
+                              gasLimit: Option[BigInteger] = None): EthereumTransaction = {
     val gasFee = gasPrice.getOrElse(value)
-    val rawTransaction = RawTransaction.createTransaction(nonce, gasFee, value, "0x", value, "")
+    val limit = gasLimit.getOrElse(value)
```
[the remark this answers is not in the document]
saratnt: Yes, you're right. This is something left from the days when the gas management was still missing. 
Removed.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
saratnt: True. The first implementation actually was as you suggest but after a comment from Sasha, it was changed as it is now.
The rational for this change is to let the forger decide when to call finalizeChanges.

JcBernack: Ok, good.

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
saratnt: After your comment about Scalafmt, I changed the formatter in Intellij in order to use Scalafmt but apparently something went wrong with this file.

i-Alex (approved):
Good!

sdk/src/test/scala/com/horizen/account/fixtures/EthereumTransactionFixture.scala
```diff
@@ -1,30 +1,32 @@
...
+    val rawTransaction = RawTransaction.createTransaction(nonce, gasPrice, gasLimit, "0x", value, "")
     createSignedTransaction(rawTransaction, pairOpt)
   }
 
   def createEIP1559Transaction(value: BigInteger,
                                nonce: BigInteger = BigInteger.ZERO,
                                pairOpt: Option[ECKeyPair] = None,
-                               gasFee: BigInteger = BigInteger.ONE,
-                               priorityGasFee: BigInteger = BigInteger.ONE): EthereumTransaction = {
+                               gasFee: BigInteger = BigInteger.valueOf(10000),
```
JcBernack, line 25: This is the `gasFeeCap` right? :sweat_smile:

JcBernack (approved):

sdk/src/test/scala/com/horizen/account/fixtures/EthereumTransactionFixture.scala
saratnt: Yes

PR #533, opened 2022-09-09
Release 0.4.2 

i-Alex:
Changes:
1. Explorer synchronization issue solved: `history.chainAfter` method was optimized. PR https://github.com/HorizenOfficial/Sidechains-SDK/pull/523
2. Certificate commitment tree calculation issue solved: fixed an inconsistency between SC and MC implementations. PR https://github.com/HorizenOfficial/Sidechains-SDK/pull/530


paolocappelletti (approved):

sdk/src/main/scala/com/horizen/block/SidechainCommitmentTree.scala
```diff
@@ -71,15 +71,18 @@ class SidechainCommitmentTree {
       new BackwardTransfer(btrOutput.pubKeyHash, btrOutput.amount)
     )
 
+    //TODO: temporal dirty porkaround until SC zendoo_commitment_tree_add_cert is fixed.
```
MarcoOl94, line 74: Do we want to call it "workaround"?

MarcoOl94 (approved):

sdk/src/main/scala/com/horizen/block/SidechainCommitmentTree.scala
i-Alex: No, it is a "porkaround" :smile:

PR #535, opened 2022-09-14
Rpc debug methods

GitMark0:
Added support for `debug_traceTransaction` and `debug_traceTransaction`. 
Return type and functionality of mentioned methods is described here: https://geth.ethereum.org/docs/rpc/ns-debug


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -348,32 +386,123 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
...
+        }
+      }
+    }
+  }
+
+  @RpcMethod("debug_traceTransaction")
+  def traceTransaction(transactionHash: String): DebugTraceTransactionView = {
+    val requestedTransaction = getTransactionAndReceipt(transactionHash) { (tx, receipt) =>
+      new EthereumTransactionView(receipt, tx)
+    }.orNull
```
i-Alex, line 459: in case transaction not found we need to return proper result message.
Otherwise we will throw an null pointer exception.
Let's add the case with unknown tx id in the python test.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -348,32 +386,123 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
...
+                    tagStateView.getStateDbHandle,
+                    if (tx.getFrom == null) null
+                    else tx.getFrom.bytes(),
+                    if (tx.getTo == null) null
+                    else tx.getTo.bytes(),
+                    tx.getValue,
+                    tx.getData,
+                    tx.getGasLimit,
+                    tx.getGasPrice,
+                    null
```
i-Alex, line 492: why we keep `null` context?
We may specify the proper one as it is in the state. See the latest changes by JC.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -348,32 +386,123 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
...
+
+          for (mcBlockRefData <- currentBlock.mainchainBlockReferencesData) {
+            tagStateView.applyMainchainBlockReferenceData(mcBlockRefData).get
+          }
+
+          breakable {
+            for ((tx, i) <- transactions.zipWithIndex) {
+              if (tx.id == Numeric.cleanHexPrefix(transactionHash)) {
+                val txResult = Evm
+                  .Trace(
```
i-Alex, line 482: let's discuss the usage of it.
It seems that we miss intrinsic gas and other important parts for the tx we want to trace.
It means that trace results are inconsistent to the real results.
We may add trace arg to the applyMessage do then pass it further.
the results (traceLogs) can be put to the `EthereumConsensusDataReceipt` as in memory only data.

evm/src/main/java/com/horizen/evm/Evm.java
```diff
@@ -19,18 +19,19 @@ public static EvmResult Apply(
             BigInteger gasPrice,
             EvmContext context
     ) {
-        return LibEvm.evmApply(stateDBHandle.handle, from, to, value, input, gasLimit, gasPrice, context);
+        return LibEvm.evmApply(stateDBHandle.handle, from, to, value, input, gasLimit, gasPrice, context, false);
     }
 
-    public static EvmResult Apply(
+    public static EvmResult Trace(
```
i-Alex, line 25: considering the comments below, we may have only `Apply` method with `boolean trace` arg there.

i-Alex (changes requested):

evm/src/main/java/com/horizen/evm/interop/EvmParams.java
```diff
@@ -9,11 +9,13 @@ public class EvmParams extends HandleParams {
     public Address to;
     public BigInteger value;
     public byte[] input;
-    public BigInteger gasLimit; // uint64
+    public BigInteger gas; // uint64
```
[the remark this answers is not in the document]
JcBernack: The difference between `gas` and `gasLimit` is subtle: The "gas limit" is provided from the outside, either via a transaction or during an RPC call. After the "buy gas" step during a state transition we have a certain amount of "gas" available that we consume until it's used up (out of gas) or we're done processing. At this stage in the process we are behind the "buy gas" step, thus using `gas` instead of `gasLimit` here is more accurate. Also this is not the original gas limit anymore as intrinsic gas was already paid.

This also reduces confusion with the `gasLimit` that is still existing on the context, which is the block gas limit, not the TX gas limit.

libevm/lib/service_evm.go
```diff
@@ -181,9 +186,10 @@ func (s *Service) EvmApply(params EvmParams) (error, *EvmResult) {
 	}
 
 	return nil, &EvmResult{
-		UsedGas:         uint64(params.GasLimit) - gas,
+		UsedGas:         uint64(params.Gas) - gas,
```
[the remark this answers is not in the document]
JcBernack: See above, `params.gas` is original available gas, `gas` is the available gas at the end.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -348,32 +386,123 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
...
+                    tagStateView.getStateDbHandle,
+                    if (tx.getFrom == null) null
```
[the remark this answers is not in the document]
JcBernack: True, it cannot be null here. During `eth_call` and `eth_estimateGas` it can be, but this comes from a real block.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
JcBernack: Agree, in libevm it already is the same thing, there is just `evmApply` with trace on or off. We need a way enable/disable tracing for MessageProcessors and to allow MessageProcessors to output trace logs.

evm/src/main/java/com/horizen/evm/interop/EvmParams.java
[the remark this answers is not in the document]
i-Alex: thanks for the explanation.
I propose to give more context to the names:
* `params.gas` -> `params.availableGas` (or even better name)
* `gas` as a result of apply -> `gasLeft`
In this way `gasUsed = availableGas - gasLeft` seems more clear than `gasUsed = params.Gas - gas`

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: As was decided AFK:
1) We need to use `stateView.applyTransaction(...)` interface in case of both normal apply and trace logic. That will allow us to be applicable to any kind of MessageProcessor and replay Tx/Block exactly in the same way as it was applied to the State.
2) We need to extend BlockContext with the `Tracer` class. that may haw enabled and disabled statuses.
In case of "enabled", EvmMessagerProcessor will call EVM.Apply(....tracer=true) and will store the TraceLogResults into the Tracer. That will allow us to get them here and then construct `evmResult`.

i-Alex (changes requested):

evm/src/main/java/com/horizen/evm/interop/TraceParams.java
```diff
@@ -0,0 +1,33 @@
+package com.horizen.evm.interop;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class TraceParams {
+    @JsonProperty("enableMemory")
```
JcBernack, line 9: Imho, all the annotations in this class are not necessary, except for `@JsonIgnoreProperties(ignoreUnknown = true)`.

evm/src/main/java/com/horizen/evm/interop/TraceParams.java
```diff
@@ -0,0 +1,33 @@
...
+    final boolean enableMemory;
+    @JsonProperty("disableStack")
+    final boolean disableStack;
+    @JsonProperty("disableStorage")
+    public boolean disableStorage;
+    @JsonProperty("enableReturnData")
+    public boolean enableReturnData;
```
JcBernack, line 16: Shouldn't all properties be both `public` and `final`?

libevm/lib/service_evm.go
```diff
@@ -180,10 +193,31 @@ func (s *Service) EvmApply(params EvmParams) (error, *EvmResult) {
 		evmError = vmerr.Error()
 	}
 
+	traceLogs := geth_internal.FormatLogs(tracer.StructLogs())
+	if params.Trace == true {
+
+		if params.TxTraceParams.DisableStack == true {
```
JcBernack, line 199: This is backwards, the parameters need to be set on the tracer, see the other comment above.

libevm/lib/service_evm.go
```diff
@@ -129,17 +141,18 @@ func (s *Service) EvmApply(params EvmParams) (error, *EvmResult) {
 		}
 		blockContext = params.getBlockContext()
 		chainConfig  = defaultChainConfig()
+		tracer       = logger.NewStructLogger(nil)
```
JcBernack, line 144: The trace parameters need to be passed to the tracer here to have an effect. I recommend adding a function to the `TraceParams` struct to produce a geth-compatible type, similar to what `getBlockContext` does and then do:
`tracer = logger.NewStructLogger(params.TxTraceParams.getTraceConfig())` here. Something like:
```
func (t *TraceParams) getTraceConfig() logger.Config {
	return logger.Config{
		EnableMemory:     t.EnableMemory,
		DisableStack:     ...,
		DisableStorage:   ...,
		EnableReturnData: ...,
	}
}
```

sdk/src/main/java/com/horizen/account/api/rpc/types/DebugTraceBlockByIdView.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.fasterxml.jackson.annotation.JsonView;
+import com.horizen.evm.interop.EvmResult;
+import com.horizen.serialization.Views;
+
+import java.util.Arrays;
+
+@JsonView(Views.Default.class)
+public class DebugTraceBlockByIdView {
+    public Object[] debugTraceTransactionViews;
```
JcBernack, line 11: Why not use `DebugTraceTransactionView` instead of `Object` and add a type cast below?

sdk/src/main/java/com/horizen/account/api/rpc/types/DebugTraceTransactionView.java
```diff
@@ -0,0 +1,22 @@
...
+import com.horizen.utils.BytesUtils;
+import org.web3j.utils.Numeric;
+
+import java.util.Arrays;
+
+@JsonView(Views.Default.class)
+public class DebugTraceTransactionView {
+    public String gas;
+    public String returnValue;
+    public Object[] structLogs;
```
JcBernack, line 15: Same here, use proper type and add type cast if possible.

JcBernack (changes requested):

evm/src/main/java/com/horizen/evm/interop/EvmParams.java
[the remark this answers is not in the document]
i-Alex: still actual

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: still actual

i-Alex (changes requested):
Agree with the comments by @JcBernack .
Beside that:
1) check older comments, some of them await for fixes.
2) We need a proof that this new RPC methods are able to work. For this let's introduce a new python test, where to test both positive and negative cases.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -464,13 +464,21 @@ class EthService(
   def traceTransaction(transactionHash: String, traceParams: TraceParams): Object = {
     val requestedTransaction = getTransactionAndReceipt(transactionHash) { (tx, receipt) =>
       new EthereumTransactionView(receipt, tx)
-    }.orNull
+    } match {
+      case Some(tx) => tx
+      case None =>
+        return ApiResponseUtil.toResponse(
```
JcBernack, line 470: We need to throw an `RpcException` here, any "normal" return value would get sent back as a success result.

i-Alex (approved):

libevm/lib/service_evm.go
```diff
@@ -24,14 +33,16 @@ type EvmContext struct {
...
-	Input      []byte           `json:"input"`
-	GasLimit   hexutil.Uint64   `json:"gasLimit"`
-	GasPrice   *hexutil.Big     `json:"gasPrice"`
-	AccessList types.AccessList `json:"accessList"`
-	Context    EvmContext       `json:"context"`
+	From          common.Address   `json:"from"`
+	To            *common.Address  `json:"to"`
+	Value         *hexutil.Big     `json:"value"`
+	Input         []byte           `json:"input"`
+	Gas           hexutil.Uint64   `json:"gas"`
```
JcBernack, line 40: You renamed this on the Java side to `availableGas`, need to change this here too otherwise it always use `MaxInt64`.

JcBernack (changes requested):

libevm/lib/service_evm.go
```diff
@@ -37,7 +37,7 @@ type EvmParams struct {
 	To            *common.Address  `json:"to"`
 	Value         *hexutil.Big     `json:"value"`
 	Input         []byte           `json:"input"`
-	Gas           hexutil.Uint64   `json:"gas"`
+	Gas           hexutil.Uint64   `json:"availableGas"`
```
JcBernack, line 40: I'm sorry, now it is working, but really confusing. Please rename the variable name too, not just the json property.

JcBernack (changes requested):

JcBernack (approved):

evm/src/main/java/com/horizen/evm/interop/EvmParams.java
```diff
@@ -9,11 +9,13 @@ public class EvmParams extends HandleParams {
     public Address to;
     public BigInteger value;
     public byte[] input;
-    public BigInteger gasLimit; // uint64
+    public BigInteger availableGas; // uint64
```
JcBernack, line 12: There is build errors now:
```
[ERROR] # libevm/interop [libevm/interop.test]
[ERROR] interop/invoke_test.go:65:3: unknown field 'Gas' in struct literal of type lib.EvmParams
[ERROR] interop/invoke_test.go:86:3: unknown field 'Gas' in struct literal of type lib.EvmParams
[ERROR] interop/invoke_test.go:95:3: unknown field 'Gas' in struct literal of type lib.EvmParams
```

JcBernack (changes requested):

i-Alex:
@GitMark0  check `EvmTest` results. There is an error

evm/src/main/java/com/horizen/evm/interop/EvmParams.java
GitMark0: fixed

JcBernack (approved):

i-Alex (approved):

PR #537, opened 2022-09-15
update forger block gas limit logic

JcBernack:
Include new transactions as long as there is at least 21.000 gas available in the block.


i-Alex (approved):
We will for sure add some Python tests later for this case merging all the changes pending

i-Alex:
@JcBernack please resolve the conflict with `dev_evm`. Then I will merge the PR

i-Alex (approved):

PR #544, opened 2022-09-22
add forward transfer view, add forward transfer api

akrckn:
- [x] add forward transfer rpc api method
- [x] add forward transfer api method
- [x] move forward transfer api to `block` route
  - [x] remove 0x prefix in to address
  - [x] implement decimal representation of value
- [x] simplify duplicate forward transfer filtering

TODO:

- [x] add more tests
- [x] Seems like we are not getting the correct address in forward transfer object or it is falsely formatted


sdk/src/main/java/com/horizen/account/api/rpc/types/ForwardTransfersView.java
```diff
@@ -0,0 +1,61 @@
...
+                value = String.valueOf(weiValue);
+            } else {
+                to = Numeric.toHexString(address.address());
+                value = Numeric.toHexStringWithPrefix(weiValue);
+            }
+            forwardTransfers.put(i, new ForwardTransferData(to, value));
+        }
+    }
+
+    public TreeMap<Integer, ForwardTransferData> getForwardTransfers() {
```
i-Alex, line 39: why do we need to use Map here?
We pass `List<ForwardTransfer> transactions` -> so the result can be List<ForwardTransferData> to be sure that the result has a proper order of FTs.

i-Alex (approved):

alsala (approved):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -384,4 +385,15 @@ class EthService(val scNodeViewHolderRef: ActorRef, val nvtimeout: FiniteDuratio
       }
     }
   }
+
+  @RpcMethod("eth_getForwardTransfers")
```
JcBernack, line 389: Actually, I think we should call this `zen_getForwardTransfers`.
Moreover, the idea was to have all the `eth_` methods in `EthService` not *all* RPC methods - but that is refactoring for later.

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -4,7 +4,7 @@ import org.web3j.crypto._
 import akka.actor.{ActorRef, ActorRefFactory}
 import akka.http.scaladsl.server.Route
 import akka.pattern.ask
-import com.fasterxml.jackson.annotation.JsonView
+import com.fasterxml.jackson.annotation.{JsonView}
```
JcBernack, line 7: press `optimize imports` :wink:

JcBernack (approved):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: so far, let's keep it as is.
Agree to keep the new methods with another prefix.

PR #546, opened 2022-09-26
Block level bloom filter

GitMark0:


sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -133,24 +134,24 @@ class AccountForgeMessageBuilder(
...
+      sidechainTransactions: Seq[SidechainTypes#SCAT],
+      mainchainHeaders: Seq[MainchainHeader],
+      ommers: Seq[Ommer[AccountBlockHeader]],
+      ownerPrivateKey: PrivateKey25519,
+      forgingStakeInfo: ForgingStakeInfo,
+      vrfProof: VrfProof,
+      forgingStakeInfoMerklePath: MerklePath,
+      companion: DynamicTypedSerializer[SidechainTypes#SCAT, TransactionSerializer[SidechainTypes#SCAT]],
+      inputBlockSize: Int,
+      logsBloom: LogsBloom,
```
i-Alex, line 152: we don't need to pass `logsBloom` as an argument here, because:
1) we don't know the value yet
2) it is applicable to the account model only.

Instead inside this method in this class we need to calculate a valid bloom filter using the real set of receipts (see `receiptList`)

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
GitMark0: Good, so it needs to be calculated in 2 places? Here and in `AccountState` (it is already calculated there)-

i-Alex: right

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
```diff
@@ -20,25 +25,24 @@ class LogsBloom() {
...
     bloomFilterIndexes.zipWithIndex
       .foreach({ case (bloomFilterIndex, i) =>
-        bloomFilter(bloomFilterIndex) =
-          (bloomFilter(bloomFilterIndex) | bloomFilterValues(i)).toByte
+        bloomFilter(bloomFilterIndex) = (bloomFilter(bloomFilterIndex) | bloomFilterValues(i)).toByte
       })
   }
 
-  def addBloomFilter(bloomFilter: LogsBloom): Unit = {
+  def addBloomFilter(bloomFilter: LogsBloom): LogsBloom = {
```
i-Alex, line 37: Let's keep the result type as Unit.
Having `LogsBloom` lead to the idea that addBloomFilter remain the existing instance immutable - which is not our case.
In the forger reuse `fromEthereumReceipt` that should take `EthereumConsensusDataReceipt` seq

GitMark0: Ok, I added that because it allowed me to use reduce on `EthereumReceipt` seq that was created by `map` from `EthereumConsensusDataReceipt` seq. Should I add another method called `fromEthereumConsensusDataReceipt` to the `LogsBloom` or I should extract `LogsBloom` seq from `EthereumConsensusDataReceipt` to be able to use `fromEthereumReceipt`. Also, approach would be to initialize a new empty `LogsBloom` and add receipts (by calling `addBloomFilter`), one by one to it.

i-Alex: "or I should extract LogsBloom seq from EthereumConsensusDataReceipt to be able to use fromEthereumReceipt. After that, approach would be to initialize a new empty LogsBloom and add receipts (by calling addBloomFilter), one by one to it." I vote for this one

qa/sc_evm_test_block_bloom_filter.py
```diff
@@ -0,0 +1,381 @@
...
+                                             gasLimit=estimated_gas)
+    print("Generating next block...")
+    generate_next_blocks(node, "first node", 1)
+    tx_receipt = node.rpc_eth_getTransactionReceipt(tx_hash)
+    assert_equal(format_evm(tx_receipt['result']['contractAddress']), format_evm(address))
+    print("Smart contract deployed successfully to address 0x{}".format(address))
+    return address
+
+
+class SCEvmDebugMethods(SidechainTestFramework):
```
i-Alex, line 227: check the name, seems irrelevant

qa/sc_evm_test_block_bloom_filter.py
```diff
@@ -0,0 +1,381 @@
...
+
+Test:
+    For the smart contract:
+        - Deploy the smart contract without initial data
+        - Check initial minting success (static_call)
+        - Check initial supply (static_call)
+        - Check a successful transfer (static_call + actual tx)
+        - Check a reverting transfer (static_call + actual tx)
+        - Check approval + transferFrom (static_call + actual tx)
+"""
```
i-Alex, line 39: update test description

i-Alex (changes requested):
Minor changes in the py test requeired.
Beside that - seems good

qa/sc_evm_test_block_bloom_filter.py
```diff
@@ -224,7 +217,7 @@ def deploy_erc20_smart_contract(node, smart_contract, from_address):
     return address
 
 
-class SCEvmDebugMethods(SidechainTestFramework):
+class SCEvmBlockBloomFilter(SidechainTestFramework):
```
i-Alex, line 220: check Main function

PR #549, opened 2022-09-27
Truffle fixes

JcBernack:


libevm/contracts/ForgerStakes.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0
+// SPDX-License-Identifier: MIT
```
i-Alex, line 1: a comment for the `struct Stake` below.
It is a bit inconsistent with the actual one: vrfPublicKey is 33 bytes

sdk/src/main/scala/com/horizen/account/receipt/EthereumLog.scala
```diff
@@ -1,60 +0,0 @@
...
-
-import com.horizen.evm.interop.EvmLog
-import com.horizen.evm.utils.{Address, Hash}
-import com.horizen.utils.{ByteArrayWrapper, BytesUtils}
-import scorex.core.serialization.ScorexSerializer
-import scorex.util.serialization.{Reader, Writer}
-
-import java.util
-
-case class EthereumLog(
```
i-Alex, line 11: @alsala @JcBernack let's discuss the usage of this class.

libevm/contracts/ForgerStakes.sol
JcBernack: True, but unfortunately there is no type `bytes33` in Solidity. The ABI serializer on our end also handles it like that, see [com.horizen.account.state.ForgerPublicKeys#vrfPublicKeyToAbi](https://github.com/HorizenOfficial/Sidechains-SDK/blob/d1180fd0160dd4adbd65de570fb2a3e293d8b2ea/sdk/src/main/scala/com/horizen/account/state/ForgerStakeMsgProcessor.scala#L460).

JcBernack: Oops, I mixed up the files. This is just my draft for the Forger Stakes contract before we decided to implement them in Java. This is not used anywhere and could be removed.

JcBernack: I was talking about the Forger Stakes Contract interface in Solidity where the vrf public key is separated to a `bytes32` property and then one more `bytes1`.

sdk/src/main/scala/com/horizen/account/state/ForgerStakeMsgProcessor.scala
```diff
@@ -108,18 +116,8 @@ case class ForgerStakeMsgProcessor(params: NetworkParams) extends FakeSmartContr
...
-      val modPreviousNode = LinkedListNode(
-        previousNode.dataKey,
-        previousNode.previousNodeKey,
-        newTip
-      )
-
-      // store the modified previous node
-      view.updateAccountStorageBytes(contractAddress, oldTip,
-        LinkedListNodeSerializer.toBytes(modPreviousNode))
+    modifyNode(view, oldTip) { previousNode =>
```
i-Alex, line 119: nice refactoring!

i-Alex (approved):
Seems good!
Note: just a comment about the outdated fake smart contract solidity code. Let's remove them at all.

libevm/contracts/ForgerStakes.sol
JcBernack: done

i-Alex (approved):

akrckn (approved):

PR #559, opened 2022-10-04
Refactoring

JcBernack:
- refactor `BigInteger` usages:
  - `compareTo(BigInteger.Zero)` => `signum()` same result, better performance
  - `longValue()` => `longValueExact()` make sure we throw an exception instead of working with unexpected values
  - `intValue()` => `intValueExact()` same
- refactor base fee calculation
  - add more test cases to make sure the edge cases are covered
- cleanup RPC methods

### DIFF
--- a/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
+++ b/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
@@ -19,7 +19,7 @@ import com.horizen.account.wallet.AccountWallet
 import com.horizen.api.http.SidechainTransactionActor.ReceivableMessages.BroadcastTransaction
 import com.horizen.chain.SidechainBlockInfo
 import com.horizen.evm.interop.TraceParams
-import com.horizen.evm.utils.Address
+import com.horizen.evm.utils.{Address, Hash}
 import com.horizen.params.NetworkParams
 import com.horizen.transaction.exception.TransactionSemanticValidityException
 import com.horizen.utils.{ClosableResourceHandler, TimeToEpochUtils}
@@ -27,8 +27,8 @@ import org.web3j.crypto.Sign.SignatureData
 import org.web3j.crypto.{SignedRawTransaction, TransactionEncoder}
 import org.web3j.utils.Numeric
 import scorex.util.{ModifierId, ScorexLogging}
-import sparkz.core.NodeViewHolder
 import sparkz.core.NodeViewHolder.CurrentView
+import sparkz.core.{NodeViewHolder, bytesToId}
 
 import java.math.BigInteger
 import java.util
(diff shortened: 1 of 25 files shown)


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -403,36 +376,18 @@ class EthService(
...
-  ) = {
-    implicit val timeout: Timeout = 5 seconds
-    val barrier = Await
-      .result(sidechainTransactionActorRef ? BroadcastTransaction(transaction), timeout.duration)
-      .asInstanceOf[Future[Unit]]
-    onComplete(barrier) {
-      // TODO: add correct responses
-      case Success(_) => ApiResponseUtil.toResponse(transactionResponseRepresentation(transaction))
-      case Failure(exp) =>
-        ApiResponseUtil.toResponse(GenericTransactionError("GenericTransactionError", JOptional.of(exp)))
```
i-Alex, line 422: It seems that we don't cover this case anymore

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -49,7 +49,8 @@ object AccountFeePaymentsUtils {
         // sum all rewards for this forger address
         val forgerTotalFee = allForgersRewards
           .filter(info => forgerKey.equals(info.address))
-          .foldLeft(BigInteger.ZERO)((sum, info) => sum.add(info.value))
+          .map(_.value)
+          .reduce(_.add(_))
```
i-Alex, line 53: does it work faster for us?
honestly, for me foldLeft approach was more clear.

sdk/src/test/scala/com/horizen/account/state/AccountStateTest.scala
```diff
@@ -87,7 +87,7 @@ class AccountStateTest
     feePayments = state.getFeePayments(0)
     assertEquals(s"Fee payments size expected to be different.", 3, feePayments.size)
 
-    var forgerTotalFee = feePayments.foldLeft(BigInteger.ZERO)((sum, payment) => sum.add(payment.value))
+    var forgerTotalFee = feePayments.map(_.value).reduce(_.add(_))
```
i-Alex, line 90: same here

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -109,7 +109,7 @@ class ForgerStakeMsgProcessorTest
     if (!view.accountExists(origin)) {
       view.addAccount(origin, randomHash)
 
-      if (amount.compareTo(BigInteger.ZERO) >= 0) {
+      if (amount.signum() == 1) {
```
i-Alex, line 112: good, no sense to increase the balance in case of `0`

sdk/src/test/scala/com/horizen/account/utils/FeeUtilsTest.scala
```diff
@@ -1,24 +1,49 @@
...
+      currentBaseFee: Long,
+      expectedBaseFee: Long
+  ): Unit = {
+    val block =
+      AccountMockDataHelper(false).getMockedBlock(BigInteger.valueOf(currentBaseFee), gasUsedPercent, 100, null, null)
+    assertEquals(message, calculateNextBaseFee(block), BigInteger.valueOf(expectedBaseFee))
+  }
+
+  /**
+   * Separate test for calculate next base fee as it is not used in base fee validator test
```
i-Alex, line 24: looks much better

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
JcBernack: I would have preferred `map(_.value).sum`, but that doesn't work with the Java BigInteger, unfortunately. It would work with the Scala BigInt but we're not using that anywhere yet. Using `reduce` is still the better option I think, because it allows the List implementation to sum the elements in any order, potentially using a faster algorithm.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
JcBernack: - If I'm not mistaken `ApiResponseUtil.toResponse()` is the wrong tool to use within the RPC api as it would write to the response directly, bypassing our RPC response serialization.
- Also I expected `Await.result` to throw in case the future does not resolve - and exceptions are caught further up and returned as an RPC error response. Did not test this though.

sdk/src/main/scala/com/horizen/account/utils/AccountFeePaymentsUtils.scala
JcBernack: done

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
JcBernack: Fixed and added a simple python test that checks that an invalid TX is rejected with an error, instead of getting a success response with the tx hash.

i-Alex (approved):

PR #564, opened 2022-10-10
Fix/fake smart contract function modifiers

GitMark0:
Add view modifiers to smart contract methods that only read state.
This needed to be added to avoid waiting for unnecessary tx receipt during testing (hardhat, truffle...).


i-Alex (approved):

JcBernack (approved):
good :+1:

PR #569, opened 2022-10-13
Reject non-zero values when calling fake contract view methods

GitMark0:
Reject calling view methods of fake smart contracts with msg.value greater than 0

### DIFF
--- a/sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
+++ b/sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
@@ -701,13 +701,13 @@ class ForgerStakeMsgProcessorTest
       val data: Array[Byte] = removeCmdInput.encode()
       var msg = getMessage(contractAddress, BigInteger.ONE, BytesUtils.fromHexString(RemoveStakeCmd) ++ data, nonce)
 
-      assertThrows[ExecutionFailedException] {
+      assertThrows[ExecutionRevertedException] {
         withGas(forgerStakeMessageProcessor.process(msg, view, _, defaultBlockContext))
       }
 
       msg = getMessage(contractAddress, BigInteger.valueOf(-1), BytesUtils.fromHexString(RemoveStakeCmd) ++ data, nonce)
 
-      assertThrows[ExecutionFailedException] {
+      assertThrows[ExecutionRevertedException] {
         withGas(forgerStakeMessageProcessor.process(msg, view, _, defaultBlockContext))
       }
 
@@ -734,13 +734,13 @@ class ForgerStakeMsgProcessorTest
       withGas { gas =>
         var msg = getMessage(contractAddress, BigInteger.ONE, BytesUtils.fromHexString(GetListOfForgersCmd), randomNonce)
 
-        assertThrows[ExecutionFailedException] {
+        assertThrows[ExecutionRevertedException] {
           forgerStakeMessageProcessor.process(msg, view, gas, defaultBlockContext)
         }
 
         msg = getMessage(contractAddress, BigInteger.valueOf(-1), BytesUtils.fromHexString(GetListOfForgersCmd), randomNonce)
 
-        assertThrows[ExecutionFailedException] {
+        assertThrows[ExecutionRevertedException] {
           forgerStakeMessageProcessor.process(msg, view, gas, defaultBlockContext)
         }
       }
--- a/sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
+++ b/sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
@@ -657,7 +657,7 @@ class ForgerStakeMsgProcessorTest
   }
 
   @Test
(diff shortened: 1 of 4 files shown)


sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -656,8 +656,173 @@ class ForgerStakeMsgProcessorTest
...
+      }
+
+      assert(returnData.getMessage === "Call value must be zero")
+
+      view.commit(bytesToVersion(getVersion.data()))
+    }
+  }
+
+  @Test
+  def testRejectSendingNegativeValueToWithdraw(): Unit = {
```
JcBernack, line 716: Don't duplicate 50 lines of code just to change the sign of value.

JcBernack: I've seen this in other places too, but why do we even need for test negative values? It is not possible to create a transaction with negative value, isn't it?

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -656,8 +656,173 @@ class ForgerStakeMsgProcessorTest
...
+        val returnData = intercept[ExecutionRevertedException] {
+          forgerStakeMessageProcessor.process(msg, view, gas, defaultBlockContext)
+        }
+        assert(returnData.getMessage === "Call value must be zero")
+      }
+    }
+  }
+
+  @Test
+  def testRejectSendingNegativeValueToGetAllForgerStakes(): Unit = {
```
JcBernack, line 798: Same here, why duplicate the whole test? Just execute two messages, one with positive, one with negative sign.

JcBernack (changes requested):

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
GitMark0: This line was calling view method with negative value before https://github.com/HorizenOfficial/Sidechains-SDK/blob/dev_evm/sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala#L296 so I wasn't sure if negatives were allowed. Before the change, this test was passing.

sdk/src/test/scala/com/horizen/account/state/WithdrawalMsgProcessorTest.scala
```diff
@@ -167,4 +171,38 @@ class WithdrawalMsgProcessorTest extends JUnitSuite with MockitoSugar with Withd
...
+    val msg = getMessage(
+      WithdrawalMsgProcessor.contractAddress,
+      BigInteger.ONE,
+      BytesUtils.fromHexString(WithdrawalMsgProcessor.GetListOfWithdrawalReqsCmdSig)
+    )
+
+    val returnData = intercept[ExecutionRevertedException] {
+      withGas(WithdrawalMsgProcessor.process(msg, mockStateView, _, defaultBlockContext))
+    }
+    assert(returnData.getMessage === "Call value must be zero")
```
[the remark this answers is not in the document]
GitMark0: Sure, but what if the method has multiple parts where the same exception is thrown? How do we know then that the exception was thrown at the exact part we are trying to test?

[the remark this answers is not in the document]
JcBernack: You don't and you shouldn't. You setup a test case with an invalid message and expect execution to revert and that's it. Anything more and you're going from testing the expected behavior to testing how it was implemented.

If this case was important enough it would have its own exception type which you could test for.

sdk/src/test/scala/com/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -656,8 +656,100 @@ class ForgerStakeMsgProcessorTest
...
+      val msgToSign = ForgerStakeMsgProcessor.getMessageToSign(forgingStakeInfo.stakeId, origin, nonce.toByteArray)
+      val msgSignatureData = Sign.signMessage(msgToSign, pair, true)
+      val msgSignature = new SignatureSecp256k1(msgSignatureData)
+
+      // create command arguments
+      val removeCmdInput = RemoveStakeCmdInput(forgingStakeInfo.stakeId, msgSignature)
+      val data: Array[Byte] = removeCmdInput.encode()
+      var msg = getMessage(contractAddress, BigInteger.ONE, BytesUtils.fromHexString(RemoveStakeCmd) ++ data, nonce)
+
+      assertThrows[ExecutionFailedException] {
```
i-Alex, line 704: I think we can be even more precise here with exception type: ExecutionFailedException -> ExecutionRevertedException
Same in all cases below

i-Alex (changes requested):
Looks good.
Let's do the minor improvements in the tests.

i-Alex (approved):

JcBernack (approved):

PR #573, opened 2022-10-18
Add more RPC functions

JcBernack:
Add RPC functions:
- eth_accounts
- eth_sendTransaction
- eth_sign (untested)

### DIFF
--- a/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
+++ b/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
@@ -50,7 +50,7 @@ class EthService(
       with ScorexLogging {
   type NV = CurrentView[AccountHistory, AccountState, AccountWallet, AccountMemoryPool]
 
-  def applyOnAccountView[R](functionToBeApplied: NV => R): R = {
+  private def applyOnAccountView[R](functionToBeApplied: NV => R): R = {
     implicit val timeout: Timeout = new Timeout(nvtimeout)
     val res = scNodeViewHolderRef
       .ask {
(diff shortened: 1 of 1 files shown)


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -192,6 +183,23 @@ class EthService(
...
   }
 
+  /**
+   * Sign calculates an ECDSA signature for: keccack256("\x19Ethereum Signed Message:\n" + len(message) + message). This
+   * gives context to the signed message and prevents signing of transactions.
+   */
+  @RpcMethod("eth_sign")
+  def sign(sender: Address, message: String): String = {
+    val data = Numeric.hexStringToByteArray(message)
+    val prefix = s"\u0019Ethereum Signed Message:\n${data.length}"
```
i-Alex, line 193: I think, that every network must define its unique prefix to avoid using message signed in network A to be used in network B.
Can you double check this with other networks?

JcBernack: Good point, but GETH has this hardcoded. It seems that replaying signed messages is a common attack vector on smart contracts that use them, not just across different networks. So, smart contract developers have to be aware of this and prevent it anyway: https://medium.com/cryptronics/signature-replay-vulnerabilities-in-smart-contracts-3b6f7596df57

i-Alex: That's why i assume we need to hard-code our specific value

JcBernack: I wonder if that would break some Ethereum tooling, because they all use this exact value.

i-Alex: ok, seems other network use the same message prefix.

i-Alex (approved):

PR #575, opened 2022-10-25
SDK-449 - Secure Enclave integration into CertificateSubmitter

i-skrypnyk:


paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/api/http/client/SecureEnclaveApiClient.scala
```diff
@@ -0,0 +1,88 @@
...
+import com.horizen.proof.SchnorrSignatureSerializer
+import com.horizen.proposition.{SchnorrProposition, SchnorrPropositionSerializer}
+import com.horizen.utils.BytesUtils
+import io.circe.generic.auto._
+import io.circe.syntax._
+import scorex.util.ScorexLogging
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class SecureEnclaveApiClient(settings: RemoteKeysManagerSettings)(implicit ec: ExecutionContext) extends Actor with ScorexLogging {
```
i-Alex, line 22: Is there any benefit of having SecureEnclaveApiClient as an Actor.
You may look at `mainchainChannel: MainchainNodeChannel` and it's usage/implementations.

I think that it is more than enough to define `trait SecureEnclaveApi`
with a couple of methods to know if it is enabled and to generate signatures.

It will reduce code complexity and will make it more stable.
Let's discuss it.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/api/http/client/SecureEnclaveApiClient.scala
```diff
@@ -0,0 +1,88 @@
...
+object SecureEnclaveApiClient {
+  case class SignWithEnclave(message: Array[Byte],
+                             publicKey_index: (SchnorrProposition, Int))
+
+  case class CreateSignatureRequest(message: Array[Byte],
+                                    publicKey: String,
+                                    `type`: String)
+
+  case class CreateSignatureResponse(signature: Option[String],
+                                     error: Option[String])
```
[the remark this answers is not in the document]
i-skrypnyk: This one can't be moved to private part, as it is used in unmarshalling. When moving as a nested case class, unmarshalling fails cause it's unable to use apply-style constructor

qa/SidechainTestFramework/secure_enclave_http_api_server.py
```diff
@@ -0,0 +1,52 @@
...
+import subprocess
+import threading
+
+from flask import Flask, request, json
+
+
+class SecureEnclaveApiServer(object):
+    app = Flask(__name__)
+
+    @app.route('/signMessage', methods=['POST'])
```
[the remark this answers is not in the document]
i-skrypnyk: I was not able to find post body scheme settings in Flask

qa/sc_cert_submitter_secure_enclave.py
```diff
@@ -93,20 +85,20 @@ def run_test(self):
...
             logging.info("Wait for certificates in the MC mempool...")
-            if sc_node2.submitter_isCertGenerationActive()["result"]["state"]:
-                logging.info("sc_node2 generating certificate now.")
+            if sc_node1.submitter_isCertGenerationActive()["result"]["state"]:
+                logging.info("sc_node1 generating certificate now.")
             time.sleep(2)
         assert_equal(1, mc_node.getmempoolinfo()["size"], "Certificates was not added to MC node mempool.")
-        logging.info("Node after synchronization was able to sign, collect signatures and emit certificate.")
-        api_server.stop()
+        logging.info("Node with Secure Enclave was able to sign, collect signatures and emit certificate.")
```
i-Alex, line 100: let's check that cert in mempool has quality = 6, to be sure that both 3 keys from wallet and 3 keys from enclave were used.

qa/sc_cert_submitter_secure_enclave.py
```diff
@@ -93,20 +85,20 @@ def run_test(self):
...
-        generate_next_block(sc_node1, "second node")  # 1 MC block to reach the end of WE
-        generate_next_block(sc_node1, "second node")  # 1 MC block to trigger Submitter logic
+        generate_next_block(sc_node1, "first node")  # 1 MC block to reach the end of WE
+        generate_next_block(sc_node1, "first node")  # 1 MC block to trigger Submitter logic
         self.sc_sync_all()  # Sync SC nodes
 
         # Wait for Certificates appearance
-        time.sleep(10)
-        while mc_node.getmempoolinfo()["size"] < 1 and sc_node2.submitter_isCertGenerationActive()["result"]["state"]:
+        time.sleep(100)
```
i-Alex, line 93: 100 -> 10 is fine.
no need to wait so long

sdk/src/main/scala/com/horizen/api/http/client/SecureEnclaveApiClient.scala
```diff
@@ -19,70 +18,97 @@ import scorex.util.ScorexLogging
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class SecureEnclaveApiClient(settings: RemoteKeysManagerSettings)(implicit ec: ExecutionContext) extends Actor with ScorexLogging {
+class SecureEnclaveApiClient(settings: RemoteKeysManagerSettings)(implicit system: ActorSystem, ec: ExecutionContext) extends ScorexLogging {
```
i-Alex, line 21: `(implicit system: ActorSystem, ec: ExecutionContext)` can be removed

sdk/src/main/scala/com/horizen/api/http/client/SecureEnclaveApiClient.scala
```diff
@@ -19,70 +18,97 @@ import scorex.util.ScorexLogging
...
   private val signatureSerializer: SchnorrSignatureSerializer = SchnorrSignatureSerializer.getSerializer
+  val SCHNORR = "schnorr"
 
-  private def serializePublicKey(publicKey: SchnorrProposition) =
-    BytesUtils.toHexString(keySerializer.toBytes(publicKey))
+  def isEnabled: Boolean = settings.enabled
+
+  def listPublicKeys(): Future[Seq[SchnorrProposition]] = {
+    http.singleRequest(
+      Post(settings.address + "/api/v1/listKeys")
```
i-Alex, line 32: "/api/v1/" is it a requirement from Infra?
if not, you should expect to have "/api/v1/" as a part of `settings.address`

sdk/src/main/scala/com/horizen/certificatesubmitter/CertificateSubmitter.scala
```diff
@@ -324,20 +324,24 @@ class CertificateSubmitter(settings: SidechainSettings,
...
-    ).asInstanceOf[Seq[Option[CertificateSignatureInfo]]]
+    if (!secureEnclaveApiClient.isEnabled) return Seq()
+
+    val signaturesFromEnclaveFuture = secureEnclaveApiClient.listPublicKeys()
+      .map(managedKeys => indexedPublicKeys.filter(key_index => managedKeys.contains(key_index._1)))
+      .map(_.map(secureEnclaveApiClient.signWithEnclave(messageToSign, _)))
+      .map(Future.sequence(_))
       .flatten
+
+    Await.result(signaturesFromEnclaveFuture, timeoutDuration).flatten
```
i-Alex, line 344: does it cover Timeout exception case?
So in case we have timeout, we return empty sequence rather than passing exception further.

sdk/src/test/scala/com/horizen/api/http/client/SecureEnclaveApiClientTest.scala
```diff
@@ -32,27 +30,99 @@ class SecureEnclaveApiClientTest extends AnyWordSpec with Matchers with MockitoS
...
   "Secure Enclave Api should " should {
 
-    "return empty future if it is disabled" in {
+    "tell if it is enabled" in {
       val (apiClient, serverMock) = prepareApiClient(false)
-      val key = generateKey().publicImage()
-      val result = Await.result(
-        apiClient ? SignWithEnclave("test".getBytes, (key, 1)), 1.second
-      ).asInstanceOf[Option[CertificateSignatureInfo]]
+      assert(false.equals(apiClient.isEnabled))
```
i-Alex, line 41: `assertFalse` can be used.

sdk/src/test/scala/com/horizen/api/http/client/SecureEnclaveApiClientTest.scala
```diff
@@ -32,27 +30,99 @@ class SecureEnclaveApiClientTest extends AnyWordSpec with Matchers with MockitoS
...
+      verifyNoInteractions(serverMock)
+
+      val (apiClientEnabled, mockEnabled) = prepareApiClient()
+      assert(true.equals(apiClientEnabled.isEnabled))
```
i-Alex, line 45: assertTrue can be used

sdk/src/test/scala/com/horizen/api/http/client/SecureEnclaveApiClientTest.scala
```diff
@@ -32,27 +30,99 @@ class SecureEnclaveApiClientTest extends AnyWordSpec with Matchers with MockitoS
...
+          .toString
+
+      when(serverMock.singleRequest(any(),any(),any(),any()))
+        .thenReturn(Future.successful(
+          HttpResponse(status = StatusCodes.OK, entity = response)
+        ))
+
+      val result = Await.result(apiClient.listPublicKeys(), 1.second)
+
+      assert(result.size.equals(3))
```
i-Alex, line 95: use `assertEquals` and add message to explain the failure.
Same in other places

sdk/src/test/scala/com/horizen/certificatesubmitter/CertificateSubmitterTest.scala
```diff
@@ -386,11 +387,8 @@ class CertificateSubmitterTest extends JUnitSuite with MockitoSugar {
     })
     val mockedSidechainNodeViewHolderRef: ActorRef = mockedSidechainNodeViewHolder.ref
 
-    val secureEnclaveApiClient = TestProbe()
-    secureEnclaveApiClient.setAutoPilot((sender: ActorRef, _: Any) => { sender ! None; TestActor.KeepRunning })
-
     val certificateSubmitterRef: TestActorRef[CertificateSubmitter] = TestActorRef(
-      Props(new CertificateSubmitter(mockedSettings, mockedSidechainNodeViewHolderRef, secureEnclaveApiClient.ref, params, mockedMainchainChannel)))
+      Props(new CertificateSubmitter(mockedSettings, mockedSidechainNodeViewHolderRef, mock[SecureEnclaveApiClient], params, mockedMainchainChannel)))
```
i-Alex, line 391: Extend `def newBlockArrived(): Unit = {` test, case "test 2" emulate signatures from both wallet and enclave.

qa/resources/template.conf
```diff
@@ -77,4 +77,9 @@ sparkz {
     restrictForgers = %(RESTRICT_FORGERS)s
     allowedForgersList = %(ALLOWED_FORGERS_LIST)s
   }
+
+  remoteKeyManager {
+    enabled = true
```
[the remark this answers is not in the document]
i-Alex: still actual

qa/resources/template_predefined_genesis.conf
```diff
@@ -60,4 +60,9 @@ sparkz {
     restrictForgers = %(RESTRICT_FORGERS)s
     allowedForgersList = %(ALLOWED_FORGERS_LIST)s
   }
+
+  remoteKeysManager {
+    enabled = true
```
i-Alex, line 65: this one should be configurable as well

i-Alex (changes requested):
Looks much better now.
Let's address all previous comments and extend tests a bit.

qa/SidechainTestFramework/secure_enclave_http_api_server.py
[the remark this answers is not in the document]
i-Alex: here I meant a bit more simpler thing.
Just put a JSON body scheme: what are the arguments names/types

qa/resources/template.conf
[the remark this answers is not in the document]
i-skrypnyk: fixed, thanks

sdk/src/main/scala/com/horizen/api/http/client/SecureEnclaveApiClient.scala
i-skrypnyk: yes, that's a requirement from here https://horizenlabs.atlassian.net/browse/INFRA-711

sdk/src/main/scala/com/horizen/certificatesubmitter/CertificateSubmitter.scala
i-skrypnyk: In case of a timeout, we'll hit this part inside API client
```
        case response: HttpResponse =>
          logger.error(s"Secure Enclave returned non-200 response: ${response.status}, ${response.entity}")
          Future(Seq())
```
and immediately return empty seq here

akka timeout can be configured using `akka.http.server.request-timeout` config setting, and is 20 sec by default

One thing I missed is that one can set restApi timeout to be something smaller, so I'll cover it to work as you said

sdk/src/test/scala/com/horizen/api/http/client/SecureEnclaveApiClientTest.scala
i-skrypnyk: changed to scala style matchers, they provide a human readable output describing what went wrong

sdk/src/test/scala/com/horizen/certificatesubmitter/CertificateSubmitterTest.scala
i-skrypnyk: I was not able to modify it to have 2 sigs in Test 2, as it keep on failing further. So I replaced one from wallet with one from enclave. Wallet sigs are tested further in a test several times, so they are covered.

qa/SidechainTestFramework/secure_enclave_http_api_server.py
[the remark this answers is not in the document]
i-skrypnyk: Still not sure that I understood you correctly, but I've added comments with json scheme descriptions above the methods

cronicc (approved):

i-Alex (approved):

paolocappelletti (approved):

PR #576, opened 2022-10-25
EVM-394 -  config parameter to manage forging rate

davehorizen:
In this pull request the ability to change the block rate has been added to the SDK.
The block rate is passed using the field _forger.forgingFrequencyInSeconds_ of the config file, if not present the block rate will be 120 seconds.


i-Alex:
Substituted by https://github.com/HorizenOfficial/Sidechains-SDK/pull/584

PR #578, opened 2022-10-27
EVM-405 - rpc id numeric datatype

davehorizen:
In this pull request the datatype the id rpc request and response was modified to long (instead of String) to be compliant with the web3.js library

### DIFF
--- a/sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
+++ b/sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
@@ -3,11 +3,11 @@ package com.horizen.account.api.rpc.request;
 import com.fasterxml.jackson.databind.JsonNode;
 
 /**
- * {"id":"1648039192785","jsonrpc":"2.0","method":"eth_chainId","params":[]}
+ * {"id":1648039192785,"jsonrpc":"2.0","method":"eth_chainId","params":[]}
  */
 public class RpcRequest {
     private String jsonrpc;
-    private String id;
+    private long id;
     private String method;
     private JsonNode params;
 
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
```diff
@@ -3,19 +3,19 @@
...
+    private long id;
     private String method;
     private JsonNode params;
 
     public RpcRequest() {}
 
     public RpcRequest(JsonNode json) {
         this.jsonrpc = json.get("jsonrpc").asText();
-        this.id = json.get("id").asText();
+        this.id = json.get("id").asLong();
```
i-Alex, line 18: According to the specs: https://www.jsonrpc.org/specification#examples
id can be of string, decimal or null.
Here we just replaces String with `long`. But we should support both.
we may keep `id` of type Object but assign Long or String value.
Then in the Response serialize depends on this type.
Need to add UTs

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
davehorizen: Requested changes pushed.
Now the rpc ID response datatype will be the same of the input.
As found in other ethereum like chains (like BSC) we will not supported numeric negative ID and null ID.

request/response examples:
_numeric id_
`curl http://localhost:9085/ethv1 -X POST --data '{"method":"eth_chainId","params":[],"id":25:"jsonrpc":"2.0"}'
{
   "jsonrpc":"2.0",
   "result":"0x7cd",
   "id":25
}`
_string id_
`curl http://localhost:9085/ethv1 -X POST --data '{"method":"eth_chainId","params":[],"id":"32759":"jsonrpc":"2.0"}'
{
   "jsonrpc":"2.0",
   "result":"0x7cd",
   "id":"32759"
}`

i-Alex (approved):

PR #585, opened 2022-11-03
Feature/evm 143

GitMark0:
Added `eth_getFeePayments` rpc endpoint


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -578,6 +579,25 @@ class EthService(
...
+    applyOnAccountView { nodeView =>
+      val feePayments =
+        nodeView.history
+          .getFeePaymentsInfo(blockId)
+
+      if (feePayments.isEmpty) {
+        throw new RpcException(RpcError.fromCode(RpcCode.ExecutionError, s"no fee payments found for block: $blockId"))
+      }
+
+      feePayments.get()
```
JcBernack, line 597: Should it really be an exception if there are no fee payments? What about empty blocks? @alsala @i-Alex 

Anyways, this could be simplified to:
```scala
  nodeView.history
    .getFeePaymentsInfo(blockId)
    .getOrElse(throw new RpcException(
      RpcError.fromCode(RpcCode.ExecutionError, s"no fee payments found for block: $blockId")
    ))
}
```

JcBernack (changes requested):
I still think we should not use `eth_` for our own custom methods. Why don't we just use `zen_getFeePayments`? This would clearly mark these as Zen specific, non-ethereum-standard RPC methods. Same goes for the already implemented `eth_getForwardTransfers`. We need to decide that now, otherwise we can't change it anymore without breaking things for others.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
GitMark0: I wasn't able to use `getOrElse` method on Java's `Optional` class, so instead I used `ifPresent`. Now the method returns null if the payments weren't found in the specified block because that makes more sense to me also.

i-Alex (approved):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: current implementation is fine and coincide with another similar call: `eth_getFeePayments`

JcBernack (approved):

JcBernack:
What about my comments regarding the RPC methods name? https://github.com/HorizenOfficial/Sidechains-SDK/pull/585#pullrequestreview-1166940027

i-Alex:
@JcBernack talked to victor, we will change `eth_` to `zen_` in the recent days.
@GitMark0 pls, merge dev_evm and resolve conflicts

PR #587, opened 2022-11-04
evm-244: as/fix sum delegated forger stakes

alsala:


sdk/src/main/scala/com/horizen/api/http/TransactionBaseApiRoute.scala
```diff
@@ -0,0 +1,174 @@
...
+        case Success(tx) =>
+          //TO-DO JSON representation of transaction
+          ApiResponseUtil.toResponse(RespDecodeTransactionBytes(tx))
+        case Failure(exp) =>
+          ApiResponseUtil.toResponse(ErrorByteTransactionParsing(exp.getMessage, JOptional.of(exp)))
+      }
+    }
+  }
+
+  def allActiveForgingStakeInfo: Route = (post & path("allActiveForgingStakeInfo")) {
```
i-Alex, line 75: I'm not sure that the stake info we return here is "active".
Actually we return the forging stake the current state. But it is not active, only the snapshot of all stakes in the end of the consensus epoch `N` will be active at epoch `N+2`.

These 2 methods can be renamed without word "Active".
Moreover, do we wan't them to be a part of `transaction` API route? We may move it to `block` as other forger related APIs. Or even to `debug` methods. What do you think?

i-Alex (changes requested):
Fix looks good.
A minor point about API

sdk/src/main/scala/com/horizen/api/http/TransactionBaseApiRoute.scala
alsala: After discussion has been decided to remove these methods. Checks can be made by inspecting the forged block header field `forgingStakeInfo` for expected `stakeAmount` value. The python test has been modified accordingly.

i-Alex (approved):

PR #588, opened 2022-11-08
libevm improvements

JcBernack:
- [EVM-420](https://horizenlabs.atlassian.net/browse/EVM-420) logger optimization
- [EVM-427](https://horizenlabs.atlassian.net/browse/EVM-427) general improvements
- [EVM-428](https://horizenlabs.atlassian.net/browse/EVM-428) database caching


evm/src/main/java/com/horizen/evm/LibEvmLogCallback.java
```diff
@@ -14,10 +14,17 @@ private LibEvmLogCallback() {
         // this is a singleton, prevent more instances
     }
 
+    static void Register() {
+        // register log callback
+        LibEvm.SetLogCallback(LibEvmLogCallback.instance);
+        // propagate log4j log level to glog
+        LibEvm.SetLogLevel(log4jToGlogLevel(logger.getLevel()));
```
i-Alex, line 21: Are we sure that at the moment of `Register` method invocation we have a proper `logger.getLevel()` defined?
If i'm not wrong, for SDK we setup logger arguments: what to print, what to put to the log file in a runtime after parsing the config file.
PS: try to print the log level here, and you will see that it is always `ALL`

I think we need to call this method in `LogInitializer`
Moreover, at the moment this code breaks the logging of SDK: hardcode it to `ALL` always.

i-Alex (changes requested):

evm/src/main/java/com/horizen/evm/LibEvmLogCallback.java
JcBernack: Hm, I tested it with "ALL" and "DEBUG" in the root log level of the log4j config file `<Root level="all">` and it was working as expected.

> Moreover, at the moment this code breaks the logging of SDK: hardcode it to ALL always.

What do you mean? What breaks?

i-Alex (approved):

PR #592, opened 2022-11-11
Ak/update json rpc api doc

akrckn:


akrckn:
Updated horizen specific methods to `zen_` instead of `eth_`, because I saw, that is was changed in the meantime

JcBernack (approved):

i-Alex (approved):

PR #593, opened 2022-11-11
St/evm 506

saratnt:


sdk/src/test/scala/com/horizen/account/performance/AccountForgeMessageBuilderPerfTest.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  val nodeView: CurrentView[AccountHistory, AccountState, AccountWallet, AccountMemoryPool] =
+    mock[CurrentView[AccountHistory, AccountState, AccountWallet, AccountMemoryPool]]
+  Mockito.when(nodeView.pool).thenReturn(mempool)
+
+  val blockContext = new BlockContext(Array.empty[Byte], 1000, BigInteger.ZERO, FeeUtils.GAS_LIMIT, 11, 2, 3)
+
+  @Test
+  @Ignore
+  def testComputeBlockInfo(): Unit = {
```
i-Alex, line 67: Let's add a comment that explain the purpose of the test

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
```diff
@@ -363,4 +341,51 @@ class MempoolMap(stateReaderProvider: AccountStateReaderProvider) extends Scorex
     }
   }
 
+  class TransactionsByPriceAndNonce(baseFee: BigInteger) extends Iterable[SidechainTypes#SCAT] {
```
i-Alex, line 344: align the class code

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
```diff
@@ -300,53 +275,56 @@ class MempoolMap(stateReaderProvider: AccountStateReaderProvider) extends Scorex
     }
 
     val execTxsOpt = executableTxs.remove(account)
-    if (execTxsOpt.nonEmpty){
-      execTxsOpt.get.dropWhile {
-        case (nonce, id) => {
-          if (nonce.compareTo(newExpectedNonce) < 0) {
+    if (execTxsOpt.nonEmpty) {
+      execTxsOpt.get
+        .withFilter {
```
i-Alex, line 280: can you explain more, why we use `.withFilter` instead of `.dropWhile` now?
why is it faster?

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -121,37 +118,37 @@ class AccountForgeMessageBuilder(
...
             cumForgerTips = cumForgerTips.add(txForgerTipPerGas.multiply(txGasUsed))
-
+            priceAndNonceIter.next()
           case Failure(e: GasLimitReached) =>
             // block gas limit reached
             // keep trying to fit transactions into the block: this TX did not fit, but another one might
             log.trace(s"Could not apply tx, reason: ${e.getMessage}")
             // skip all txs from the same account
-            accountsToSkip += tx.getFrom
+            priceAndNonceIter.pop()
```
i-Alex, line 127: seems a bit tricky that `pop` excludes all the entries for given account.
Maybe we may choose another name

i-Alex (changes requested):
Nice!
Similar notes about the comments as for the base branch PR.

sdk/src/test/scala/com/horizen/account/performance/AccountForgeMessageBuilderPerfTest.scala
saratnt: Done

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
saratnt: I used `dropWhile` in the beginning because I wanted to stop checking the nonce after having found the first element with nonce >= newExpectedNonce. In fact, being the collections of txs ordered by nonce, all the subsequent txs must have nonce > newExpectedNonce. The problem is that the `dropWhile` builds a new collection and this is really very expensive. `withFilter` instead doesn't create a new collection, but it is like adding a condition that will be evaluated in the next `foreach`.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
saratnt: OK, I'll change the name of pop method to removeAndSkipAccount.

i-Alex (approved):

PR #596, opened 2022-11-15
Q8 smart contract tests

akrckn:
Moved duplicate function defitions to one `ac_utils` file, moved eoa / address and evm utils functions to `ac_utils`, fix comments, fix init values.


akrckn:
Please have a look here or in #603 - I created #603, because I need the changes here for all other tickets I currently work on.

PR #608, opened 2022-11-17
Python test improvements Q5 / Q6

akrckn:
Renaming of files will be done after review, because it would make the review unnecessarily difficult now.


qa/SidechainTestFramework/account/httpCalls/wallet/balance.py
```diff
@@ -1,7 +1,7 @@
 import json
 
 #executes a wallet/balance call on an account sidechain
-def http_wallet_balance(sidechainNode, evm_address, api_key = None):
+def http_wallet_balance(sidechainNode, evm_address, api_key = 'Horizen'):
```
i-Alex, line 4: I'm not sure that default is `Horizen`, should be `None` as for all other methods.
Pls, check the usage.

i-Alex (changes requested):
Looks very good!

i-Alex (approved):

PR #609, opened 2022-11-17
Fix/evm 451

GitMark0:


sdk/src/main/java/com/horizen/account/api/rpc/service/RpcService.java
```diff
@@ -78,6 +78,10 @@ public Object execute(RpcRequest req) throws Throwable {
             return method.invoke(this, args);
         } catch (InvocationTargetException e) {
             // unpack and rethrow potential RpcException
+            LogManager.getLogger().warn("RPC call failed: " + method, e);
```
i-Alex, line 81: let's make the log level `trace` for all these 3 logging invocations

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -0,0 +1,69 @@
...
+        this.timestamp = Numeric.prependHexPrefix(Long.toHexString(block.timestamp()));
+        this.baseFeePerGas = Numeric.toHexStringWithPrefix(blockHeader.baseFee());
+
+        var transactions = scala.collection.JavaConverters.seqAsJavaList(block.transactions());
+
+        if (!hydratedTx) {
+            this.transactions = transactions.stream().map(t -> Numeric.prependHexPrefix((String) t.id())).collect(Collectors.toList());
+        }
+        else {
+            this.transactions = transactions;
```
i-Alex, line 65: before it was:
```
stateView
                    .getTransactionReceipt(Numeric.hexStringToByteArray(tx.id))
                    .map(new EthereumTransactionView(_, tx, block.header.baseFee))
```
So I guess, we need to work with `EthereumTransactionView` here

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -0,0 +1,69 @@
...
+    public final String mixHash;
+    public final String extraData;
+    public final String size;
+    public final String gasLimit;
+    public final String gasUsed;
+    public final String timestamp;
+    public final List<?> transactions;
+    public String author;
+    public String difficulty;
+    public String totalDifficulty;
```
i-Alex, line 33: still not used. check the comments in the task

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -0,0 +1,69 @@
...
+    public List<String> uncles;
+    public List<String> sealFields;
```
i-Alex, line 35: still null, need to assign empty list value

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
```diff
@@ -11,15 +18,22 @@ public class RpcRequest {
     private String method;
     private JsonNode params;
 
-    public RpcRequest() {}
+    public RpcRequest(JsonNode json) throws RpcException {
+        List<String> keys = new ArrayList<>();
+        var iterator = json.fieldNames();
+        iterator.forEachRemaining(e -> keys.add(e));
+
+        if (!keys.containsAll(List.of("jsonrpc", "id", "method", "params"))) {
```
davehorizen, line 26: Throwing an exception in this point can be very useful to check the correctness of a single rpc request or a multiple batch request, but if we want to manage rpc notifications we have to leave out the check on the _id_ field.

A notification is a request object without an _id_ field. A request object that is a notification signifies the client's lack of interest in the corresponding response object, and as such no response object needs to be returned to the client.

@i-Alex I ask for your confirmation on the need to handle this case as well

davehorizen (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcRequest.java
i-Alex: yes, we need to manage notifications case as well.
Please, add it on your PR after the rebase later.

davehorizen (approved):

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
i-Alex: @GitMark0 I see that these two fields has been removed at all now.
It is the issue.
Instead we need to:
1) keep this two fields
2) assign `difficulty = 1` always
3) assign `totalDifficulty = blockNumber`

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -64,10 +64,10 @@ public EthereumBlockView(Long blockNumber, String hash, boolean hydratedTx, Acco
         this.timestamp = Numeric.prependHexPrefix(Long.toHexString(block.timestamp()));
         this.baseFeePerGas = Numeric.toHexStringWithPrefix(blockHeader.baseFee());
 
-        var transactions = (List<EthereumTransaction>) scala.collection.JavaConverters.seqAsJavaList((block.transactions()));
+        var transactions = scala.collection.JavaConverters.seqAsJavaList(block.ethereumTransactions());
```
i-Alex, line 67: Ok, i got the issue on java side.
But it is not good to keep `ethereumTransactions` method inside the AccountBlock. No one should use it.
Let's instead move it to some scala util object in the `rpc` package.

i-Alex (changes requested):
Check the comment for `difficulty`

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
JcBernack: Shouldn't the difficulty be zero? In Ethereum PoS it is always `difficulty = 0x0` imho. I don't know what `totalDifficulty` is in Ethereum now though.

Example response by Alchemy RPC on current Ethereum Mainnet:
```
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": {
    "baseFeePerGas": "0x4696470c1",
    "difficulty": "0x0",
    "extraData": "0x496c6c756d696e61746520446d6f63726174697a6520447374726962757465",
    "gasLimit": "0x1c9c380",
    "gasUsed": "0xb68af9",
    "hash": "0x3cb935908566e5f255c08629838aa10107a1e18c9c975ad8567ebf6142389339",
    "logsBloom": "0x1c212572e1c3eb4213942454d130922af92bc800719ac20eab99c0255626981240022fb68d51400b2334b84a520f29f132731ed45fc8b84fc2c16b00d02c296961834d005c6f01296b4b2008902007a5929712a048448edd01d21de08601203a126102c1323e811a10c830853b6298cd0c6c0c63c735366c620441522dfd48621374077c906a104984df9c090938110084013945815084b9507650c439701a03bf73514f40007011800574e89c4e366c4b41862f988ea882682626c885e909156540c6f2b98d5af01c1289132807360683b73ed08aaaa050dd9c9fc6dc41f45440b5e948a40352146083318908411c8aa50818500b6807f4682f0e43d8127113",
    "miner": "0xdafea492d9c6733ae3d56b7ed1adb60692c98bc5",
    "mixHash": "0xe8a098ab9940c376e1d29c02d803b1b7e4d094728964ce2596955b4a4cf48824",
    "nonce": "0x0000000000000000",
    "number": "0xf612bf",
    "parentHash": "0xd8dc01fefab792f2816f113845fb3a45cc46157efdeb9a334ef60d0e7ec6df77",
    "receiptsRoot": "0x80b415f75be82fa9742257a9606dcee7541b123ff2d1ce2f25caf227c2655ccf",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "size": "0x15689",
    "stateRoot": "0x0bcf4463b56e8fc2a5d2ce7044b874b55a778cebed88b1b542e3c2bee587204a",
    "timestamp": "0x638f5fdf",
    "totalDifficulty": "0xc70d815d562d3cfa955",
    "transactions": [
      "0xb33042a944ebaa36a692e60220693b364e9fc9b54be5c716ff7c8a76ad4a8c2b",
      "..... removed"
    ],
    "transactionsRoot": "0x402ccf0682a11b7cb5cb5a108faeaa202fce71ca2b271dbc4606d9fb3d35a1e7",
    "uncles": []
  }
}
```

sdk/src/main/scala/com/horizen/account/block/AccountBlock.scala
```diff
@@ -35,6 +37,8 @@ class AccountBlock(override val header: AccountBlockHeader,
 
   override lazy val transactions: Seq[SidechainTypes#SCAT] = sidechainTransactions
 
+  def ethereumTransactions: Seq[EthereumTransaction] = sidechainTransactions.map(_.asInstanceOf[EthereumTransaction])
```
JcBernack, line 40: Move to a util like Alex suggested, or maybe make this non-specific to Ethereum:
```suggestion
  def transactionsAs[T]: Seq[T] = sidechainTransactions.map(_.asInstanceOf[T])
```

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
GitMark0: Just confirming that the `difficulty` param should be `0` since totalDifficulty is calculated as the total of the difficulty of all the previous blocks plus the previous blocks. We can see, from Etherscan, that the `totalDifficulty` didn't change since the merge.

GitMark0: There are some weird things though, they said that difficulty bomb hits at: `58,700,000,000,000,000,000,000`, but current `totalDifficulty` is `58,750,003,716,598,352,816,469` and that number didn't change in last 10 blocks that I've checked...

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -29,10 +40,11 @@ public class EthereumBlockView {
...
     public String author;
-    public String difficulty;
-    public String totalDifficulty;
-    public List<String> uncles;
-    public List<String> sealFields;
+    public List<String> uncles = new ArrayList<>();
+    public List<String> sealFields = new ArrayList<>();
+
+    public final String difficulty = "0x";
+    public final String totalDifficulty = "0x";
```
i-Alex, line 47: `0x` -> `0x0`

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
JcBernack: The difficulty bomb was delay multiple times, the final terminal total difficulty was set to `58_750_000_000_000_000_000_000`.

If the `totalDifficulty` should be the sum of the `difficulty` of all previous blocks then I assume, for us, a zero is correct here too. We need to output `difficulty` and `totalDifficulty` just for compatibility, they don't have a meaning for us - just like it is on Ethereum mainnet now, after the merge.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -41,8 +53,7 @@ public EthereumBlockView(Long blockNumber, String hash, boolean hydratedTx, Acco
         this.number = blockNumber;
         this.hash = hash;
         this.parentHash = Numeric.prependHexPrefix((String) block.parentId());
-        // no nonce, but we explicity set it to all zeroes as some RPC clients are very strict (e.g. GETH)
-        this.nonce = "0x0000000000000000";
+        this.nonce = "0x"; // no nonce
```
JcBernack, line 56: Why was this changed? The previous comment was there for a reason :wink:

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumBlockView.java
```diff
@@ -57,13 +68,12 @@ public EthereumBlockView(Long blockNumber, String hash, boolean hydratedTx, Acco
...
+        var transactions = scala.collection.JavaConverters.seqAsJavaList(AccountBlockUtil.ethereumTransactions(block.transactions()));
 
         if (!hydratedTx) {
-            this.transactions = transactions.stream().map(t -> Numeric.prependHexPrefix((String) t.id())).collect(Collectors.toList());
-        }
-        else {
-            this.transactions = transactions;
+            this.transactions = transactions.stream().map(t -> Numeric.prependHexPrefix(t.id())).collect(Collectors.toList());
+        } else {
+            this.transactions = transactions.stream().map(t -> new EthereumTransactionView(null, t, block.header().baseFee())).collect(Collectors.toList());
```
JcBernack, line 76: I don't think the constructor of `EthereumTransactionView` currently accepts `null`, it will throw NullPointerException.

JcBernack (changes requested):

JcBernack:
Checking again with the (long) list of things in [EVM-451](https://horizenlabs.atlassian.net/browse/EVM-451), I think currently there is two main things missing from the PR:

- RpcRequest: remove unused getters/setters and handle invalid json
- RpcService: Unit Tests (mainly for  the argument conversions I think)

JcBernack (changes requested):

JcBernack:
FYI: something weird happened in this PR during a merge of dev_evm into this branch. The GitHub UI here does not show all the changes. Reviewing should either be done outside or we have to fix and force-push this branch.

Other then that, I think the last remaining issue here is:
- RpcService: Unit Tests (mainly for the argument conversions I think)

I suggest putting that into a new ticket and proceed with merging. This PR is already getting quite old.

JcBernack:
Added a fix for the missing `0x` prefix on the "to" address here , see [EVM-596](https://horizenlabs.atlassian.net/browse/EVM-596).

JcBernack (approved):

JcBernack:
Meh, now with `dev_evm` merged in some of the changes in this branch make some tests in `EthServiceTest` fail. Need to update.

i-Alex (approved):

PR #612, opened 2022-11-21
Fix

saratnt:


i-Alex (approved):

alsala (approved):

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
```diff
@@ -131,7 +131,7 @@ abstract class SidechainBlockBase[TX <: Transaction, H <: SidechainBlockHeaderBa
       case Failure(e) => throw e
     }
 
-    if(verifyTransactionsListSize)
+    if(transactionsListExceedsSizeLimit)
```
alsala, line 134: Consider using "number of transactions %d" instead of "transactions amount"

PR #616, opened 2022-11-22
Manage RPC Batch

davehorizen:


sdk/src/main/java/com/horizen/account/api/rpc/handler/RpcHandler.java
```diff
@@ -30,4 +34,29 @@ public ApiResponse apply(RpcRequest request) {
...
+            } catch (RpcException e) {
+                responseList.add(new RpcResponseError(rpcRequest.getId(), e.error));
+            } catch (Throwable e) {
+                responseList.add(new RpcResponseError(rpcRequest.getId(), RpcError.fromCode(RpcCode.InternalError, e.getMessage())));
+            }
+        }
+
+        // add invalid requests to the batch response
+        for(int i=0; i<request.getInvalidRequestsNumber(); i++)
+            responseList.add(new RpcResponseError(new RpcId(), RpcError.fromCode(RpcCode.InvalidRequest, null)));
```
i-Alex, line 58: It is not clear to me this part.
It seems we append invalid requests messages to the end of the result array. And moreover we miss `rpcId` fields.

Instead in case of batch request we should return the results in the same order as requests and with proper id fields.

I guess, we miss some UTs for this case

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcBatchRequest.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        List<RpcRequest> inputRpcRequests = new ArrayList<>();
+        for(JsonNode jsonItem: json) {
+            try {
+                RpcRequest rpcRequest = new RpcRequest(jsonItem);
+                inputRpcRequests.add(rpcRequest);
+            } catch (RpcException e) {
+                invalidRequestsNumber++;
+            }
```
i-Alex, line 35: I think we should detect that the JSON is an array and keep entries as json object elements of that array.
It will allow us to properly process single requests and properly return errors (in case of invalid request, for example)

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcId.java
```diff
@@ -7,6 +7,8 @@ public class RpcId {
     private Long longId;
     private String stringId;
 
+    public RpcId() {}
```
i-Alex, line 10: do we really need this constructor?

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
```diff
@@ -72,10 +74,29 @@ case class AccountEthRpcRoute(
...
-        val res = rpcHandler.apply(req)
+
+        val res = {
+          if(req.isInstanceOf[RpcResponseError]) {
+            req}
+          else if(req.isInstanceOf[RpcRequest]) {
+            rpcHandler.apply(req.asInstanceOf[RpcRequest])}
+          else if(req.isInstanceOf[RpcBatchRequest]) {
+            rpcHandler.apply(req.asInstanceOf[RpcBatchRequest])}
+        }
```
i-Alex, line 96: Let's think more about the code structure here, it seems that we mix things:
1) from one perspective the parse of the body is done here and the knowledge of the the single/batch requests cases is here.
2) from other side, batch requests responses construction are done in `rpcHandler` with a lack of proper parsing of invalid request entries there.
3) error handling is spread between two classes: this one and rpcHandler.

Considering this, I feel there is some room for improvement.
`AccountEthRpcRoute` should be responsible for data parsing and collecting results.
`RpcHandler` should be responsible for single request processing.
In such case we may remove the `BatchRequest` class and leave  `RpcHandler` unchanged

i-Alex (changes requested):
Let me know your thoughts about the comments, we may discuss them before doing any changes.

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
davehorizen: My approach aimed to keep the management of input and output creation separate by maintaining the original structure of the `AccountEthRpcRoute` class method, but this led to an incorrect delegation of the error handling between the various components.

Your proposal, which I followed, seems very valid. The `RpcBatchRequest` class has been eliminated and now
the `RpcHandler` is limited to the execution of single rpc requests no longer discriminating between single request and batch. The `AccountEthRpcRoute` takes care of collecting the results instead.

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcId.java
davehorizen: Yes, we need this default constructor because we need to initialize a `RpcId` object to correctly serialize a null response id

sdk/src/main/java/com/horizen/account/api/rpc/handler/RpcHandler.java
davehorizen: As reported the `AccountEthRpcRoute` pr comment we refactor the previous implementation so right now the `RpcHandler` is handling single request only

sdk/src/main/java/com/horizen/account/api/rpc/request/RpcBatchRequest.java
davehorizen: Agree on this point, following the refactor the class `RpcBatchRequet` was eliminated from the project

i-Alex (approved):

PR #620, opened 2022-11-24
St/evm 444

saratnt:


i-Alex:
Because of task specification issue, this PR is a duplication.

PR #621, opened 2022-11-25
EVM-537 - return null result in Rpc response

davehorizen:


sdk/src/main/java/com/horizen/account/api/rpc/response/RpcResponseSuccess.java
```diff
@@ -14,6 +14,11 @@ public RpcResponseSuccess(RpcId id, Object result) {
         this.result = result;
     }
 
+    @JsonInclude
```
i-Alex, line 17: do we need `@JsonProperty("result")` above in line 9?

davehorizen: no, the _JsonProperty_ is not needed, it is now removed from the `RpcResponseSuccess` class. It is the same in the `RpcResponseError` class with the _error_ parameter.

i-Alex (approved):

PR #622, opened 2022-11-25
As/restore storages after restart

alsala:
Restore the node storages in the consistent state after node restart. Added also a py test.


sdk/src/main/scala/com/horizen/AbstractSidechainNodeViewHolder.scala
```diff
@@ -63,9 +63,16 @@ abstract class AbstractSidechainNodeViewHolder[
     new ConsensusValidator(timeProvider)
   )
 
+  // this method is called at the startup after the load of the storages from the persistent db. It might happen that the node was not
+  // stopped gracefully and therefore the consistency among storages might not be ensured. This method tries to recover this situation
+  def checkAndRecoverStorages(restoredData: Option[(HIS, MS, VL, MP)]): Option[(HIS, MS, VL, MP)]
+
   def dumpStorages(): Unit
```
i-Alex, line 70: `dumpStorages` can be defined in the Abstract class.
Both Utxo and Account model have the same implementation

alsala: Ok, moved in Abstract NVH

i-Alex (approved):

PR #623, opened 2022-11-25
libevm mega review

JcBernack:
- [x] [EVM-425](https://horizenlabs.atlassian.net/browse/EVM-425)
- [x] [EVM-430](https://horizenlabs.atlassian.net/browse/EVM-430)


akrckn (approved):

libevm/lib/refund.go
```diff
@@ -59,3 +62,11 @@ func (s *Service) setStateWithRefund(statedb *state.StateDB, address common.Addr
 	setStateRefunds(statedb, address, key, value)
 	statedb.SetState(address, key, value)
 }
+
+func (s *Service) StateGetRefund(params HandleParams) (error, hexutil.Uint64) {
```
i-Alex, line 66: Seeems method is not tested.
At the moment we use statedb method directly in the test: `if refund := statedb.GetRefund(); refund != tt.refund {`
Let's use the new method to test the instance properly.

i-Alex (changes requested):
Looks very good!
A minor point in UTs left.

i-Alex (approved):

PR #625, opened 2022-11-28
As/burn ft amount to smart contract

alsala:
FT amount sent to a smart contract now is burnt by adding it to the 0xdead address.


qa/SidechainTestFramework/scutil.py
```diff
@@ -1219,6 +1219,133 @@ def create_certificate_for_alien_sc(mcTest, scid, mc_node, fePatternArray):
     assert_equal(True, cert in mc_node.getrawmempool())
     return cert
 
+# 10 ^ 10
```
saratnt, line 1222: All the account-related methods and constant were moved to account/utils.py, so this part should be moved there again

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -326,7 +327,7 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
           secret match {
             case Some(secret) =>
 
-              val to = BytesUtils.toHexString(ForgerStakeMsgProcessor.ForgerStakeSmartContractAddress)
+              val to = BytesUtils.toHexString(FORGER_STAKE_SMART_CONTRACT_ADDRESS_BYTES)
```
saratnt, line 330: Should we use Java or Scala naming conventions for constants in Scala?

alsala: I vote for the uppercase way

sdk/src/main/scala/com/horizen/account/utils/WellKnownAddresses.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.horizen.utils.BytesUtils
+
+object WellKnownAddresses {
+
+  // fake smart contract address
+  val WITHDRAWAL_REQ_SMART_CONTRACT_ADDRESS_BYTES: Array[Byte] = BytesUtils.fromHexString("0000000000000000000011111111111111111111")
+  val FORGER_STAKE_SMART_CONTRACT_ADDRESS_BYTES: Array[Byte] = BytesUtils.fromHexString("0000000000000000000022222222222222222222")
+
+  // used to burn coins
+  val DEAD_ADDRESS_BYTES: Array[Byte] = BytesUtils.fromHexString("000000000000000000000000000000000000dead")
```
i-Alex, line 12: The null address shall be 0x000…000, like one of the most commonly used null addresses in Ethereum https://etherscan.io/address/0x0000000000000000000000000000000000000000.

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/utils/WellKnownAddresses.scala
alsala: Done

saratnt (approved):

PR #634, opened 2022-12-02
Fix/evm 454

GitMark0:


sdk/src/main/java/com/horizen/account/secret/PrivateKeySecp256k1.java
```diff
@@ -47,7 +44,7 @@ public SecretSerializer serializer() {
 
     @Override
     public AddressProposition publicImage() {
-        var hashedKey = Hash.sha3(this.publicKey);
+        var hashedKey = Secp256k1.sha3(this.publicKey);
         return new AddressProposition(Arrays.copyOfRange(hashedKey, hashedKey.length - Account.ADDRESS_SIZE, hashedKey.length));
```
i-Alex, line 48: let's put this code as a separate method of `byte[] Secp256k1.getAddress(byte[] publicKey)`
inside: calculate hash -> get last 20 bytes.

Another point that now we do not need to cache `this.publicKey`, instead we may cache directly the `address` field.

sdk/src/main/java/com/horizen/account/secret/PrivateKeySecp256k1Creator.java
```diff
@@ -32,16 +26,9 @@ public static PrivateKeySecp256k1Creator getInstance() {
 
     @Override
     public PrivateKeySecp256k1 generateSecret(byte[] seed) {
-        try {
-            ECKeyPair keyPair = Keys.createEcKeyPair(new SecureRandom(seed));
-            // keyPair private key can be 32 or 33 bytes long
```
i-Alex, line 37: So how do we manage 33 bytes case now?
I see that in the new code the commented line was removed and another line was added: "// Check why is this failing"

sdk/src/test/scala/com/horizen/fixtures/SecretFixture.scala
```diff
@@ -126,7 +126,8 @@ trait SecretFixture {
     val payload = "This is string to sign"
     val message = payload.getBytes(StandardCharsets.UTF_8)
     val pair = Keys.createEcKeyPair
-    new SignatureSecp256k1(Sign.signMessage(message, pair, true))
+    val signatureData = Sign.signMessage(message, pair, true)
+    new SignatureSecp256k1(signatureData.getV, signatureData.getR, signatureData.getS)
```
i-Alex, line 130: let's use `Secp256k1.sign` method here and in all other similar places in tests.
The goal is to avoid using `Sign.signMessage(message, pair, true)` directly as well as in the main code.

i-Alex (changes requested):
Looks good, a couple of places left.
Let's focus on the key generation exception.

i-Alex (approved):

sdk/src/main/java/com/horizen/account/secret/PrivateKeySecp256k1.java
```diff
@@ -71,7 +68,8 @@ public boolean owns(ProofOfKnowledgeProposition proposition) {
 
     @Override
     public SignatureSecp256k1 sign(byte[] message) {
-        return Secp256k1.sign(privateKey, message);
+        Secp256k1.Signature signature = Secp256k1.sign(privateKey, message);
+        return new SignatureSecp256k1(signature.v, signature.r, signature.s);
```
GitMark0, line 72: What is the reason behind this change?

i-Alex: To make `Secp256k1` class agnostic to SDK types structure. One of them is `SignatureSecp256k1`

alsala (approved):
Just minor things, for me it is OK, I have only a couple of doubts which I commented.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -68,7 +68,7 @@ public enum EthereumTransactionType {
     private synchronized String getTxHash() {
         if (this.hashString == null) {
             byte[] encodedMessage = encode(isSigned());
-            this.hashString = BytesUtils.toHexString(Hash.sha3(encodedMessage, 0, encodedMessage.length));
+            this.hashString = BytesUtils.toHexString((byte[]) Keccak256.hash(encodedMessage));
```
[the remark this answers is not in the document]
i-Alex: actually under the hood both of them use bouncycastle.
The goal here is:
1) to have a standard unified method for hashing, and honestly `Keccak256.hash` is more clear than `Hash.sha3`
2) to get id of Web3j at all.

sdk/src/main/java/com/horizen/account/utils/EthereumTransactionDecoder.java
```diff
@@ -13,12 +11,38 @@
...
+
 
 public class EthereumTransactionDecoder {
 
-    public EthereumTransactionDecoder() {
+    private EthereumTransactionDecoder() {
+        // prevent instantiation
+    }
+
+    // TODO: replace with EthereumTransaction.EthereumTransactionType
```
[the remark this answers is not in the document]
i-Alex: yes, it is just added in this PR.
I think, it will be addressed later by you in the https://horizenlabs.atlassian.net/browse/EVM-535 task

sdk/src/main/java/com/horizen/account/utils/EthereumTransactionDecoder.java
```diff
@@ -106,11 +130,15 @@ private static EthereumTransaction decodeLegacyTransaction(byte[] transaction) {
...
+    private static Long decodeEip155ChainId(byte[] bv) {
         long v = convertToLong(bv);
         if (v == LOWER_REAL_V || v == (LOWER_REAL_V + 1)) {
             return null;
         }
         return (v - CHAIN_ID_INC) / 2;
     }
+
+    private static byte[] getVFromRecId(int recId) {
+        return new byte[]{(byte)(27 + recId)};
```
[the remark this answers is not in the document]
i-Alex: done

sdk/src/main/java/com/horizen/account/utils/EthereumTransactionEncoder.java
```diff
@@ -115,4 +127,32 @@ public static byte[] encodeEip1559AsRlpValues(EthereumTransaction tx, boolean ac
                 .put(encoded)
                 .array();
     }
+
+    private static byte[] createEip155v(byte[] realV, long chainId) {
```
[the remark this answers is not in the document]
i-Alex: The goal is to get rid of web3j.
Regarding constants - done.

alsala (approved):

PR #638, opened 2022-12-06
EVM-552 - debug_traceBlockByHash rpc method

davehorizen:
In this pull request:
- debug_traceBlockByHash rpc method implementation
- python test sc_evm_test_debug_methods updated

### DIFF
--- a/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
+++ b/sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
@@ -522,7 +522,7 @@ class EthService(
 
   @RpcMethod("debug_traceBlockByHash")
   @RpcOptionalParameters(1)
-  def traceBlockByHash(hash: Hash, traceParams: TraceParams): DebugTraceBlockView = {
+  def traceBlockByHash(hash: Hash, config: TraceOptions): DebugTraceBlockView = {
     applyOnAccountView { nodeView =>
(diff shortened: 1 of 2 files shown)


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -515,6 +520,36 @@ class EthService(
     }
   }
 
+  @RpcMethod("debug_traceBlockByHash")
+  @RpcOptionalParameters(1)
+  def traceBlockByHash(hash: Hash, config: TraceOptions): DebugTraceBlockView = {
```
JcBernack, line 525: Please refactor to not duplicate so much code. The only difference between `traceBlockByHash` and `traceBlockByNumber` is that the `byNumber` version needs to find the block hash via block-number or "tag" first, everything else is identical:
- trace-by-hash is `hash => traceBlock(hash)`
- trace-by-number is `tag => traceBlock(getBlockIdByTag(tag))`

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Agree, method `traceBlockByNumber` refactored

i-Alex (approved):

PR #639, opened 2022-12-06
Fix/evm 473

GitMark0:


sdk/src/main/scala/com/horizen/account/receipt/EthereumReceipt.scala
```diff
@@ -11,70 +11,85 @@ import java.util
...
-  override def hashCode: Int =  {
+  override def hashCode: Int = {
     var result = consensusDataReceipt.hashCode()
     result = 31 * result + util.Arrays.hashCode(transactionHash)
     result = 31 * result + Integer.hashCode(transactionIndex)
     result = 31 * result + util.Arrays.hashCode(blockHash)
     result = 31 * result + Integer.hashCode(blockNumber)
     result = 31 * result + gasUsed.hashCode()
-    result = 31 * result + util.Arrays.hashCode(contractAddress)
+    result = 31 * result + util.Arrays.hashCode(contractAddress.get)
```
i-Alex, line 86: In case `contractAddress == None` -> it will throw an exception.
same in other places.

sdk/src/main/scala/com/horizen/account/receipt/EthereumReceipt.scala
```diff
@@ -128,10 +144,10 @@ object EthereumReceiptSerializer extends SparkzSerializer[EthereumReceipt]{
     val gasUsedLength: Int = reader.getInt
     val gasUsed: BigInteger = new BigInteger(reader.getBytes(gasUsedLength))
 
-    //optional field
+    // optional field
```
i-Alex, line 147: here we should refactor and serialize/deserialize  `contractAddress` as Option.
So in case the receipt is not for a smart contract declaration -> `None`.

Otherwise, what is the sense of using Option type?

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
```diff
@@ -80,10 +80,14 @@ class LogsBloom() extends BytesSerializable {
...
   override def equals(obj: Any): Boolean = {
     obj match {
-      case l: LogsBloom => (this.bloomFilter diff l.getBloomFilter()).isEmpty
+      case l: LogsBloom => this.bloomFilter.sameElements(l.getBloomFilter())
       case _ => false
     }
   }
+
+  override def hashCode(): Int = {
+    31.hashCode() * this.bloomFilter.hashCode()
```
i-Alex, line 89: I think `this.bloomFilter.hashCode()` is enough here.
why do we need `31.hashCode() *` part?

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -148,7 +148,7 @@ class AccountState(
 
             // get a receipt obj with non consensus data (logs updated too)
             val fullReceipt =
-              EthereumReceipt(consensusDataReceipt, txHash, txIndex, blockHash, blockNumber, txGasUsed, contractAddress)
+              EthereumReceipt(consensusDataReceipt, txHash, txIndex, blockHash, blockNumber, txGasUsed, Option(contractAddress))
```
i-Alex, line 151: A few line above, I expect `val contractAddress: Option[Byte[Array]]`.
And in case it is not a smart contract declaration -> set it to `None` instead of `new Array[Byte](0)`

sdk/src/test/scala/com/horizen/account/receipt/ReceiptFixture.scala
```diff
@@ -46,7 +46,7 @@ trait ReceiptFixture {
     val receipt = EthereumReceipt(consensusDataReceipt,
       txHash, 33, Keccak256.hash("blockhash".getBytes).asInstanceOf[Array[Byte]], 22,
       BigInteger.valueOf(1234567),
-      contractAddress
+      Option(contractAddress)
```
i-Alex, line 49: this is a mistake.
In case `!contractAddressPresence` value is `None`, not a `Some(new Array[Byte](0)`

i-Alex (changes requested):
Please, check comments in the Jira tasks. I put the notes for the missing points.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumReceiptView.java
```diff
@@ -39,7 +39,7 @@ public EthereumReceiptView(EthereumReceipt receipt, EthereumTransaction tx, BigI
         to = (tx.getTo().isPresent()) ? Numeric.toHexString(tx.getTo().get().address()) : null;
         cumulativeGasUsed = Numeric.encodeQuantity(receipt.consensusDataReceipt().cumulativeGasUsed());
         gasUsed = Numeric.encodeQuantity(receipt.gasUsed());
-        contractAddress = receipt.contractAddress().length != Account.ADDRESS_SIZE ? null : Numeric.toHexString(receipt.contractAddress());
+        contractAddress = receipt.contractAddress().get().length != Account.ADDRESS_SIZE ? null : Numeric.toHexString(receipt.contractAddress().get());
```
JcBernack, line 42: I don't think we need to check the length anymore. Previously length could zero or ADDRESS_SIZE, but now it will always be ADDRESS_SIZE - or the whole thing is `None`.

sdk/src/main/scala/com/horizen/account/receipt/EthereumReceipt.scala
```diff
@@ -101,9 +116,9 @@ object EthereumReceiptSerializer extends SparkzSerializer[EthereumReceipt]{
     writer.putInt(gasUsedBytes.length)
     writer.putBytes(gasUsedBytes)
 
-    //optional field
-    writer.putInt(receipt.contractAddress.length)
-    writer.putBytes(receipt.contractAddress)
+    // optional field
+    writer.putInt(receipt.contractAddress.get.length)
+    writer.putBytes(receipt.contractAddress.get)
```
JcBernack, line 121: Making `contractAddress` an `Option[]` is good, but just calling `get()` everywhere is not using it right. This will throw exceptions whenever it is `None`. Instead we need to explicitly handle the `None` case.

You could do something like this:
```suggestion
    val addr = receipt.contractAddress.getOrElse(Array.empty)
    writer.putInt(addr.length)
    writer.putBytes(addr)
```

sdk/src/main/scala/com/horizen/account/receipt/LogsBloom.scala
JcBernack: I did these changes already in this PR: https://github.com/HorizenOfficial/Sidechains-SDK/pull/644/files#diff-deaa0e70e60ddbc5961d8b6188c566623deeea8110511599db00b24c0d5b1232R98
I used `util.Arrays.hashCode(filter)` as we want the hashCode to depend on the contents of the bitmask. I'm not sure if `filter.hashCode()` would do that too.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/receipt/EthereumReceipt.scala
JcBernack: @GitMark0 This is still missing. Read the length, if zero contractAddress should be `None`.

i-Alex: fxied

i-Alex (approved):

PR #641, opened 2022-12-06
Signers keys in bootstrapping tool should not be equal to masters keys

sgonta:


sgonta:
Opened PR from another branch name

PR #645, opened 2022-12-09
SC bootstrap tool improvements

akrckn:
use key generator, remove comment and avoid using block version as marker for account or utxo model

### DIFF
--- a/tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
+++ b/tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.horizen.account.secret.PrivateKeySecp256k1Creator;
 import com.horizen.account.utils.FeeUtils;
 import com.horizen.account.block.AccountBlock;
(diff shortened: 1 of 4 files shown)


tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
```diff
@@ -506,12 +504,12 @@ private void printGenesisInfoUsageMsg(String error) {
                       "\t\t\"updateconfig\": boolean - Optional. Default false. If true, put the results in a copy of source config.\n" +
                       "\t\t\"sourceconfig\": <path to in config file> - expected if 'updateconfig' = true.\n" +
                       "\t\t\"resultconfig\": <path to out config file> - expected if 'updateconfig' = true.\n" +
-                      "\t\t\"blockversion\": int - Optional, default = 1. UTXO model block version=1, Account model block version=2.\n" +
+                      "\t\t\"model\": String - Optional, default = 'utxo', 'account' model.\n" +
```
i-Alex, line 507: we need to update STF tests to start using `model` instead of `blockversion`.
@akrckn do you want to do it by yourself?

i-Alex (changes requested):

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
akrckn: Yes, updated the tests here, but still running - I'll resolve the conversation and re-request after all tests are executed.

akrckn:
Just a side note: I saw a new python test, maybe that one should be rewritten to use the `ac_chain_setup`, too - but not in this PR.

i-Alex (approved):

PR #652, opened 2022-12-14
Merge dev -> dev_evm

i-skrypnyk:
Update: All unit and python tests passed


alsala (changes requested):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -433,6 +460,11 @@ class AccountState(
   def isWithdrawalEpochLastIndex: Boolean = {
     WithdrawalEpochUtils.isEpochLastIndex(getWithdrawalEpochInfo, params)
   }
+
+  override def utxoMerkleTreeRoot(withdrawalEpoch: Int): Option[Array[Byte]] = {
+    // TODO: no CSW support expected for the Eth sidechain
+    Some(Array[Byte]())
```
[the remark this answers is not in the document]
i-skrypnyk: I want this Option to be defined but contain an empty array. Otherwise, there's a check for it up in a trace that will fail

sdk/src/main/scala/com/horizen/account/certificatesubmitter/AccountCertificateSubmitter.scala
```diff
@@ -39,29 +46,41 @@ class AccountCertificateSubmitter(settings: SidechainSettings,
   type MP = AccountMemoryPool
   type PM = AccountBlock
 
-  override def getUtxoMerkleTreeRoot(referencedEpoch: Int, state: AccountState): Optional[Array[Byte]] = Optional.empty()
-
   override def getWithdrawalRequests(state: AccountState, referencedEpochNumber: Int): Seq[BackwardTransfer] =
```
i-Alex, line 49: no one is using this method, it was replaced by `AbstractState.backwardTransfers`.
So we may remove it from here, from base class and from UTXO class as it was done for `getUtxoMerkleTreeRoot`

sdk/src/main/scala/com/horizen/certificatesubmitter/strategies/CeasingSidechain.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.horizen.params.NetworkParams
+import com.horizen.utils.{BytesUtils, WithdrawalEpochInfo, WithdrawalEpochUtils}
+import com.horizen.websocket.client.{MainchainNodeChannel, WebsocketErrorResponseException, WebsocketInvalidErrorMessageException}
+import com.horizen.{AbstractHistory, AbstractState}
+import scorex.util.{ModifierId, ScorexLogging}
+import sparkz.core.NodeViewHolder.CurrentView
+
+import scala.util.{Failure, Success, Try}
+
+class CeasingSidechain(mainchainChannel: MainchainNodeChannel, params: NetworkParams)
```
i-Alex, line 12: As an option, we may make this class parameterized by History and State, rather that have only the generic method inside.
I think it will feet better the whole design.

sdk/src/main/scala/com/horizen/certificatesubmitter/strategies/CertificateSubmissionStrategy.scala
```diff
@@ -0,0 +1,14 @@
+package com.horizen.certificatesubmitter.strategies
+
+import com.horizen.certificatesubmitter.AbstractCertificateSubmitter.SignaturesStatus
+import com.horizen.{AbstractHistory, AbstractState}
+import scorex.util.ModifierId
+import sparkz.core.NodeViewHolder.CurrentView
+
+trait CertificateSubmissionStrategy {
```
i-Alex, line 8: same here

sdk/src/main/scala/com/horizen/certificatesubmitter/strategies/WithKeyRotationCircuitStrategy.scala
```diff
@@ -0,0 +1,180 @@
...
+class WithKeyRotationCircuitStrategy[
+  TX <: Transaction,
+  H <: SidechainBlockHeaderBase,
+  PM <: SidechainBlockBase[TX, H] : ClassTag,
+  _FPI <: AbstractFeePaymentsInfo,
+  _HSTOR <: AbstractHistoryStorage[PM, _FPI, _HSTOR],
+  _HIS <: AbstractHistory[TX, H, PM, _FPI, _HSTOR, _HIS],
+  _MS <: AbstractState[TX, H, PM, _MS],
+  _VL <: Wallet[SidechainTypes#SCS, SidechainTypes#SCP, TX, PM, _VL],
+  _MP <: MemoryPool[TX, _MP]](settings: SidechainSettings, params: NetworkParams,
```
i-Alex, line 34: I have a feeling that here we have some mess with the generics and with the type definitions.
In case we decide to refactor it we need to start from `AbstractCertifcateSubmitter` and move `type` definitions to class templates.

sdk/src/test/scala/com/horizen/api/http/SidechainSubmitterApiRouteTest.scala
```diff
@@ -0,0 +1,86 @@
...
+import scala.collection.JavaConverters._
+import scala.language.postfixOps
+
+class SidechainSubmitterApiRouteTest extends SidechainApiRouteTest {
+
+  override val basePath = "/submitter/"
+
+  "The Api should to" should {
+
+//    "reply at /getKeyRotationProofs" in {
```
i-Alex, line 21: do not forget to uncomment and fix tests

i-Alex (changes requested):
Amazing work!
Let's focus on the comments first and on the comments in the original PR.
Then we may proceed with the last step - python tests for key rotation and non-ceasing for account model.

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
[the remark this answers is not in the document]
i-skrypnyk: Ok, to double-check, here's a part of the code that uses this method, if account model indeed cannot be with CSW enabled that we should not ever call it:
image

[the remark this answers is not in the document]
i-skrypnyk: https://horizenlabs.atlassian.net/browse/EVM-598

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -545,6 +550,58 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
     }
   }
 
+  def createKeyRotationTransaction: Route = (post & path("createKeyRotationTransaction")) {
```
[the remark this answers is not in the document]
i-skrypnyk: Thanks, fixed the first part of the comment. Will work on the python test in a next phase

sdk/src/main/scala/com/horizen/certificatesubmitter/strategies/CeasingSidechain.scala
i-skrypnyk: discussed below

sdk/src/main/scala/com/horizen/certificatesubmitter/strategies/CertificateSubmissionStrategy.scala
i-skrypnyk: discussed below

sdk/src/test/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessorTest.scala
```diff
@@ -80,12 +81,14 @@ class CertificateKeyRotationMsgProcessorTest
   }
 
   private def buildKeyRotationProof(`type`: KeyRotationProofType, index: Int, newKey: SchnorrSecret, oldSigningKey: SchnorrSecret, oldMasterKey: SchnorrSecret) = {
+    val messageToSign = SchnorrPublicKey.deserialize(newKey.publicImage().pubKeyBytes()).getHash.serializeFieldElement()
```
i-Alex, line 84: memory leak here.
Need to "free" `SchnorrPublicKey`.
The same bug we have in other places (both account and utxo)
Let's move to utility function and be agnostic to cryptolib types.

i-Alex (changes requested):

alsala (approved):

sgonta (approved):

i-skrypnyk:
Update: All unit and python tests passed

sdk/src/main/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessor.scala
```diff
@@ -66,10 +66,10 @@ case class CertificateKeyRotationMsgProcessor(params: NetworkParams) extends Fak
       .getOrElse(configKey)
   }
 
-  private def getLatestSigningKey(view: BaseAccountStateView, configKey: SchnorrProposition, epoch: Int, index: Int): SchnorrProposition = {
+  private def getLatestSigningKey(view: BaseAccountStateView, configKey: SchnorrProposition, requestedEpoch: Int, index: Int): SchnorrProposition = {
     val signingKeyChangeHistory = getKeysRotationHistory(SigningKeyRotationProofType, index, view)
 
-    signingKeyChangeHistory.epochNumbers.find(epoch > _)
+    signingKeyChangeHistory.epochNumbers.find(_ < requestedEpoch)
```
i-Alex, line 72: let's do the same for `getLatestMasterKey` method

sdk/src/test/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessorTest.scala
i-Alex: this one will be fixed in 0.6.0-snapshot3 and ported to this PR.

sdk/src/main/java/com/horizen/node/NodeState.java
```diff
@@ -1,8 +1,14 @@
...
 import com.horizen.state.SidechainStateReader;
+import com.horizen.utils.WithdrawalEpochInfo;
 
 public interface NodeState extends NodeStateBase, SidechainStateReader {
 
     boolean hasCeased();
+
+    NetworkParams params();
+
+    WithdrawalEpochInfo getWithdrawalEpochInfo();
```
i-Alex, line 13: why do we need this new interfaces?
I tempted to remove them from `NodeState`

i-Alex (changes requested):

sdk/src/main/java/com/horizen/node/NodeState.java
i-skrypnyk: SidechainNodeView uses a NodeState type in it that we cannot override, and we cannot use other View type cause  SidechainApiRoute has a SidechainNodeViewBase type in it, which is only extended by SidechainNodeView.

Its another mess with types, I can show you in a bit more detail on a call

i-Alex (approved):
We will merge here 0.6.0-snapshot3 when it is ready and then will merge this PR.

PR #654, opened 2022-12-16
EVM-548 - eth_syncing rpc method

davehorizen:
In this pull request:
- eth_syncing rpc method and ut

### DIFF
--- a/sdk/src/test/scala/com/horizen/account/api/rpc/service/EthServiceTest.scala
+++ b/sdk/src/test/scala/com/horizen/account/api/rpc/service/EthServiceTest.scala
@@ -190,6 +190,11 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
   }
 
   @Test
+  def eth_syncing(): Unit = {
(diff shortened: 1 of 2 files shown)


i-Alex (approved):

sdk/src/test/scala/com/horizen/account/api/rpc/service/EthServiceTest.scala
```diff
@@ -189,6 +189,11 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
     )
   }
 
+  @Test
+  def eth_syncing(): Unit = {
+    assertEquals(false, ethService.execute(getRpcRequest()).asInstanceOf[Quantity].value)
```
JcBernack, line 194: This will fail, as the function returns `Boolean` not `Quantity`.

JcBernack (changes requested):

sdk/src/test/scala/com/horizen/account/api/rpc/service/EthServiceTest.scala
davehorizen: Right! UT fixed

i-Alex (approved):

PR #657, opened 2022-12-16
Fixed potential null pointer dereference.

VolodymyrBabaiev:
Fixed potential null pointer dereference.

### DIFF
--- a/sdk/src/main/java/com/horizen/account/api/rpc/types/TransactionArgs.java
+++ b/sdk/src/main/java/com/horizen/account/api/rpc/types/TransactionArgs.java
@@ -117,6 +117,12 @@ public class TransactionArgs {
      * Reimplementation of the same logic in GETH.
      */
     public Message toMessage(BigInteger baseFee) throws RpcException {
+        if (baseFee == null) {
+            // Practically it's not possible. Because baseFee is always arrived from block header and every block header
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/horizen/account/api/rpc/types/TransactionArgs.java
```diff
@@ -153,7 +153,10 @@ public Message toMessage(BigInteger baseFee) throws RpcException {
             }
             // Backfill the legacy gasPrice for EVM execution, unless we're all zeroes
             if (gasFeeCap.bitLength() > 0 || gasTipCap.bitLength() > 0) {
-                effectiveGasPrice = baseFee.add(gasTipCap).min(gasFeeCap);
+                if (baseFee != null) {
```
JcBernack, line 156: Is there a reason `baseFee` should ever be `null`? I think we should rather throw an `IllegalArgumentException` if it is.

VolodymyrBabaiev: It's not possible that baseFee will ever be null. This change was requester from NCC audit.

JcBernack (approved):

davehorizen (approved):

i-Alex (approved):

alsala (approved):

PR #658, opened 2022-12-16
St/evm 464

saratnt:


sdk/src/main/scala/com/horizen/SidechainApp.scala
```diff
@@ -248,6 +261,8 @@ class SidechainApp @Inject()
       SidechainWalletCswDataProviderCSWDisabled()
   }
 
+  //Waits for actors to start
+  Thread.sleep(500)
```
i-Alex, line 265: this part is not clear to me.
why do we need to sleep?
how it worked before?

sdk/src/main/scala/com/horizen/SidechainAppModule.scala
```diff
@@ -28,6 +27,9 @@ abstract class SidechainAppModule extends com.google.inject.AbstractModule {
     bind(classOf[TransactionSubmitHelper])
       .to(classOf[TransactionSubmitHelperImpl])
 
+    bind(classOf[AbstractSidechainApp])
+      .to(classOf[SidechainApp])
```
i-Alex, line 31: why do we need this?
same for account model class

sdk/src/main/scala/com/horizen/account/AccountSidechainApp.scala
```diff
@@ -149,30 +144,41 @@ class AccountSidechainApp @Inject()
...
+    PMOD,
+    AccountFeePaymentsInfo,
+    NodeAccountHistory,
+    NodeAccountState,
+    NodeWalletBase,
+    NodeAccountMemoryPool,
+    AccountNodeView] = nodeViewProvider
+
+  //Waits for actors to start
+  Thread.sleep(500)
```
i-Alex, line 181: same question here

sdk/src/main/scala/com/horizen/SidechainApp.scala
saratnt: I have moved to the base class some operations that were before performed between the creation of the CertificateSubmitter actor and the publishing of SidechainApplicationStart event. So it happens that the CertificateSubmitter actor is not able to subscribe to the SidechainApplicationStart event in time to receive it and this causes that the CertificateSubmitter never starts. I have tried to see if there were ways to know if an actor has started or not, but I haven't found anything so I have put a sleep. Probably I can reduce the sleep time, I was very conservative. Same for AccountSidechainApp.

sdk/src/main/scala/com/horizen/SidechainAppModule.scala
saratnt: This way I was able to make SecretSubmitHelperImpl class work for both Account and UTXO models (`class SecretSubmitHelperImpl @Inject()(val appProvider: Provider[AbstractSidechainApp])`)

sdk/src/main/scala/com/horizen/account/AccountSidechainApp.scala
saratnt: Same reply

sdk/src/main/scala/com/horizen/SidechainApp.scala
i-Alex: let's check deeply the behavior

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
```diff
@@ -265,7 +265,7 @@ object SidechainNodeViewHolderRef {
             genesisBlock: SidechainBlock)
            (implicit system: ActorSystem): ActorRef =
     system.actorOf(props(sidechainSettings, historyStorage, consensusDataStorage, stateStorage, forgerBoxStorage, utxoMerkleTreeProvider, walletBoxStorage, secretStorage,
-      walletTransactionStorage, forgingBoxesInfoStorage, cswDataProvider, backupStorage, params, timeProvider, applicationWallet, applicationState, genesisBlock).withMailbox("akka.actor.deployment.prio-mailbox"))
```
i-Alex, line 268: `.withMailbox("akka.actor.deployment.prio-mailbox"))` is critical to have because we need to prioritize some messages.
Why was it removed from this and other places?

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/SidechainNodeViewHolder.scala
saratnt: It was removed because in `props` method is already called

i-Alex: got it, thanks!

i-Alex (approved):

PR #662, opened 2022-12-16
EVM-550/EVM-551 - txpool_status and txpool_content rpc methods

davehorizen:


sdk/src/main/java/com/horizen/account/api/rpc/types/TxPoolTransaction.java
```diff
@@ -0,0 +1,85 @@
...
+import org.web3j.utils.Numeric;
+
+import java.math.BigInteger;
+
+@JsonView(Views.Default.class)
+@JsonInclude()
+public class TxPoolTransaction {
+
+    private final static String DEFAULT_BLOCK_HASH = "0x0000000000000000000000000000000000000000000000000000000000000000";
+    private final String blockHash;         // fixed default value
```
i-Alex, line 15: as was mentioned in other PR, let's make these field `public` and remove all get methods like
```
public String getBlockHash() {
        return blockHash;
    }
```
it is enough for the proper json serialization

i-Alex (changes requested):

sdk/src/main/java/com/horizen/account/api/rpc/types/TxPoolTransaction.java
davehorizen: modifiers switch to `public` and getter methods removed

PR #664, opened 2022-12-19
Non ceasing pending certs

i-Alex:
Non-ceasing sidechain zen 4.0.0 compatibility fixes:
1) fixing the issue with certificate rejection in case of previous certificate delay).
2) MC block ref data structure has been reverted to be identical to the ceasing case, since no more multiple certificates for different epochs are allowed.
3) memory leak fix for zendoo-sc-cryptolib

Test with MC tag `v4.0.0-rc2`

### DIFF
--- a/qa/sc_multiple_pending_certs_non_ceasing.py
+++ b/qa/sc_multiple_pending_certs_non_ceasing.py
@@ -37,8 +37,12 @@ Test:
     - Check that certificate submission is in progress on SC node 1
     - Wait for the cert in MC mempool and generate 1 MC block and 1 SC block with that Cert.
     - Check that we started generating certificate for the next epoch in the queue.
+    -- Verify that first certificate endCumulativeScTxCommitmentTreeRoot equals to last one in the withdrawal epoch 0.
     - Repeat previous 2 steps to see that we generated all certificates except the one for epoch 5 (not a moment)
+    -- Verify that certificate endCumulativeScTxCommitmentTreeRoot equals to the one that contains previous epoch cert.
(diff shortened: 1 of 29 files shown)


qa/sc_multiple_pending_certs_non_ceasing.py
```diff
@@ -37,8 +37,12 @@
     - Check that certificate submission is in progress on SC node 1
     - Wait for the cert in MC mempool and generate 1 MC block and 1 SC block with that Cert.
     - Check that we started generating certificate for the next epoch in the queue.
+    -- Verify that first certificate endCumulativeScTxCommitmentTreeRoot equals to last one in the withdrawal epoch 0.
     - Repeat previous 2 steps to see that we generated all certificates except the one for epoch 5 (not a moment)
+    -- Verify that certificate endCumulativeScTxCommitmentTreeRoot equals to the one that contains previous epoch cert.
+       Note: due to the certificate timing check in the MC we need to shift the endCumulativeScTxCommitmentTreeRoot.
     - Generate more MC and SC blocks. Check the submission of the Cert for epoch 5.
+    -- Verify that certificate endCumulativeScTxCommitmentTreeRoot equals to last one in the withdrawal epoch 5.
```
VolodymyrBabaiev, line 45: Let's use one dash in the beginning of every item as it was before.

i-skrypnyk (approved):

qa/sc_multiple_pending_certs_non_ceasing.py
i-Alex: done

VolodymyrBabaiev (approved):

PR #668, opened 2022-12-21
Potential Signature Forgery Due to

sgonta:
Noncanonical Encoding of message to
FieldElement


davehorizen (changes requested):
lgtm but the unit tests of the `ConsensusPackageTests` need to be updated to manage the check on the field length

sdk/src/main/java/com/horizen/cryptolibprovider/FieldElementUtils.java
```diff
@@ -12,11 +12,11 @@ public static int fieldElementLength() {
     }
 
     public static FieldElement messageToFieldElement(byte[] message) {
-        if (message.length > fieldElementLength()) {
+        if (message.length != fieldElementLength()) {
```
alsala, line 15: There are lots of UT failing for this modification. As a matter of fact we do use messages shorter than 32 bytes. I think we should opt for keeping the original code and explicitly 
left-padding byte arrays prior to calling FieldElement.deserialize(), as described in the NCC recommendation paragraph.

alsala (changes requested):

sgonta:
This is not the final code. PR is Draft. No need to merge. First of all need NCC approval of approach

sgonta:
Ready for review, not a draft anymore

i-Alex (changes requested):
Some Unit tests are failing.
Same to python tests, due to a wrong usage/implementation of `messageToFieldElement` method.
Solution is incomplete.

sdk/src/main/java/com/horizen/cryptolibprovider/FieldElementUtils.java
sgonta: private def generateHashAndCleanUp(elements: Array[Byte]*): Array[Byte] = {
    val digest = PoseidonHash.getInstanceConstantLength(elements.length)
    elements.foreach { message =>
      if (message.length > fieldElementLength) {
        throw new IllegalArgumentException("Message length is exceed allowed message len. Message len " +
          message.length + " but it shall be less than " + fieldElementLength)
      }
      val fieldElement = FieldElement.deserialize(util.Arrays.copyOf(message, fieldElementLength))
      digest.update(fieldElement)
      fieldElement.freeFieldElement()
    }
    val hash = digest.finalizeHash()
    val result = hash.serializeFieldElement()
    digest.freePoseidonHash()
    hash.freeFieldElement()
    result
  }

sgonta: @alsala so in the place where lenght is shorter than 32, I implemented in place

sgonta:
[Signature_forgery.pdf](https://github.com/HorizenOfficial/Sidechains-SDK/files/10537769/Signature_forgery.pdf)

sdk/src/main/java/com/horizen/cryptolibprovider/utils/FieldElementUtils.java
```diff
@@ -11,6 +11,14 @@ public static int fieldElementLength() {
...
 
+    public static FieldElement messageToFieldElementStrict(byte[] message) {
+        if (message.length != fieldElementLength()) {
+            throw new IllegalArgumentException("Message length is exceed allowed message len. Message len " +
+                    message.length + " but it shall be equal to " + fieldElementLength());
+        }
+        return FieldElement.deserialize(message);
+    }
+
     public static FieldElement messageToFieldElement(byte[] message) {
```
alsala, line 22: If I am not wrong we agreed to change this method name because it is not meant anymore to deal with messages to sign. And also to change the variable names in the code of the method `generateHashAndCleanUp()` in `sdk/src/main/scala/com/horizen/consensus/package.scala`

alsala (changes requested):

sdk/src/main/java/com/horizen/cryptolibprovider/utils/FieldElementUtils.java
sgonta: Renamed message to element

alsala (approved):

qa/SidechainTestFramework/sc_test_framework.py
```diff
@@ -41,6 +41,9 @@
 '''
 class SidechainTestFramework(BitcoinTestFramework):
 
+    def __init__(self):
+        self.number_of_sidechain_nodes = 1
+
```
i-Alex, line 46: why do we need this change?

sdk/src/main/java/com/horizen/cryptolibprovider/implementations/ThresholdSignatureCircuitWithKeyRotationImplZendoo.java
```diff
@@ -133,12 +134,23 @@ public byte[] generateMessageToBeSigned(List<BackwardTransfer> bt,
                 btrFee,
                 customFieldsElements
         );
-        CreateProofResult proofAndQuality = null;
+        Optional<String> proofAndQuality = null;
         try {
-            proofAndQuality = NaiveThresholdSignatureWKeyRotation.createProof(validatorKeysUpdatesList,
+
```
i-Alex, line 139: why do I see these changes?

i-Alex (changes requested):

qa/SidechainTestFramework/sc_test_framework.py
sgonta: Test was failing without it. Removed

sdk/src/main/java/com/horizen/cryptolibprovider/implementations/ThresholdSignatureCircuitWithKeyRotationImplZendoo.java
```diff
@@ -164,7 +165,7 @@ public byte[] generateMessageToBeSigned(List<BackwardTransfer> bt,
         signatures.forEach(SchnorrSignature::freeSignature);
         customFieldsElements.forEach(FieldElement::freeFieldElement);
 
-        return new Pair<>(proofAndQuality.getProof(), proofAndQuality.getQuality());
+        return new Pair<>(Ints.toByteArray(1), 2L);
```
i-Alex, line 168: check the PR changes:
still see inconsistent changes in this file

sgonta:
> Some Unit tests are failing. Same to python tests, due to a wrong usage/implementation of `messageToFieldElement` method. Solution is incomplete.

That's because we need to merge changes from https://github.com/HorizenOfficial/Sidechains-SDK/pull/680 related to messageToSign

i-Alex (approved):

davehorizen (approved):

PR #670, opened 2022-12-21
EVM-614 - txpool_content and txpool_status rpc methods

davehorizen:


sdk/src/main/java/com/horizen/account/api/rpc/types/TxPoolTransaction.java
```diff
@@ -0,0 +1,85 @@
...
+import org.web3j.utils.Numeric;
+
+import java.math.BigInteger;
+
+@JsonView(Views.Default.class)
+@JsonInclude()
+public class TxPoolTransaction {
+
+    private final static String DEFAULT_BLOCK_HASH = "0x0000000000000000000000000000000000000000000000000000000000000000";
+    private final String blockHash;         // fixed default value
```
i-Alex, line 15: can be `public` then we don't need multiple methods like 
```
public String getBlockHash() {
        return blockHash;
    }
```

i-Alex (approved):
The comment can be addressed in the original PR to dev_evm: https://github.com/HorizenOfficial/Sidechains-SDK/pull/662

PR #675, opened 2022-12-22
Fix/python tests cert rotation

i-skrypnyk:


qa/sc_evm_cert_key_rotation.py
```diff
@@ -97,10 +98,39 @@ def run_test(self):
...
+        private_master_keys.append(new_master_key.secret)
+        public_master_keys.append(new_public_key_3)
+        private_master_keys.append(new_signing_key_4.secret)
+        public_master_keys.append(new_public_key_4)
+
+        # Change ALL the signing keys and ALL tee master keys
+        new_signing_keys = []
+        new_master_keys = []
+        for i in range(cert_max_keys):
+            new_s_key = generate_cert_signer_secrets("random_seed5", 1)[0]
```
i-Alex, line 123: it seems that here w generate `cert_max_keys` times the same key, because of using the same randomness. Let's account index `i` for the see as well.

PS: actually they are different, because currently key generation for given key type is not deterministic.

Same below.

i-Alex (changes requested):
LGTM!
minor thing regarding key generation in tests

qa/sc_evm_cert_key_rotation.py
i-skrypnyk: fixed!

i-Alex (approved):

PR #678, opened 2022-12-23
M2-V5 - backport

davehorizen:


sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -545,15 +545,15 @@ class EthService(
 
   @RpcMethod("debug_traceBlockByHash")
   @RpcOptionalParameters(1)
-  def traceBlockByHash(hash: Hash, config: TraceOptions): DebugTraceBlockView = {
+  def traceBlockByHash(hash: Hash, traceOption: TraceOptions): DebugTraceBlockView = {
```
i-Alex, line 548: i think the name `config` is the right one according to the docs.
Same in traceBlockByNumber and traceTransaction

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: input parameters renamed in config for `traceBlockByNumber` and `traceTransaction` rpc methods

i-Alex (approved):

PR #689, opened 2023-01-08
Updated outdated dependencies

sgonta:


i-Alex (changes requested):
Dependencies looks good.

I saw that `zendoo-sc-cryptolib` was added as a submodule.
please, remove it, since it is not a part of SDK project.

i-Alex (approved):

i-Alex (changes requested):
Check the PR, something wrong was pushed into this branch, contains a mix of strange changes

sdk/pom.xml
```diff
@@ -186,7 +186,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
-      <version>2.13.3</version>
+      <version>2.14.1</version>
```
i-Alex, line 189: Try to run application.
We can't update jackson dependency as is to `2.14.1`.
Assume that it requires to have some updates starting from Sparkz.
For now we may keep it as is.

i-Alex (changes requested):

sdk/pom.xml
i-Alex: Try to run application.

i-Alex:
This PR has been reverted.
We need to open a new one.

i-Alex:
Substituted by https://github.com/HorizenOfficial/Sidechains-SDK/pull/704

PR #690, opened 2023-01-10
Python test threading fix

i-skrypnyk:
Changed the thread type to daemon, so that it is terminated whenever the main thread exits


i-Alex (approved):
Good!
Looks much simpler

PR #691, opened 2023-01-10
As/code review evm476

alsala:


saratnt (approved):

sdk/src/main/scala/com/horizen/account/state/InvalidMessageExceptions.scala
```diff
@@ -51,6 +51,13 @@ case class GasUintOverflowException() extends InvalidMessageException("gas uint6
 case class IntrinsicGasException(have: BigInteger, want: BigInteger)
   extends InvalidMessageException(s"intrinsic gas too low: have $have, want $want")
 
+/*** TODO check this:
+ * No one is currently throwing these three exceptions.
+ * We check those conditions in EthereumTransaction.semanticValidity() method,
+ * but throwing `TransactionSemanticValidityException`. We could use them there instead, but
+ * then we would break the uniformity of the code and would not add anything useful since the
+ * exception message is already detailed
```
i-Alex, line 59: Agree, keep using `TransactionSemanticValidityException` exception in semantic validity seems good and consistent.
We may remove this commented section at all.

i-Alex (approved):

PR #696, opened 2023-01-13
EVM-470

JcBernack:
- [EVM-470](https://horizenlabs.atlassian.net/browse/EVM-470?atlOrigin=eyJpIjoiMzhiNWViNWY3NGNjNGZmZWIyZTM5MDgzOWI1MjAwNGMiLCJwIjoiaiJ9)

[EVM-470]: https://horizenlabs.atlassian.net/browse/EVM-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


saratnt (approved):

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -110,65 +118,61 @@ class AccountForgeMessageBuilder(
...
+              case _: FeeCapTooLowException =>
+                // stop forging because all the remaining txs cannot be executed for the nonce, if they are from the
+                // same account, or, if they are from other accounts, they will have a lower fee cap
+                return Success(receiptList, listOfTxsInBlock, cumBaseFee, cumForgerTips)
+              case e: NonceTooLowException =>
+                // SHOULD NEVER HAPPEN, but in case just skip this tx
+                log.error(s"******** Could not apply tx for NonceTooLowException ******* : ${e.getMessage}")
+                priceAndNonceIter.next()
+              case e =>
+                // skip all txs from the same account but remove any changes caused by the rejected tx
```
i-Alex, line 170: "but remove any changes caused by the rejected tx" comment is not actual anymore, because the revert is not anymore processed here, but instead inside `applyTransaction`

sdk/src/main/scala/com/horizen/account/state/GasPool.scala
```diff
@@ -12,6 +12,8 @@ class GasPool(initialGas: BigInteger) extends SparkzLogging {
 
   def getUsedGas: BigInteger = initialGas.subtract(currentGas)
 
+  def getInitialGas: BigInteger = initialGas
```
i-Alex, line 15: method is not used.
In case we want to have it as a part of the interface we may do:
`class GasPool(initialGas: BigInteger)` -> `case class GasPool(initialGas: BigInteger)`

i-Alex (changes requested):
looks good!

sdk/src/main/scala/com/horizen/account/state/GasPool.scala
JcBernack: This class has mutable internal state, aren't case classes supposed to be immutable? :thinking:

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
JcBernack: agree, will remove

sdk/src/main/scala/com/horizen/account/state/GasPool.scala
i-Alex: good point, then we may have:
 class GasPool(**val** initialGas: BigInteger)

JcBernack: done

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
JcBernack: done

i-Alex (approved):

PR #701, opened 2023-01-20
EVM-628 - debug rpc methods - geth native tracer 

davehorizen:
In this pull request:
- `debug_traceBlockByNumber`, `debug_traceBlockByHash` and `debug_traceTransaction` debug rpc methods updated to use geth native tracers (struct/opcode logger, callTracer and 4byteTracer)
- go unit tests added


libevm/lib/evm.go
```diff
@@ -40,10 +44,31 @@ type EvmContext struct {
...
-	EnableMemory     bool `json:"enableMemory"`
-	DisableStack     bool `json:"disableStack"`
-	DisableStorage   bool `json:"disableStorage"`
-	EnableReturnData bool `json:"enableReturnData"`
+	EnableMemory     bool          `json:"enableMemory"`
+	DisableStack     bool          `json:"disableStack"`
+	DisableStorage   bool          `json:"disableStorage"`
+	EnableReturnData bool          `json:"enableReturnData"`
+	Tracer           *string       `json:"tracer"`
+	Timeout          *string       `json:"timeout"`
```
JcBernack, line 52: Timeout is not used

libevm/lib/evm.go
```diff
@@ -235,12 +275,26 @@ func (s *Service) EvmApply(params EvmParams) (error, *EvmResult) {
...
+			if traceErr = json.Unmarshal(traceResultJson, &traceResult); traceErr != nil {
+				return fmt.Errorf("failed to unmarshal trace result %v", traceErr), nil
+			}
+			result.FourByteTracerLogs = traceResult
+		} else if *params.TraceOptions.Tracer == "callTracer" { // callTracer logger
+			var traceResult *CallFrame
+			if traceErr = json.Unmarshal(traceResultJson, &traceResult); traceErr != nil {
+				return fmt.Errorf("failed to unmarshal trace result %v", traceErr), nil
+			}
+			result.CallTracerLogs = traceResult
```
JcBernack, line 296: I think it would make sense to reduce overhead like this by just treating the in and output, i.e. tracer options and the results, as "raw" json.

My suggestion:
- Remove all the types for tracer results and their decoding and encoding. Just pass on the raw json from `tracer.GetResult()` and put it into our response to the RPC call. See `json.RawMessage` in GO and `@JsonRawValue` for the Jackson serializer on how to do that.
- Also remove the decoding and reencoding of tracer options. Just pass them on as raw json to the tracer. All of them have different options, which is why they do json parsing anyway. The only exception is the default StructLogger here.
- This is also how GETH does it to enable each tracer to have custom in and output formats without having to touch the other parts of the code.

libevm/lib/evm.go
```diff
@@ -40,10 +44,31 @@ type EvmContext struct {
...
 type TraceOptions struct {
-	EnableMemory     bool `json:"enableMemory"`
-	DisableStack     bool `json:"disableStack"`
-	DisableStorage   bool `json:"disableStorage"`
-	EnableReturnData bool `json:"enableReturnData"`
+	EnableMemory     bool          `json:"enableMemory"`
+	DisableStack     bool          `json:"disableStack"`
+	DisableStorage   bool          `json:"disableStorage"`
+	EnableReturnData bool          `json:"enableReturnData"`
+	Tracer           *string       `json:"tracer"`
```
JcBernack, line 51: Prefer using just `string` (not a pointer), and check against empty string instead of `nil`.

JcBernack (changes requested):

libevm/lib/evm.go
```diff
@@ -9,11 +9,15 @@ import (
...
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/eth/tracers"
 	"github.com/ethereum/go-ethereum/eth/tracers/logger"
 	"github.com/ethereum/go-ethereum/params"
 	"math"
 	"math/big"
+	"strconv"
 	"time"
+
+	_ "github.com/ethereum/go-ethereum/eth/tracers/native"
```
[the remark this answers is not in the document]
davehorizen: Comment added

libevm/lib/evm.go
davehorizen: Timeout field removed from the Go part

libevm/lib/evm.go
davehorizen: Agree on you point, I proceed to refactor the code as reported.
Now the input `tracerConfig` field and the tracer result response are managed are `json.RawMessage` in the Go part and as generic Object in the Scala/Java part. There was no need to flag them with `@JsonRawValue` annotation as they are correctly serialized as `LinkedHashMap` of `String` and `Object`.
Now we manage all the _geth_ native tracers and all its future changes in the interface (regarding the _geth_ version used) regardless of their interface.

The Go unit tests were modified just checking if the result is not null and that we don't receive any error from the `tracer.getResult()` method call. The checks on the response are now demanded to python tests where we interact with a live session of the application (test python `sc_evm_test_debug_methods.py`)

evm/src/main/java/com/horizen/evm/interop/EvmResult.java
```diff
@@ -3,25 +3,20 @@
...
-    public Map<String,Integer> fourByteTracerLogs;
-    public EvmCallTraceLog callTracerLogs;
     public Boolean reverted;
+    public Object tracerResult;
 
     public static EvmResult emptyEvmResult() {
         var evmResult = new EvmResult();
-
         evmResult.usedGas = BigInteger.ZERO;
         evmResult.returnData = new byte[] {0};
```
JcBernack, line 18: Unrelated to your changes: Why do we have have `new byte[] {0}` here? I assume it should be an empty array, i.e. `new byte[0]`.

evm/src/main/java/com/horizen/evm/interop/EvmResult.java
```diff
@@ -3,25 +3,20 @@
...
 public class EvmResult {
     public BigInteger usedGas;
     public String evmError;
     public byte[] returnData;
     public Address contractAddress;
-    public EvmTraceLog[] traceLogs;
-    public Map<String,Integer> fourByteTracerLogs;
-    public EvmCallTraceLog callTracerLogs;
     public Boolean reverted;
+    public Object tracerResult;
```
JcBernack, line 13: We could also use `JsonNode` here, to clarify that this is unparsed JSON.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -572,27 +570,14 @@ class EthService(
...
-            }
-            fourByteTracerReturnList
-          }
-        }
-        // Struct/opcode logger
-        else {
-          val debugTraceBlockView = new DebugTraceBlockView(evmResults.toArray)
-          debugTraceBlockView.getDebugTraceTransactionViews
+        // return the list of tracer results from the evm
+        val tracerResultList = new ListBuffer[Any]
```
JcBernack, line 574: Here too, would be a bit more clear to return a list of `JsonNode` instead of `Any`.

JcBernack (changes requested):
Nice, so much boilerplate removed! :+1:

libevm/lib/evm.go
JcBernack: Agree on all points, except maybe using `JsonNode` instead of `Object`. See comments below.

libevm/lib/evm_test.go
```diff
@@ -146,11 +127,8 @@ func TestEvmCallTracerWithTracerConfig(t *testing.T) {
 			DisableStack:     false,
 			DisableStorage:   false,
 			EnableReturnData: true,
-			Tracer:           &TracerName,
-			TracerConfig: &TracerConfig{
-				OnlyTopCall: true,
-				WithLog:     true,
-			},
+			Tracer:           TracerName,
+			TracerConfig:     TracerConfig,
```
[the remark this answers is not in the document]
davehorizen: Done

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -633,25 +618,10 @@ class EthService(
...
-          }
-        }
-        // Struct/opcode logger
-        else {
-          if(blockContext.getEvmResult.traceLogs != null)
-            new DebugTraceTransactionView(blockContext.getEvmResult)
-        }
+        // return the tracer result from the evm
+        if(blockContext.getEvmResult.tracerResult != null) {
+          blockContext.getEvmResult.tracerResult
```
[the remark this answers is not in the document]
davehorizen: Agree, this check was needed before because we initialized the `EvmResult` with its custom constructor, but this constructor is now removed as reported in the comment below.
Now I just check the presence of the `EvmResult`, in this way if the transaction is a simple _EOAtoEAO_ that doesn't involve the evm I'm able to return an empty response instead of `null`

evm/src/main/java/com/horizen/evm/interop/EvmResult.java
davehorizen: This constructor was introduced in another PR to resolve some `nullPointerException` in case we call the rpc method for transactions that doesn't involve the evm.
It was an hotfix, I remove it

evm/src/main/java/com/horizen/evm/interop/EvmResult.java
davehorizen: Datatype changed to the less generic `JsonNode`

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Done

libevm/lib/evm.go
```diff
@@ -40,10 +44,25 @@ type EvmContext struct {
...
-	EnableReturnData bool `json:"enableReturnData"`
+	EnableMemory     bool            `json:"enableMemory"`
+	DisableStack     bool            `json:"disableStack"`
+	DisableStorage   bool            `json:"disableStorage"`
+	EnableReturnData bool            `json:"enableReturnData"`
+	Tracer           string          `json:"tracer"`
+	TracerConfig     json.RawMessage `json:"tracerConfig"`
+}
+
+type CallFrame struct {
```
i-Alex, line 55: where do we use this struct?

libevm/lib/evm_test.go
```diff
@@ -47,26 +48,130 @@ func TestEvmTrace(t *testing.T) {
...
+		TraceOptions: &TraceOptions{
+			EnableMemory:     true,
+			DisableStack:     false,
+			DisableStorage:   false,
+			EnableReturnData: true,
+			Tracer:           "callTracer",
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
```
i-Alex, line 90: can we do a sanity check of the result structure/data?
Same in other tests

libevm/lib/evm_test.go
```diff
@@ -85,13 +190,27 @@ func TestEvmOpCodes(t *testing.T) {
...
+	err, resultDeploy := instance.EvmApply(evmParamsTemp)
+	/*
+		err, resultDeploy := instance.EvmApply(EvmParams{
+			HandleParams: HandleParams{Handle: stateHandle},
+			From:         user,
+			To:           nil,
+			Input:        test.OpCodesContractDeploy(),
+			AvailableGas: 200000,
+		})
+	*/
```
i-Alex, line 210: commented code can be removed

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -534,16 +537,18 @@ class EthService(
...
+        Hash.fromBytes(idToBytes(getBlockIdByTag(nodeView, number)))
+      }
+    }
     traceBlockByHash(hash, config)
   }
 
   @RpcMethod("debug_traceBlockByHash")
   @RpcOptionalParameters(1)
-  def traceBlockByHash(hash: Hash, config: TraceOptions): DebugTraceBlockView = {
+  def traceBlockByHash(hash: Hash, config: TraceOptions): Any = {
```
i-Alex, line 551: why result type is `Any`?
I think we can be more accurate here, using `List[JsonNode]`

i-Alex (changes requested):
Looks quite good!
Would like to see a bit more checks in libevm tests (we may discuss the need).

libevm/lib/evm_test.go
davehorizen: Sanity checks added to the Go unit tests, now we check the correct structure of the tracer result and check its values.
I also restore some checks done on the struct logger response present before.
We must take into account that they could fail due to future changes of the native loggers structure, so unrelated to our `libevm` implementation.

libevm/lib/evm_test.go
davehorizen: Removed

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Agree, return datatype changed to `List[JsonNode]`

libevm/lib/evm_test.go
```diff
@@ -125,13 +177,27 @@ func TestEvmCallTracerWithTracerConfig(t *testing.T) {
 			DisableStack:     false,
 			DisableStorage:   false,
 			EnableReturnData: true,
```
JcBernack, line 179: These settings don't do anything when it is not the struct tracer, right? If so, might omit them.

libevm/lib/evm_test.go
```diff
@@ -169,13 +233,18 @@ func TestEvmFourByteTrace(t *testing.T) {
...
```
JcBernack, line 235: same

JcBernack (approved):

libevm/lib/evm_test.go
davehorizen: Yes, they are ignored by the geth implementation in case the tracer is not the default one, removed from the unit test.

libevm/lib/evm_test.go
davehorizen: removed

i-Alex (approved):

PR #703, opened 2023-01-23
feature/certsubmetter_package_refactoring

i-skrypnyk:


i-Alex (changes requested):
Changes seem good.
Check UTs, couple of them are failing because of CircuitStrategy.
Assume that some data was not mocked/defined in the test

i-Alex (approved):

PR #709, opened 2023-01-26
Sc2sc/sdk 644

paolocappelletti:
Sender implementation:
covers most of the sender part implementation, with the following missing parts:
- cryptolib interfaces are still mocked and not tested - and still not used in certificate creation and api endpoints

### DIFF
--- a/sdk/src/main/java/com/horizen/cryptolibprovider/Sc2scCircuit.java
+++ b/sdk/src/main/java/com/horizen/cryptolibprovider/Sc2scCircuit.java
@@ -17,7 +17,7 @@ public interface Sc2scCircuit {
 
     byte[] getCrossChainMessageTreeRoot(List<CrossChainMessage> messages);
 
-    MerklePath gerCrossChainMessageMerklePath(List<CrossChainMessage> messages, CrossChainMessage leaf);
+    MerklePath getCrossChainMessageMerklePath(List<CrossChainMessage> messages, CrossChainMessage leaf);
 
     byte[] createRedeemProof(CrossChainMessageHash messageHash,
                              MerklePath messageMerkePath,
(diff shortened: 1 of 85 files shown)


sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -296,6 +336,25 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+    val total = alreadyMined + boxInThisBlock
+    if (total > allowed) {
+      throw new IllegalStateException(s"Exceeded the maximum number of CrosschainMessages allowed!")
+    }
+  }
+
+  def getAlreadyMinedCrosschainMessagesInCurrentEpoch: Int = {
+    stateStorage.getCrossChainMessagesPerEpoch(getWithdrawalEpochInfo.epoch).size
+  }
+
```
MarcoOl94, line 352: In the UTXO model we can have more CrossChainMessage boxes in the mempool then the ones allowed per epoch. We need to add a rule inside the forger to prevent the creation of blocks that contain more boxes then the max limit per epoch or we can run into the creation of invalid blocks.

MarcoOl94 (changes requested):

sdk/src/main/java/com/horizen/cryptolibprovider/Sc2scCircuit.java
```diff
@@ -0,0 +1,38 @@
...
+
+public interface Sc2scCircuit {
+
+    int getMaxMessagesPerCertificate();
+
+    CrossChainMessageHash getCrossChainMessageHash(CrossChainMessage msg);
+
+    byte[] getCrossChainMessageTreeRoot(List<CrossChainMessage> messages);
+
+    MerklePath gerCrossChainMessageMerklePath(List<CrossChainMessage> messages, CrossChainMessage leaf);
```
paologalligit, line 20: typo: `gerCrossChainMessageMerklePath` instead of `getCrossChainMessageMerklePath`

MarcoOl94 (approved):

PR #711, opened 2023-01-27
As/ethtx rlp codec evm535

alsala:


sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
```diff
@@ -122,7 +120,7 @@ public void ethereumLegacyEIP155TransactionTest() {
         // - to address: 0xd830603264bd3118cf95f1fc623749337342f9e9
         // - value: 3000000000000000000
         // - chain id: 1997
-        String metamaskHexStr = "de01f86d02843b9aca0082520894d830603264bd3118cf95f1fc623749337342f9e98829a2241af62c000080820fbea0f638802002d7c0a3115716f7d24d646452d598050ffc2d6892ba0ed88aeb76bea01dd5a7fb9ea0ec2a414dbb93b09b3e716a3cd05c1e333d40622c63d0c34e3d35";
+        String metamaskHexStr = "f86d02843b9aca0082520894d830603264bd3118cf95f1fc623749337342f9e98829a2241af62c000080820fbea0f638802002d7c0a3115716f7d24d646452d598050ffc2d6892ba0ed88aeb76bea01dd5a7fb9ea0ec2a414dbb93b09b3e716a3cd05c1e333d40622c63d0c34e3d35";
```
i-Alex, line 123: just to double check.
Here we removed the encoded length of the raw tx data, right?

sdk/src/test/java/com/horizen/account/utils/RlpStreamCodecTest.java
```diff
@@ -0,0 +1,239 @@
...
+import scorex.util.serialization.Reader;
+import scorex.util.serialization.VLQByteBufferReader;
+import scorex.util.serialization.VLQByteBufferWriter;
+import scorex.util.serialization.Writer;
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import static org.junit.Assert.*;
+
+public class RlpStreamCodecTest {
```
i-Alex, line 20: Pls, take a look at the original implementation tests, maybe we may reuse the, as well: https://github.com/web3j/web3j/blob/master/rlp/src/test/java/org/web3j/rlp/RlpDecoderTest.java
https://github.com/web3j/web3j/blob/master/rlp/src/test/java/org/web3j/rlp/RlpEncoderTest.java

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -486,4 +487,8 @@ public Message asMessage(BigInteger baseFee) {
     public byte[] encode(boolean accountSignature) {
         return EthereumTransactionEncoder.encodeAsRlpValues(this, accountSignature);
     }
```
i-Alex, line 489: At the moment methods `encode(boolean accountSignature)` and `encode(boolean accountSignature, Writer writer)` have 2 independent implementations.
Could be better to make one backed by another. What do you think?

i-Alex (commented):
Looks good to me!
Couple of comments to discuss

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionTest.java
alsala: Yes correct

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
alsala: I see the point,  I would  prefer to leave both of them since the rlp encoding with byte array is a little faster. Moreover they share some code for getting `EthereumTransaction` tx obj from `RlpList` and vice versa.

sdk/src/test/java/com/horizen/account/utils/RlpStreamCodecTest.java
alsala: Ok, done

i-Alex (approved):

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
i-Alex: ok, fine

PR #713, opened 2023-01-27
fix gas tests

JcBernack:
- fixed `sc_evm_closed_forger.py` by increasing default gas limits in the http api
- fixed `NullPointerException` in  `StateDbAccountStateViewGasTracked.removeAccountStorage`
- updated gas usage in `ForgerStakeMsgProcessorTest`
- added gas usage verification to `WithdrawalMsgProcessorTest`
- updated EOA and EVM message processor tests to not use `applyMessage` but the processors `process` method directly

Tests of gas usage for the ForgerStakesMsgProcessor were using wrong numbers, because they did not reset the access list between multiple calls ("warm" account access uses less gas, etc.).


JcBernack:
Updated the message processor tests as discussed.

I skipped the key rotation processor test for now because it has slightly random gas usage even without the intrinsic gas. The reason is less gas usage if a storage value is written to all-zeroes, which can happen sometimes because of random data generated in the test.

alsala (approved):

i-Alex (approved):

PR #714, opened 2023-01-27
Ak/block tag changes

akrckn:


akrckn:
For pendingBlock and -State I used some duplicate code, which could be simplified if we do some changes to the AccountForgeMessageBuilder, AccountState and SidechainState, but I was not sure if we want that.
If so, I'll change it.

JcBernack (changes requested):
I vote in favor of refactoring to reduce code duplication.

i-Alex (changes requested):
The overall code looks good.
I left consensus specific suggestion.
Beside that we need to add a python test to verify that everything work as expected.
I want to see the following:
1) Configuration: 1 MC node, 2 SC nodes. sc_node0 -> forger and connected to MC node via websocket. sc_node1 -> not a forger, not connected to MC.
2) test all the RPC methods that may take `Tag` as an argument on both SC nodes.
try with "earliest", "finalized", "safe", "latest", "pending"
Note 1: the results on both sc nodes expected to be the same
Note 2: check that we have proper errors when we use  "finalized", "safe" beofre reaching the proper sc height.
Note 3: during the test of mempool related methods be sure that we have some txs in the mempool and that this Txs were accounted in the "pending" block.

RPC methods to test:
1) eth_getBlockByNumber
2) eth_getBlockTransactionCountByNumber
3) eth_call
4) eth_estimateGas (update its own estimate gas py test)
5) eth_getBalance
6) eth_getTransactionCount

sdk/src/main/scala/com/horizen/account/api/rpc/service/utils/PendingBlock.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import java.util.{ArrayList => JArrayList}
+import scala.concurrent.duration.SECONDS
+import scala.util.{Failure, Success}
+
+case class BranchPointInfo(
+    branchPointId: ModifierId,
+    referenceDataToInclude: Seq[MainchainHeaderHash],
+    headersToInclude: Seq[MainchainHeaderHash]
+)
```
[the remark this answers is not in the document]
akrckn: Yes, I oversaw it while refactoring

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -516,38 +521,110 @@ class EthService(
...
+      parentId,
+      System.currentTimeMillis / 1000,
+      ModifierSemanticValidity.Unknown,
+      null,
+      null,
+      WithdrawalEpochUtils.getWithdrawalEpochInfo(
+        0,
+        parentInfo.withdrawalEpochInfo,
+        networkParams
+      ),
```
i-Alex, line 544: can be replaced with `parentInfo.withdrawalEpochInfo`, since the result of this method will not increase the parent value

qa/sc_evm_rpc_block_tags.py
```diff
@@ -0,0 +1,192 @@
...
+                                         maxFeePerGas=900000000, value=1))
+            nonce_addr_1 += 1
+
+        # Check pending state and block functions
+        expResp = {'Balance': '0x1b1ae49220f02adffc', 'TransactionCount': '0x4', 'BlockTransactionCount': '0x4',
+                   'Block': '0x3'}
+        self.__send_and_assert_tag_rpc_methods('pending', add_0x_prefix(evm_address_sc2), **expResp)
+
+        # Create tracer result for pending block after smart contract deployment was added to mempool
+        traceBlock_pending = sc_node_1.rpc_debug_traceBlockByNumber('pending')['result']
```
i-Alex, line 170: let's check that pending block contains txs from the list `common_tx_list`
Currently `common_tx_list` is never used, we check only the number of txs in the block, but not their ids.

qa/sc_evm_rpc_block_tags.py
```diff
@@ -0,0 +1,192 @@
...
+        common_tx_list = []
+        for i in range(4):
+            common_tx_list.append(
+                createEIP1559Transaction(sc_node_1, fromAddress=evm_address_sc1, toAddress=evm_address_sc2,
+                                         nonce=nonce_addr_1, gasLimit=230000, maxPriorityFeePerGas=900000000,
+                                         maxFeePerGas=900000000, value=1))
+            nonce_addr_1 += 1
+
+        # Check pending state and block functions
+        expResp = {'Balance': '0x1b1ae49220f02adffc', 'TransactionCount': '0x4', 'BlockTransactionCount': '0x4',
```
i-Alex, line 165: please add the comment explaining the value `0x1b1ae49220f02adffc`
what is it decimal form
and why do we expect exactly this value.

i-Alex (changes requested):
The changes looks good to me.
I left a couple of minor non-critical suggestions.
Py test is amazing!

qa/sc_evm_rpc_block_tags.py
akrckn: Implemented it - did not do it before, because we already check this while doing some contract calls.

qa/sc_evm_rpc_block_tags.py
akrckn: I refactored this a bit to be a lot more generic and understandable

i-Alex (approved):

JcBernack (approved):
Nice! :+1:

PR #716, opened 2023-02-01
Address and Hash type refactoring

JcBernack:
- [EVM-419](https://horizenlabs.atlassian.net/browse/EVM-419)

Todo
- [x] fix issue with mocked `AccountStateView.getProof` (this currently breaks a number of tests)

[EVM-419]: https://horizenlabs.atlassian.net/browse/EVM-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdk/src/main/java/com/horizen/account/api/rpc/types/TransactionArgs.java
```diff
@@ -157,8 +157,8 @@ public Message toMessage(BigInteger baseFee, BigInteger rpcGasCap) throws RpcExc
             }
         }
         return new Message(
-            from == null ? Optional.empty() : Optional.of(new AddressProposition(from.toBytes())),
-            to == null ? Optional.empty() : Optional.of(new AddressProposition(to.toBytes())),
+            getFrom(),
```
i-Alex, line 160: In prev implementation we returned `Optional.empty()` in case `from` is null.
Currently, we use `getFrom()` that return `Address.ZERO` instead of `null`
Note: related to comment below.

sdk/src/main/java/com/horizen/account/state/Message.java
```diff
@@ -24,16 +24,16 @@ public class Message {
...
-            Optional<AddressProposition> to,
-            BigInteger gasPrice,
-            BigInteger gasFeeCap,
-            BigInteger gasTipCap,
-            BigInteger gasLimit,
-            BigInteger value,
-            BigInteger nonce,
-            byte[] data,
-            boolean fakeMsg
+        Address from,
```
i-Alex, line 27: let's discuss the type of from:
`Address` or `Optional<Address>` ?
If I remember that empty `from` can be a case that we need to care about. It is negative flow for sure.
@alsala if you remember, pls join.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -385,6 +386,7 @@ public synchronized AddressProposition getFrom() {
                         )
                 );
             } catch (Exception e) {
+                // TODO: is this a valid case? what are supposed to do with a TX that has a signature we cannot process?
```
i-Alex, line 389: this may happen, so we should reject this Tx later as invalid.
That was also a reason why we kept Message.from field as `Optional`

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
```diff
@@ -75,7 +76,7 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
     val secret = allAccounts.find(
       a => (fromAddress.isEmpty ||
         BytesUtils.toHexString(a.asInstanceOf[PrivateKeySecp256k1].publicImage
-          .address) == fromAddress.get) &&
+          .address.toBytes) == fromAddress.get) &&
```
i-Alex, line 79: check this line: here we compare `byte[]` with `string`

sdk/src/main/scala/com/horizen/account/receipt/EthereumReceipt.scala
```diff
@@ -130,7 +113,9 @@ object EthereumReceiptSerializer extends SparkzSerializer[EthereumReceipt] {
 
     // optional field
     val contractAddressLength = reader.getInt
-    val contractAddress: Option[Array[Byte]] = if(contractAddressLength == 0) None else Some(reader.getBytes(contractAddressLength))
+    val contractAddressBytes = reader.getBytes(contractAddressLength)
```
i-Alex, line 116: honestly it is a bit misleading to read 0 bytes.
let's reformat to:
`if (contractAddressLength == 0) None else Some(new Address(reader.getBytes(contractAddressLength)))`

i-Alex (changes requested):
Amazing work!
couple minor points + let's discuss Message class "from" field.

sdk/src/main/java/com/horizen/account/state/Message.java
JcBernack: I think this case is only valid for Transactions, not for Messages.

When we create a Message we need to know what the from address will be - it cannot be processed without it. Currently this was very misleading, because if `msg.from` was `None` it would often be treated as `null`, which turns into the Zero-Address when passed to libevm. See the original implementation of `msg.getFromAddressBytes`. One part of the refactoring is to make this behavior explicit.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
JcBernack: I think a transaction is rejected based on this before it is turned into a message, also see the other comment.

JcBernack: So basically this case is "Invalid Signature"?

sdk/src/main/java/com/horizen/account/state/Message.java
i-Alex: Ok, seems clear to me.
Let's double check the place where we transform Tx to Message. In case signature check occurs ALWAYS (by design) before, then it seems fine that `Message.from` is of type `Address`.
But in case signature validation is done after `Message` obj construction it seems more natural to fail it because of `None`.
PS: I mean the case when the received/constructed Tx miss signature at all.

sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
i-Alex: yes: signature is present, but for some reason "from" address was not derived from it

sdk/src/main/scala/com/horizen/account/api/http/AccountTransactionApiRoute.scala
JcBernack: no, it is `string == string`

sdk/src/main/java/com/horizen/account/state/Message.java
JcBernack: The signature check is included in `EthereumTransaction.semanticValidity()`. We go from Tx to Message in `StateDbAccountStateView.applyTransaction`, just before we call `applyMessage`, so I think we're good.

There were some checks on the `from` address in the `ForgerStakeMsgProcessor` which I removed because they made no sense to me. Without these it is now possible to create a forger stake from funds at the zero-address - but only within a non-persistent  `eth_call` of course. Things like these are also possible on Ethereum. If you omit `from` during `eth_call` or `eth_estimateGas` it will default to the zero-address.

i-Alex: has sense!

PR #717, opened 2023-02-01
As/det key gen evm655

alsala:


sdk/src/test/java/com/horizen/account/utils/ChaChaPrngSecureRandomTest.java
```diff
@@ -0,0 +1,67 @@
+package com.horizen.account.utils;
```
i-Alex, line 1: move this test to  `package com.horizen.utils` as well

alsala: Done

i-Alex (approved):

PR #718, opened 2023-02-02
EVM-690 - debug_traceCall rpc method

davehorizen:


i-Alex (approved):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -624,6 +630,52 @@ class EthService(
...
+
+        // declare the ethereum transaction from the input parameters
+        val ethTx = tx.toTransaction(networkParams)
+        val ethTxScat = ethTx.asInstanceOf[SidechainTypes#SCAT]
+
+        // use default trace params if none are given
+        blockContext.setTraceParams(if (config == null) new TraceOptions() else config)
+
+        // apply requested transaction with tracing enabled
+        tagStateView.applyTransaction(ethTxScat, block.transactions.length, gasPool, blockContext, signatureRequired = false)
```
JcBernack, line 669: I would like to change this to the same method we use for `eth_call`:
```
val msg = params.toMessage(blockContext.baseFee, settings.globalRpcGasCap)
tagStateView.applyMessage(msg, new GasPool(msg.getGasLimit), blockContext)
```
Let's not make it look like we have a transaction here, because we don't.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -624,6 +630,52 @@ class EthService(
     }
   }
 
+  @RpcMethod("debug_traceCall")
+  @RpcOptionalParameters(1)
+  def traceCall(tx: TransactionArgs, tag: String, config: TraceOptions): Any = {
```
JcBernack, line 635: This function also accepts "tag", we need to make sure this also works with the new "pending" block.

Nit pick: I would prefer a different variable name than `tx`, because it is _not_ a transaction.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -624,6 +630,52 @@ class EthService(
...
+
+    applyOnAccountView { nodeView =>
+      // retrieve block hash from input
+      val hash = getBlockHashByTag(nodeView, tag)
+
+      // get block to trace
+      val (block, blockInfo) = getBlockById(nodeView, bytesToId(hash.toBytes))
+
+      // get state at previous block
+      getStateViewAtTag(nodeView, (blockInfo.height - 1).toString) { (tagStateView, blockContext) =>
```
JcBernack, line 645: Why do we need to go to the parent block here?

sdk/src/main/scala/com/horizen/account/state/StateDbAccountStateView.scala
```diff
@@ -179,18 +179,20 @@ class StateDbAccountStateView(stateDb: StateDB, messageProcessors: Seq[MessagePr
...
     val ethTx = tx.asInstanceOf[EthereumTransaction]
 
-    // should never happen if the tx has been accepted in mempool.
+    // The signature check is skipped for some specific cases (condition checked with signatureRequired flag)
+    // but it should never happen if the tx has been accepted in mempool.
     // In some negative test scenario this can happen when forcing an unsigned tx to be forged in a block.
     // In this case the 'from' attribute in the msg would not be 
     // set, and it would be difficult to rootcause the reason why gas and nonce checks would fail
-    if (!ethTx.isSigned)
+    if (signatureRequired && !ethTx.isSigned)
```
JcBernack, line 195: We don't need this with a bit of refactoring, see other comments.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: Davide, pls keep in mind this PR: https://github.com/HorizenOfficial/Sidechains-SDK/pull/714
has been merged to dev_evm this morning

sdk/src/main/scala/com/horizen/account/state/StateDbAccountStateView.scala
davehorizen: `signatureRequired` flag removed from method signature and logic

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: typo, fixed

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Rebase with the changes introduced with the PR https://github.com/HorizenOfficial/Sidechains-SDK/pull/714
Python test updated with _traceCall_ method call using the _pending_ tag

qa/sc_evm_test_debug_methods.py
```diff
@@ -19,11 +18,13 @@
...
 Test:
-    For the smart contract:
         - Deploy a smart contract without initial data
         - Transfer tokens and trace transaction
         - EOA2EOA transfer and trace transaction
-        - Trace block with transaction
+        - Trace block with transaction using the debug_traceTransaction and debug_traceBlockByNumber/debug_traceBlockByHash methods
+        - Call traceCall method method on current block
+        - Generate a new block without transaction and call tracer methods
+        - Call traceCall method method on pending block
```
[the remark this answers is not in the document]
davehorizen: typo fixed, thanks!

i-Alex (approved):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -786,6 +794,44 @@ class EthService(
...
+        for (mcBlockRefData <- block.mainchainBlockReferencesData) {
+          tagStateView.applyMainchainBlockReferenceData(mcBlockRefData)
+        }
+
+        val gasPool = new GasPool(block.header.gasLimit)
+
+        // apply block transactions without tracing
+        for ((tx, i) <- block.transactions.zipWithIndex) {
+          tagStateView.applyTransaction(tx, i, gasPool, blockContext)
+        }
```
JcBernack, line 829: Previously you we're opening the state after the parent block and then reapplying the transactions of this block. Now you open the state at the desired block - so that already includes the transactions of the block.
I think we can remove this - it should fail anyway because the nonces are not valid anymore.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -786,6 +794,44 @@ class EthService(
...
+        if(blockContext.getEvmResult != null) {
+          blockContext.getEvmResult.tracerResult
+        } else Unit
+      }
+    }
+  }
+
+  @RpcMethod("debug_traceCall")
+  @RpcOptionalParameters(1)
+  def traceCall(params: TransactionArgs, tag: String, config: TraceOptions): Any = {
```
JcBernack, line 807: The expected behavior of `debug_traceCall` is inconsistent with all the other functions accepting blocknumber/blockhash/tag, right?
Everwhere else there is `getSomethingByBlockNumber` which accepts block number or tag and `getSomethingByBlockHash` which only accepts a block hash. Here it is all in one function.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -786,6 +794,44 @@ class EthService(
...
+
+    applyOnAccountView { nodeView =>
+      // get block to trace
+      val (block, blockInfo) = getBlockById(nodeView, getBlockIdByHashOrTag(nodeView, tag))
+
+      // get state at selected block
+      getStateViewAtTag(nodeView, if(tag=="pending") "pending" else (blockInfo.height).toString) { (tagStateView, blockContext) =>
+
+        // use default trace params if none are given
+        blockContext.setTraceParams(if (config == null) new TraceOptions() else config)
```
JcBernack, line 817: Enabling tracing here will also enable it for the transactions of the block - but probably doesn't matter, see the other comment.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -474,6 +474,14 @@ class EthService(
     blockId
   }
 
+  private def getBlockIdByHashOrTag(nodeView: NV, tag: String): ModifierId = {
+    if (tag.length == 66 && tag.substring(0, 2) == "0x")
+      bytesToId(Converter.fromHexString(tag.substring(2)))
```
JcBernack, line 479: Please use any of the other conversion tools, e.g. `ByteUtils.fromHexString`. `Converter` is supposed to be internal to the `evm` package.

qa/sc_evm_test_debug_methods.py
```diff
@@ -156,6 +157,63 @@ def run_test(self):
...
+                                             "enableReturnData": False
+                                         }]})
+        res = sc_node.ethv1(request)['result']
+        assert_true("error" not in res, "debug_traceCall failed for successful smart contract transaction")
+        trace_logs_length = len(res['structLogs'])
+        assert_true(trace_logs_length > 130, "unexpected number of trace logs, less than 130")
+
+        # call tracer - native tracer
+        request = json.dumps({"jsonrpc": "2.0", "method": "debug_traceCall", "id": 48, "params": [trace_call_transaction, block_number, {"tracer": "callTracer"}]})
+        res = sc_node.ethv1(request)['result']
```
JcBernack, line 188: Please use the existing mechanism for making RPC calls:
```suggestion
        res = sc_node.rpc_debug_traceCall(trace_call_transaction, block_number, {"tracer": "callTracer"})
```

JcBernack: Any function call on a node starting with `rpc_` will automatically translate into an RPC call.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Agree, the previous implementation was not correct. Removed

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: yes, removing the code block reported in the comment above the trace params will be applied only to the message create with the input parameters.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Fixed

qa/sc_evm_test_debug_methods.py
davehorizen: Call syntax changed to the one proposed for this and the others `debug_traceCall` method calls

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: According to the _geth_ documentation the block can be specified either by hash or by number as the second argument (https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall). Moreover, the `geth` current implementation seems to manage both block number or block hash (https://github.com/ethereum/go-ethereum/blob/master/eth/tracers/api.go#L895). 
This led to our current implementation that need to some modifications to interact with methods that manage by default a block by its number or tag.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -786,6 +794,32 @@ class EthService(
...
+    }
+  }
+
+  @RpcMethod("debug_traceCall")
+  @RpcOptionalParameters(1)
+  def traceCall(params: TransactionArgs, tag: String, config: TraceOptions): Any = {
+
+    applyOnAccountView { nodeView =>
+      // get block to trace
+      val (block, blockInfo) = getBlockById(nodeView, getBlockIdByHashOrTag(nodeView, tag))
```
JcBernack, line 811: One last note:
`block` is unused now and `blockInfo` is only used to get the block height, which is just used to lookup the block hash again :sweat_smile: 
There is potential to refactor here, but could also do that later.

JcBernack (approved):
Good :+1:

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
davehorizen: Did a small refactor adding and calling a method that retrieve only the `blockInfo` by _id_ and not the entire `block`

PR #719, opened 2023-02-02
As/rlp enc check evm583

alsala:


sdk/src/main/java/com/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -236,7 +236,8 @@ public void semanticValidity() throws TransactionSemanticValidityException {
             throw new TransactionSemanticValidityException(String.format("Transaction [%s] is semantically invalid: " +
                     "non-positive gas limit", id()));
         if (!BigIntegerUtil.isUint64(getGasLimit()))
-            throw new GasUintOverflowException();
```
i-Alex, line 239: i'm curios why before the exception type was `GasUintOverflowException`?
Has it any impact on state validation? For example, it led to different result in terms of gas consumption in case of such an error?

alsala: No, I just changed the exception type for having uniform error handling in this method

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
```diff
@@ -164,9 +150,225 @@ public void rlpEthTxWithTrimmedBytes() {
...
+        VLQByteBufferWriter writer = new VLQByteBufferWriter(new ByteArrayBuilder());
+        ethTx.encode(true, writer);
+        byte[] result = writer.toBytes();
+        assertEquals(BytesUtils.toHexString(b), BytesUtils.toHexString(result));
+    }
+
+    // useful when developing tests
+    @Test
+    @Ignore
+    public void checkSingleStream() throws TransactionSemanticValidityException {
```
i-Alex, line 360: Do we still need this test?

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
```diff
@@ -164,9 +150,225 @@ public void rlpEthTxWithTrimmedBytes() {
...
+        Reader reader = new VLQByteBufferReader(ByteBuffer.wrap(b));
+        EthereumTransaction ethTx = EthereumTransactionDecoder.decode(reader);
+        ethTx.semanticValidity();
+        System.out.println(ethTx);
+    }
+
     @Test
     @Ignore
-    public void checkPerfStream() {
+    public void checkPerf() {
```
i-Alex, line 371: I'm not sure if comparison has sense now,
the difference between 2 approaches is that the first one also get bytes from the write stream.
Maybe some changes to the test notes/prints needed. Or even can be removed fully 
What do you think?

i-Alex (changes requested):
Looks very good!

sdk/src/main/java/com/horizen/account/utils/EthereumTransactionDecoder.java
```diff
@@ -74,99 +81,126 @@ private static EthereumTransaction decodeEIP1559Transaction(Reader reader) {
...
+            if (r.length != 32)
+                throw new IllegalArgumentException("r byte array length: " + r.length + " != 32");
+            if (s.length != 32)
+                throw new IllegalArgumentException("s byte array length: " + s.length + " != 32");
+
+            chainId = decodeEip155ChainId(v);
+            realSignature = new SignatureSecp256k1(getRealV(v), r, s);
+        }
+
+        if (chainId != null && chainId != 0) {
```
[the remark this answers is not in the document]
alsala: Ok, I modified the  `decodeEip155ChainId()` so that just one condition is enough to be checked

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
alsala: Removed

sdk/src/test/java/com/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
alsala: Ok, I removed this test too

i-Alex (approved):

PR #723, opened 2023-02-07
add big int bit length check to deserialization

akrckn:


akrckn:
FYI - I tested with a small scala worksheet for the byte size of uint256 as ByteArray and bit length of BigInteger - have a look below.

The implementation can be seen as a proposal, we could also consider not supporting the max value of uint256, as `uint256_max_value - 1 = 32 bytes` and we could check for that instead.

```
import java.math.BigInteger

for (i <- (256 to 264)) {
  val bigIntTest = BigInteger.ONE.shiftLeft(i).subtract(BigInteger.ONE)
  val bigIntByteArray = bigIntTest.toByteArray
  if (bigIntByteArray.length > 33)
    println(s"Byte length of ${bigIntByteArray.length} for value $bigIntTest is above byte limit of 33")
  else if (bigIntByteArray.length == 33 && bigIntTest.bitLength() > 256)
    println(s"Bit length of ${bigIntTest.bitLength()} for value $bigIntTest is above bit limit for uint256")
  else
    println(s"valid big integer value of $bigIntTest with bit size of ${bigIntTest.bitLength()}")
}
```

sdk/src/main/scala/com/horizen/account/state/ForgerStakeMsgProcessorData.scala
```diff
@@ -277,8 +278,12 @@ object ForgerStakeDataSerializer extends SparkzSerializer[ForgerStakeData] {
   override def parse(r: Reader): ForgerStakeData = {
     val forgerPublicKeys = ForgerPublicKeysSerializer.parse(r)
     val ownerPublicKey = AddressPropositionSerializer.getSerializer.parse(r)
+
     val stakeAmountLength = r.getInt()
     val stakeAmount = new BigInteger(r.getBytes(stakeAmountLength))
+    val bigIntBitLength = stakeAmount.bitLength()
+    if (bigIntBitLength > Account.BIG_INT_MAX_BIT_SIZE)
+      throw new IllegalArgumentException(s"Base Fee bit size $bigIntBitLength exceeds the limit ${Account.BIG_INT_MAX_BIT_SIZE}")
```
i-Alex, line 286: "Base Fee" -> "Stake amount"

sdk/src/main/scala/com/horizen/account/block/AccountBlockHeader.scala
```diff
@@ -210,14 +211,25 @@ object AccountBlockHeaderSerializer extends SparkzSerializer[AccountBlockHeader]
...
     val baseFee = new BigInteger(r.getBytes(baseFeeSize))
+    bigIntBitLength = baseFee.bitLength()
+    if (bigIntBitLength > Account.BIG_INT_MAX_BIT_SIZE)
+      throw new IllegalArgumentException(s"Base Fee bit size $bigIntBitLength exceeds the limit ${Account.BIG_INT_MAX_BIT_SIZE}")
 
     val gasUsedSize = r.getInt()
     val gasUsed = new BigInteger(r.getBytes(gasUsedSize))
+    bigIntBitLength = baseFee.bitLength()
+    if (bigIntBitLength > Account.BIG_INT_MAX_BIT_SIZE)
+      throw new IllegalArgumentException(s"Base Fee bit size $bigIntBitLength exceeds the limit ${Account.BIG_INT_MAX_BIT_SIZE}")
```
i-Alex, line 226: "base fee" -> "Gas used"

sdk/src/main/scala/com/horizen/account/block/AccountBlockHeader.scala
```diff
@@ -210,14 +211,25 @@ object AccountBlockHeaderSerializer extends SparkzSerializer[AccountBlockHeader]
...
 
     val gasLimitSize = r.getInt()
     val gasLimit = new BigInteger(r.getBytes(gasLimitSize))
+    bigIntBitLength = gasLimit.bitLength()
+    if (bigIntBitLength > Account.BIG_INT_MAX_BIT_SIZE)
+      throw new IllegalArgumentException(s"Base Fee bit size $bigIntBitLength exceeds the limit ${Account.BIG_INT_MAX_BIT_SIZE}")
```
i-Alex, line 232: "base fee" -> "Gas limit"

sdk/src/main/java/com/horizen/account/utils/EthereumTransactionDecoder.java
```diff
@@ -83,12 +83,35 @@ private static EthereumTransaction decodeEIP1559Transaction(byte[] transaction)
     private static EthereumTransaction RlpList2EIP1559Transaction(RlpList rlpList) {
         RlpList values = (RlpList)rlpList.getValues().get(0);
         long chainId = ((RlpString)values.getValues().get(0)).asPositiveBigInteger().longValueExact();
+
+        Integer bigIntBitLength;
         BigInteger nonce = ((RlpString)values.getValues().get(1)).asPositiveBigInteger();
+        bigIntBitLength = nonce.bitLength();
+        if (bigIntBitLength > Account.BIG_INT_MAX_BIT_SIZE)
+            throw new IllegalArgumentException(String.format("Base Fee bit size %d exceeds the limit %d", bigIntBitLength, Account.BIG_INT_MAX_BIT_SIZE));
```
i-Alex, line 91: "Base Fee" -> "Nonce"
Check messages in all other places

sdk/src/main/scala/com/horizen/account/utils/AccountBlockFeeInfo.scala
```diff
@@ -27,10 +27,18 @@ object AccountBlockFeeInfoSerializer extends SparkzSerializer[AccountBlockFeeInf
   }
 
   override def parse(r: Reader): AccountBlockFeeInfo = {
+    var bigIntBitLength: Integer = 0
```
i-Alex, line 30: here we may remove the check because it is an internal object, so node creates and holds it by itself.
Let's keep the check only in the objects that can be retreived from outside.

i-Alex (changes requested):
Overall direction looks good.
Let's merge dev_evm and resolve conflicts

i-Alex (approved):

PR #731, opened 2023-02-10
Account block size check

alsala:


sdk/src/main/resources/sidechain-sdk-settings.conf
```diff
@@ -19,6 +19,6 @@ sparkz {
     bindAddress = "127.0.0.1:9084"
     knownPeers = []
     agentName = "2-Hop"
-    maxPacketSize = 5242980
+    maxModifiersSpecMessageSize = 5242980
```
i-Alex, line 22: I think, it must be 7mb as for EVM.
to be discussed

sdk/src/main/scala/com/horizen/account/block/AccountBlock.scala
```diff
@@ -88,15 +88,22 @@ class AccountBlock(override val header: AccountBlockHeader,
...
+  override def blockExceedsOverheadSizeLimit(blockOverheadSize: Long): Boolean = {
+    blockOverheadSize > MAX_ACCOUNT_BLOCK_OVERHEAD_SIZE
+  }
 }
 
 
 object AccountBlock {
 
   val ACCOUNT_BLOCK_VERSION: Block.Version = 2: Byte
+  val MAX_ACCOUNT_BLOCK_OVERHEAD_SIZE: Int = 5000000
```
i-Alex, line 105: let's put also a comment here and describe what "overhead" means

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
```diff
@@ -135,7 +138,7 @@ abstract class SidechainBlockBase[TX <: Transaction, H <: SidechainBlockHeaderBa
       throw new InvalidSidechainBlockDataException(s"${getClass.getSimpleName} $id sidechain transactions amount exceeds the limit.")
 
     // Check Block size
-    val blockSize: Int = bytes.length
+    val blockSize: Long = bytes.length
```
i-Alex, line 141: why do we use `Long`.
In most of other places we use `Int`, for example: `def getMaxBlockOverheadSize() : Int`

sdk/src/main/resources/sidechain-sdk-settings.conf
alsala: Ok, set the fallback conf to 7MB

sdk/src/main/scala/com/horizen/account/block/AccountBlock.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
alsala: `EthereumTransaction size()` method returns a `Long`

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
```diff
@@ -146,17 +149,25 @@ abstract class SidechainBlockBase[TX <: Transaction, H <: SidechainBlockHeaderBa
...
+        totalTxSize = totalTxSize + tx.size()
       } match {
         case Success(_) =>
         case Failure(e) => throw new InvalidSidechainBlockDataException(
           s"${getClass.getSimpleName} $id Transaction ${tx.id} is semantically invalid: ${e.getMessage}.")
       }
     }
 
+    // Check we do not exceed the block overhead size
+    val blockOverheadSize = blockSize-totalTxSize
```
i-Alex, line 166: here we consider totalTxSize as a sum of tx bytes, but we do not consider `TxCompanion` and `List`  overhead for the Txs.
In UTXO forger we have rough estimation:
`val txSize = tx.bytes.length + 4 // placeholder for Tx length`
But here we need to be more precise (so may get lower value):
let's introduce another method like: `protected blockTxsSize(): Long`
that on Account and UTXO block will use `companion` to serialize the list of sidechain transactions.

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -108,50 +111,60 @@ class AccountForgeMessageBuilder(
...
-          case Failure(e) =>
-            // skip all txs from the same account but remove any changes caused by the rejected tx
-            log.warn(s"Could not forge tx, reason: ${e.getMessage}", e)
-            priceAndNonceIter.removeAndSkipAccount()
-            stateView.revertToSnapshot(revisionId)
-            // Restore gas
-            val usedGas = initialBlockGas.subtract(blockGasPool.getGas)
-            blockGasPool.addGas(usedGas)
+      val tx = priceAndNonceIter.peek
+      val txSize = tx.size()
```
i-Alex, line 115: let's act as in UTXO forger:
`val txSize = tx.bytes.length + 4 // placeholder for Tx length`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
alsala: Ok, done

sdk/src/main/scala/com/horizen/account/forger/AccountForgeMessageBuilder.scala
alsala: Ok, added 1 byte for companion encoding contribution

i-Alex (approved):

PR #732, opened 2023-02-10
DIFFICULTY/PREVRANDAO opcode support

i-Alex:


saratnt (approved):

libevm/lib/evm.go
```diff
@@ -38,7 +38,7 @@ type EvmContext struct {
 	BlockNumber *hexutil.Big   `json:"blockNumber"`
 	Time        *hexutil.Big   `json:"time"`
 	BaseFee     *hexutil.Big   `json:"baseFee"`
-	Random      *common.Hash   `json:"random"`
+	Random      common.Hash    `json:"random"`
```
JcBernack, line 41: This does not really change anything. Just that the default value will be the zero-hash instead of a `nil` pointer (in case "random" is omitted in the incoming JSON).  If that is what you were trying to do: good - if not, what were you trying to do? :wink:

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -562,7 +562,7 @@ class EthService(
       null,
       null,
       parentInfo.withdrawalEpochInfo,
-      None,
+      parentInfo.vrfOutputOpt, // same as for parent block, used as a source of BlockContext.random field
```
JcBernack, line 565: Did you verify this with Ethereum? `Hash.ZERO` would also make sense I think, as the real data is unknown at this point.

sdk/src/main/scala/com/horizen/consensus/ForgingStakeInfo.scala
```diff
@@ -1,3 +1,4 @@
+
```
JcBernack, line 1: Accidental new line?

qa/sc_evm_test_prevrandao.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Generate some more SC blocks to be able to interact with smart contract for  "safe" tag
+        safe_block_id = generate_next_block(sc_node, "first node")
+        generate_next_blocks(sc_node, "first node", 100)
+
+        # Get best block info and expected random value
+        best_block = sc_node.block_best()["result"]
+        best_block_height = best_block["height"]
+        best_block_vrf_output = best_block["block"]["header"]["vrfOutput"]["bytes"]
+        expected_random = int('0x' + best_block_vrf_output, 16)
```
JcBernack, line 109: This surely must overflow the `int`..?

JcBernack (approved):
Looks good :+1:

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: in Ethereum `pending` tag is forbidden for the places where we want to use it.

libevm/lib/evm.go
i-Alex: I expect now that `random` must be always passed from outside, so `nil` is not a case.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: after discussion: ok to be in this way, we need some real random value, not Zeros

qa/sc_evm_test_prevrandao.py
i-Alex: actually works

PR #739, opened 2023-02-16
RPC refactor

JcBernack:
- [EVM-421](https://horizenlabs.atlassian.net/browse/EVM-421)
- fix RPC error responses: a lot of "bad request" type of errors were masked and would return an "internal error"
- introduce RPC specific serialization: `EthJsonMapper`
  - use it for all RPC requests and responses
  - refactor all RPC response "view" types to remove manual serialization
- refactor and cleanup of `AccountEthRpcRouteTest` and `EthServiceTest`
- change `zen_getForwardTransfers` to only accept block hashes, not block numbers or tags (latest, earliest, etc.)

Note: Previously, the RPC api would also accept decimal representations for "Quantity" values, which was non-standard and was removed. This surfaced some formatting issues in the python tests, because these were unintentionally using decimal values. Instead of changing everything to hex in the python tests, an automatic int-to-hex conversion was retrofitted to the serialization of RPC requests.

[EVM-421]: https://horizenlabs.atlassian.net/browse/EVM-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -7,68 +7,68 @@
...
+    public final BigInteger gas;
+    public final BigInteger value;
+    public final byte[] input;
+    public final BigInteger maxPriorityFeePerGas;
+    public final BigInteger maxFeePerGas;
+    public final BigInteger gasPrice;
+    // as we don't support transactions with accessList we don't include it here
+    // note: accessList should be either the value used in the TX or empty-array []
+    // if the transaction type does not support it, it should not be present at all - don't serialize as "null"
+//    public final ArrayList<Object> accessList;
```
i-Alex, line 33: Does it mean that we need to skip it for `legacy` tx type, but keep it as `[]` for eip-1559? because eip1559 actually supports this field (but in our case it is always empty)

sdk/src/main/java/com/horizen/account/serialization/EthJsonMapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+    static {
+        var module = new SimpleModule();
+        module.addSerializer(BigInteger.class, new BigIntegerSerializer());
+        module.addDeserializer(BigInteger.class, new BigIntegerDeserializer());
+        module.addSerializer(byte[].class, new EthByteSerializer());
+        module.addDeserializer(byte[].class, new EthByteDeserializer());
+        mapper = new ObjectMapper();
+        mapper.registerModule(module);
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
i-Alex, line 23: So do we need to skip unknown RPC arguments or fail?
I assume before we used a policy `FAIL_ON_UNKNOWN_PROPERTIES`

sdk/src/main/java/com/horizen/account/serialization/EthJsonMapper.java
```diff
@@ -0,0 +1,47 @@
...
+        // do not serialize null or empty values
+//        mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+//        mapper.setSerializationInclusion(JsonInclude.Include.NON_EMPTY);
+//        mapper.setSerializationInclusion(JsonInclude.Include.NON_ABSENT);
+//        mapper.enable(SerializationFeature.INDENT_OUTPUT);
```
i-Alex, line 28: how does it exactly work for us now?
`null` fields are serialized as `null`? and what about the deserialization?

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
```diff
@@ -76,49 +75,37 @@ case class AccountEthRpcRoute(
...
-                  rpcHandler.apply(new RpcRequest(body))
-                } catch {
-                  case _: RpcException => new RpcResponseError(new RpcId(), RpcError.fromCode(RpcCode.InvalidRequest, null))
-                }
-              }
-              SerializationUtil.serialize(response)
+
+            val responses = requests.map(json => Try.apply(new RpcRequest(json)).map(rpcHandler.apply) match {
+              case Success(value) => value
+              case Failure(exception: RpcException) => new RpcResponseError(new RpcId(), exception.error);
```
i-Alex, line 93: why do we "new RpcId()" instead of getting the real `id` from the request?

JcBernack: This error case here only happens when there was a problem parsing the `RpcRequest`, which is why the Id is not used. I didn't change the behavior here, but you are right this is a little incomplete. If the id is valid it should be used, even if something else in the request failed to parse.. :thinking:

sdk/src/main/java/com/horizen/account/serialization/EthJsonMapper.java
JcBernack: Yeah, I'm a bit unclear on how to do this properly. We can not just omit every `null`, because at least in a transaction we must include the `"to":null`. Maybe we can omit `null`, but overwrite the behavior for `tx.to`.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
JcBernack: Yes, see also the comment about `null`. Legacy TXs must definitely not include `"accessList":null`, for the TX types that support it, it should be empty-array instead `[]`. It was inconsistent before (null in every TX) and still is (now it is not included at all). We should fix this.

sdk/src/main/java/com/horizen/account/serialization/EthJsonMapper.java
JcBernack: This was copied from the `ObjectMapper` that was used from `ApplicationJsonSerializer`. For reference, this is a working request on Goerli via Infura:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "eth_call",
  "params": [
    {
      "to": "0x0000000000000000000000000000000000000002",
      "data": "0x516f2c0f",
      "foo": "bar"
    },
    "latest"
  ]
}
```
I think accepting unknown properties is an easy way to hide errors and would vote for removing it. We can still allow unknown properties on specific types via `@JsonIgnoreProperties(ignoreUnknown = true)`, e.g. on the `TransactionArgs` type to allow a request like the one avove.

i-Alex: agree, let's fail

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
i-Alex: I think it is important to fix now.
Waht we can do:
1) return back `accessList` field
2) set to empty array for eip1559 and to null for legacy
3) use `@JsonInclude(Include.NON_NULL)` annotation for this field.

As a result I expect it to be serialized as expected in all cases.

sdk/src/main/java/com/horizen/account/serialization/EthJsonMapper.java
i-Alex: ok, clear.
I think we may keep only `mapper.enable(SerializationFeature.INDENT_OUTPUT);`. It helps to read the outputs during the debug, for example.

other commented configs can be removed. We need to serialize `null`s by default and where it is not needed -> add annotation, like `@JsonInclude(Include.NON_NULL)` or `@JsonInclude(Include.NON_EMPTY)`

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
i-Alex: ok, we will keep it as an improvement for future.

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
JcBernack: done

akrckn (approved):

akrckn:
As mentioned in the comment at the top, there will be changes to some python tests, too, the rest looks good to me 👍

JcBernack:
python tests are all good again

akrckn (approved):
Like the additional small changes to the trace test 🙈

i-Alex (approved):

PR #740, opened 2023-02-17
py test for account size limit

alsala:
Added UT for testing account size limit checks; added json size in Si…dechainBlockBase class; fix for eth_view block size


qa/sc_evm_block_size_limit.py
```diff
@@ -0,0 +1,214 @@
...
+Test:
+    Generate a huge number of ft outputs in some FT for this SC in order to have a large set of MC ref data to be
+    included in the next SC block. Verify that only up to 5 MB of the SC block size is used for MC ref data.
+    Moreover send many Eth txes having large size and check that only up to 2 MB of the remaining block size is used
+    for including tx.
+     
+"""
+
+
+class SCEvmBlockSizeLimit(AccountChainSetup):
```
i-Alex, line 27: let's add this test to the run_tests.sh script

sdk/src/main/scala/com/horizen/api/http/BlockBaseApiRoute.scala
```diff
@@ -119,6 +119,24 @@ abstract class BlockBaseApiRoute[
     }
   }
 
+
+  /**
+   * Return here best sidechain block height in active chain
+   */
+  def getCurrentHeight: Route = (post & path("currentHeight")) {
```
i-Alex, line 126: we need update `sidechainApi.yaml` file: add new method description

sdk/src/main/scala/com/horizen/block/SidechainBlockBase.scala
```diff
@@ -37,6 +38,9 @@ abstract class SidechainBlockBase[TX <: Transaction, H <: SidechainBlockHeaderBa
 
   def feePaymentsHash: Array[Byte] = header.feePaymentsHash
 
+  @JsonProperty("size")
+  def size : Long
```
i-Alex, line 42: to we need to keep 2 separate but equal implementations for account and utxo block?
We may keep it here, no?

```suggestion
  def size: Long = bytes.length
```

i-Alex (changes requested):
looks good, minor points left

sdk/src/main/scala/com/horizen/api/http/BlockBaseApiRoute.scala
alsala: done

i-Alex (approved):

PR #742, opened 2023-02-17
Implement RPC methods net_peerCount, net_listening and web3_clientVersion

akrckn:


sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
```diff
@@ -121,4 +128,17 @@ case class AccountEthRpcRoute(
       }
     }
   }
+
+  private def getClientVersion: String = {
+    val default = "unknown"
+    val architecture = Try(System.getProperty("os.arch")).getOrElse(default)
+    val javaVersion = Try(System.getProperty("java.specification.version")).getOrElse(default)
+    val pom = Try(new MavenXpp3Reader().read(new FileReader(new File("pom.xml")))).toOption
```
akrckn, line 136: Not sure if we can rely on using this with a relative path.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -43,24 +45,27 @@ import scala.collection.JavaConverters.seqAsJavaListConverter
...
 import scala.util.{Failure, Success, Try}
 
 class EthService(
     scNodeViewHolderRef: ActorRef,
+    networkControllerRef: ActorRef,
     nvtimeout: FiniteDuration,
     networkParams: NetworkParams,
     settings: EthServiceSettings,
+    maxIncomingConnections: Int,
+    rpcClientVersion: String,
```
JcBernack, line 60: We should separate `EthService` into multiple services, like originally intended. Probably a task for later though.

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -506,6 +511,20 @@ class EthService(
   @RpcMethod("net_version")
   def version: String = String.valueOf(networkParams.chainId)
 
+  private def getPeerCount: Int = {
+    val peerReq = (networkControllerRef ? GetConnectedPeers).asInstanceOf[Future[Seq[ConnectedPeer]]]
+    Await.result(peerReq, nvtimeout).size
+  }
+
+  @RpcMethod("net_peerCount")
+  def peerCount: Quantity = new Quantity(getPeerCount)
```
JcBernack, line 520: After my RPC cleanup has been merged, `Quantity` does not exist anymore. Just return `BigInteger`, hex conversion happens automagically.

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
JcBernack: You could try `this.getClass.getPackage.getImplementationVersion`.

Note: This will return `null` in a non-packaged environment, i.e. during development when run from the IDE. It requires the JAR manifest and should work in a properly packaged build. Maybe also use the default of "unknown" or "development" when it returns `null`.

JcBernack (changes requested):

sdk/src/main/scala/com/horizen/account/api/http/AccountEthRpcRoute.scala
i-Alex: I have a doubts about this implementation as well.
We need something more stable, like here: https://stackoverflow.com/a/70307580

sdk/src/main/scala/com/horizen/account/api/rpc/service/EthService.scala
i-Alex: good point, but for now let's keep as is.

sdk/pom.xml
```diff
@@ -157,6 +157,12 @@
       <artifactId>mockito-4-2_2.12</artifactId>
       <version>3.2.11.0</version>
     </dependency>
+    <!-- https://mvnrepository.com/artifact/org.apache.maven/maven-model -->
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-model</artifactId>
+      <version>3.9.0</version>
+    </dependency>
```
i-Alex, line 165: why do we need this dependency?

qa/sc_evm_rpc_net_web3_methods.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Connect SC node
+        connect_sc_nodes(sc_node_1, 1)
+
+        assert_true(int(sc_node_1.rpc_net_peerCount()['result'], 16) == 1)
+        assert_true(sc_node_1.rpc_net_listening()['result'] is True)
+
+        # Disconnect SC node
+        disconnect_sc_nodes_bi(self.sc_nodes, 0, 1)
+        assert_true(int(sc_node_1.rpc_net_peerCount()['result'], 16) == 0)
```
i-Alex, line 40: add the check for `net_listening`

qa/sc_evm_rpc_net_web3_methods.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Check web3_clientVersion exists
+        assert_true(len(sc_node_1.rpc_web3_clientVersion()['result']) > 6)
```
i-Alex, line 43: let's add at least some basic structure checks.

i-Alex (changes requested):

sdk/pom.xml
akrckn: We don't anymore - that was for reading the pom.xml

JcBernack (approved):
nice :+1:

qa/sc_evm_rpc_net_web3_methods.py
```diff
@@ -0,0 +1,58 @@
...
+        # Connect SC node
+        connect_sc_nodes(sc_node_1, 1)
+
+        assert_true(int(sc_node_1.rpc_net_peerCount()['result'], 16) == 1)
+        assert_true(sc_node_1.rpc_net_listening()['result'] is True)
+
+        # Disconnect SC node
+        disconnect_sc_nodes_bi(self.sc_nodes, 0, 1)
+        assert_true(int(sc_node_1.rpc_net_peerCount()['result'], 16) == 0)
+        assert_true(sc_node_1.rpc_net_listening()['result'] is True)
```
i-Alex, line 43: one last improvement for `rpc_net_listening`.
Let's have check for both `True` and `False ` cases.
How can we do it? See `SCNodeConfiguration` class constructor. It has `max_incoming_connections=100` parameter. We may configure both nodes with value `1`.

Then connect SC node 1 to SC node 2 (as now). In this case we achieve that node 1 has 1 outgoing connection, so may listen to the incoming. Node 2 has 1 incoming connection, so doesn't listen anymore.

Here after nodes being disconnected: both should return `True`

i-Alex (changes requested):
Looks good, let's extend py test a bit more.

davehorizen (approved):
lgtm but, as a non-blocking activity for this PR, I suggest to separate the python test between `net` and `web3` rpc namespace

i-Alex (approved):

PR #743, opened 2023-02-20
St/mempool non exec

saratnt:


qa/SidechainTestFramework/sc_boostrap_info.py
```diff
@@ -138,7 +138,9 @@ def __init__(self,
                  storage_backup_delay='5m',
                  max_nonce_gap=16,
                  max_account_slots=16,
-                 max_mempool_slots=6144,):
+                 max_mempool_slots=6144,
+                 max_nonexec_pool_slots=1024,
```
i-Alex, line 142:
```suggestion
                 max_nonexec_pool_slots=1024
```

qa/SidechainTestFramework/scutil.py
```diff
@@ -502,6 +502,7 @@ def initialize_sc_datadir(dirname, n, bootstrap_info=SCBootstrapInfo, sc_node_co
         'MAX_NONCE_GAP': sc_node_config.max_nonce_gap,
         'MAX_ACCOUNT_SLOTS': sc_node_config.max_account_slots,
         'MAX_MEMPOOL_SLOTS': sc_node_config.max_mempool_slots,
+        'MAX_NONEXEC_SLOTS': sc_node_config.max_nonexec_pool_slots,
```
i-Alex, line 505:
```suggestion
        'MAX_NONEXEC_SLOTS': sc_node_config.max_nonexec_pool_slots
```

sdk/src/main/scala/com/horizen/SidechainSettings.scala
```diff
@@ -92,10 +93,16 @@ case class EthServiceSettings(
...
+                                  maxMemPoolSlots: Int = 6144,
+                                  maxNonExecMemPoolSlots: Int = 1024){
   require(maxNonceGap > 0, s"Maximum Nonce Gap not positive: $maxNonceGap")
   require(maxAccountSlots > 0, s"Maximum Account Slots not positive: $maxAccountSlots")
-  require(maxMemPoolSlots > 0, s"Maximum Memory Pool Slots not positive: $maxMemPoolSlots")
+  require(maxMemPoolSlots >= MempoolMap.MaxNumOfSlotsForTx, s"Maximum Memory Pool Slots number should be at least " +
+    s"${MempoolMap.MaxNumOfSlotsForTx} but it is $maxMemPoolSlots")
+  require(maxNonExecMemPoolSlots >= MempoolMap.MaxNumOfSlotsForTx, s"Maximum Non Executable Memory Sub Pool Slots number " +
+    s"should be at least ${MempoolMap.MaxNumOfSlotsForTx} but it is $maxNonExecMemPoolSlots")
+  require(maxNonExecMemPoolSlots < maxMemPoolSlots, s"Maximum Non Executable Memory Sub Pool Slots " +
```
i-Alex, line 104: `<` or `<=` ?
do we allow max non exec slots to be equal to max slots? I think, yes

sdk/src/main/scala/com/horizen/account/mempool/TxCache.scala
```diff
@@ -69,14 +94,49 @@ class TxCache {
 
   def getSizeInSlots: Int = sizeInSlots
 
+  def getNonExecSizeInSlots: Int = nonExecSizeInSlots
+
   def getOldestTransaction(): Option[SidechainTypes#SCAT] = oldestTx.map(_.tx)
 
   def getYoungestTransaction(): Option[SidechainTypes#SCAT] = youngestTx.map(_.tx)
 
+  def getNonExecIterator(): NonExecTransactionIterator = new NonExecTransactionIterator
```
i-Alex, line 103: let's put a comment to the method: from oldest to youngest

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
```diff
@@ -346,8 +371,8 @@ class MempoolMap(
             (balance.compareTo(tx.maxCost) >= 0) &&
             (currAccountSlots + txSizeInSlots <= MaxSlotsPerAccount)
           ) {
-            //txCache.add(tx)
-            txsToReinject += tx
+            val tmp = (tx, haveBecomeNonExecutable)
+            txsToReinject += tmp
```
i-Alex, line 375:
```suggestion
            txsToReinject += (tx, haveBecomeNonExecutable)
```

i-Alex (commented):
LGTM!

sdk/src/main/scala/com/horizen/SidechainSettings.scala
saratnt: If max non exec is less than max slots, there will be always room for an executable tx and a malicious user won't be able to use non-exec txs to evict exec ones. Let's say the max slots = max non exec  = 10. Let's suppose that in mempool there are 4 exec txs and 6 non exec txs. I could create enough non exec txs (the exact number depends on how old are the exec txs) so to evict the 4 exec ones. 
But it is a complicated matter, maybe it is better to discuss it.

sdk/src/main/scala/com/horizen/account/mempool/TxCache.scala
saratnt: Done

sdk/src/main/scala/com/horizen/account/mempool/TxCache.scala
```diff
@@ -69,14 +94,50 @@ class TxCache {
...
+    @tailrec
+    final private[mempool] def findNext(txInfo: Option[TxMetaInfo]): Option[TxMetaInfo] = {
+      if (txInfo.isEmpty || txInfo.get.isNotExecutable)
+        txInfo
+      else {
+        findNext(txInfo.get.next)
+      }
+    }
+
+    def hasNext: Boolean = {
```
alsala, line 125: I see that we just use it in UT, that is because we are using this iterator only when we are sure (comparing slot sizes) there is at least a non exec tx in mempool, correct?

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
```diff
@@ -122,33 +122,35 @@ class MempoolMap(
           val existingTxWithSameNonceIdOpt = nonExecTxsPerAccount.get(ethTransaction.getNonce)
           if (existingTxWithSameNonceIdOpt.isDefined) {
             val existingTxWithSameNonceId = existingTxWithSameNonceIdOpt.get
-            replaceTransaction(existingTxWithSameNonceId, ethTransaction, nonExecTxsPerAccount)
+            replaceTransaction(existingTxWithSameNonceId, ethTransaction, nonExecTxsPerAccount, isNotExec = true)
```
alsala, line 125: Cosmetic: here and in other method signatures, use is made of  a boolean for distinguishing exec/nonexec. Consider using an Enumeration or a constant, which would render the caller code more readable

alsala (commented):
LGTM, just a couple of comments

sdk/src/main/scala/com/horizen/account/mempool/TxCache.scala
saratnt: Correct. In fact, this method could be removed. I have left it just because iterators have hasNext.

sdk/src/main/scala/com/horizen/account/mempool/MempoolMap.scala
saratnt: Done

sdk/src/main/scala/com/horizen/SidechainSettings.scala
i-Alex: it seems like 1 slot will not make a big difference.
It is a responsibility of the Node holder to choose the values in a way to keep some space for executables only.

i-Alex (approved):

alsala (approved):

PR #745, opened 2023-02-21
Add protection against replay attacks to RPC/API

i-skrypnyk:


sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -456,6 +458,10 @@ class AccountState(
 
     ethTx.semanticValidity()
 
+    if (!params.allowUnprotectedTxs && tx.isLegacy && !tx.isEIP155) {
+      throw new IllegalArgumentException("Legacy unprotected transaction are not allowed.")
+    }
```
i-Alex, line 463: we need to keep this check in the `AccountSidechainNodeViewHolder.applyLocallyGeneratedTransactions` method, since it is not a consensus rule.
Legacy unprotected transactions are allowed to be a part of the block.

sdk/src/main/scala/com/horizen/params/NetworkParams.scala
```diff
@@ -47,6 +47,7 @@ trait NetworkParams {
   val cswVerificationKeyFilePath: String
   val sidechainCreationVersion: SidechainCreationVersion
   val isCSWEnabled: Boolean
+  val allowUnprotectedTxs: Boolean
```
i-Alex, line 50: I think we don't need to change NetworkParams and ancestors of it because `allowUnprotectedTxs` is a Node config not a Network config.
In nodeViewHolder we have an access to the `SidechainSettings`

i-Alex (changes requested):

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
saratnt: Another reason to move this check in `AccountSidechainNodeViewHolder.applyLocallyGeneratedTransactions`is that `validate` is called also for remotely generated txs, so legacy txs created by nodes that allow them would be rejected.

saratnt (approved):
LGTM!

qa/sc_evm_forbid_unprotected_txs.py
```diff
@@ -0,0 +1,141 @@
...
+    assert_true, assert_equal, bootstrap_sidechain_nodes, AccountModel, connect_sc_nodes
+from httpCalls.block.best import http_block_best
+from httpCalls.transaction.allTransactions import allTransactions
+from test_framework.util import websocket_port_by_mc_node_index, forward_transfer_to_sidechain, fail
+
+"""
+Check the EVM SC allowUnprotectedTxs property behaviour.
+
+Configuration: bootstrap 1 SC node and start it with allowUnprotectedTxs set to false.
+    - Create 1 SC node
```
i-Alex, line 22: update description: now we start 2 nodes

sdk/src/main/scala/com/horizen/account/AccountSidechainNodeViewHolder.scala
```diff
@@ -171,6 +173,19 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
       else
 
        */
+      if (tx.isInstanceOf[EthereumTransaction]) {
+        val ethTx: EthereumTransaction = tx.asInstanceOf[EthereumTransaction]
+        if (!sidechainSettings.mempool.allowUnprotectedTxs && ethTx.isLegacy && !ethTx.isEIP155) {
+          context.system.eventStream.publish(
+            FailedTransaction(
+              tx.asInstanceOf[Transaction].id,
```
i-Alex, line 181:
```suggestion
              tx.id
```
I think it should work, in a worse case use `ModifierId @@ tx.id`.
No sense to cast is to `Transaction` type

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -19,8 +19,10 @@ import com.horizen.evm.interop.EvmLog
 import com.horizen.evm.utils.{Address, Hash}
 import com.horizen.params.NetworkParams
 import com.horizen.state.State
+import com.horizen.transaction.Transaction
```
i-Alex, line 22: import can be removed

sdk/src/main/scala/com/horizen/account/state/AccountState.scala
```diff
@@ -19,8 +19,10 @@ import com.horizen.evm.interop.EvmLog
...
 import com.horizen.utils.{ByteArrayWrapper, BytesUtils, ClosableResourceHandler, MerkleTree, TimeToEpochUtils, WithdrawalEpochInfo, WithdrawalEpochUtils}
 import sparkz.core._
+import sparkz.core.network.NodeViewSynchronizer.ReceivableMessages.FailedTransaction
```
i-Alex, line 25: import can be removed

sdk/src/test/scala/com/horizen/account/state/AccountStateTest.scala
```diff
@@ -156,17 +156,4 @@ class AccountStateTest
...
-    val tx = mock[EthereumTransaction]
-
-    Mockito.when(tx.semanticValidity()).thenAnswer(_ => true)
-    Mockito.when(tx.getGasLimit).thenReturn(FeeUtils.GAS_LIMIT.add(BigInteger.ONE))
-
-    state.validate(tx) match {
-      case Failure(_) =>
-      case Success(_) => Assert.fail("Transaction with gas limit greater than block is expected to fail")
-    }
-  }
```
i-Alex, line 171: why this test method was removed?

i-Alex (changes requested):
Looks good!
Some minor cleanup needed

qa/sc_evm_forbid_unprotected_txs.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # tx3: forced tx
+        tx_bytes = createLegacyTransaction(sc_node_1,
+                                           fromAddress=evm_hex_address,
+                                           toAddress=recipient_proposition,
+                                           value=1,
+                                           output_raw_bytes=True)
+
+        forced_tx = signTransaction(sc_node_1, fromAddress=evm_hex_address, payload=tx_bytes)
+        block_id = generate_next_block(sc_node_1, "first node", forced_tx=[forced_tx])
```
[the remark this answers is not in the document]
i-skrypnyk: I don't want to spend another couple of hours figuring out how to make node 2 capable of forging :D Also, I wanted to check specifically this path of forced inclusion, and the next transaction is created on node_2 in a regular way

sdk/src/main/scala/com/horizen/account/AccountSidechainNodeViewHolder.scala
i-skrypnyk: tx.id works, thanks

sdk/src/test/scala/com/horizen/account/state/AccountStateTest.scala
i-skrypnyk: by mistake, returned it back

qa/sc_evm_forbid_unprotected_txs.py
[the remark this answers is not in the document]
saratnt: :D For me it's ok :)

i-Alex (approved):

PR #747, opened 2023-02-22
EVM package cleanup

JcBernack:
Last round of refactoring before the `evm` package and `libevm` are moved into a separate repository:
- renamed namespace `com.horizen.evm` to `io.horizen.evm`
- moved heavily used types like `Address` and `Hash` from the `utils` sub-package to the main package `io.horizen.evm`
- added `EthereumConsensusDataLog` to the SDK and refactored all usages of `EvmLog`
- reduced `EvmLog` to a POJO like originally intended (its only usage in the SDK is now in `StateDbAccountStateView` where it is mapped to `EthereumConsensusDataLog`)
- reduced visibility of internal types within as much as possible
- refactored all POJO types to be immutable as much as possible (we don't have immutable arrays)
- simplified mapping of data to and from JSON for libevm invocations


evm/src/main/java/com/horizen/evm/interop/TracerConfig.java
```diff
@@ -1,23 +0,0 @@
-package com.horizen.evm.interop;
-
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-
-@JsonIgnoreProperties(ignoreUnknown = true)
-public class TracerConfig {
```
i-Alex, line 6: As I understood it is data-class specific to `callTracer` only.
So we now keep this field as a part of `TraceOptions` class in a free JsonNode form and delegate go part to validate and use it to configure a specific tracer.
So in case of any custom creation error:
```
if err != nil {
			log.Warn("failed to create tracer: %v", err)
		}
```
we will just ignore tracer the same as in Geth, correct?

evm/src/main/java/io/horizen/evm/LibEvmCallback.java
```diff
@@ -56,6 +56,7 @@ private static String invoke(int handle, String args) {
     static class CallbackProxy implements Callback {
         public String callback(int handle, Pointer msg) {
             try {
+                // we do not need to free the Pointer here, as it is freed on the libevm side when the callback returns
```
i-Alex, line 59: in general, would be nice to check libevm/evm modules against the memory leaks

evm/src/main/java/io/horizen/evm/TraceOptions.java
```diff
@@ -0,0 +1,46 @@
+package io.horizen.evm;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class TraceOptions {
```
i-Alex, line 6: `timeout` field was removed during the refactoring.
why?

i-Alex (commented):
review is in progress

i-Alex (changes requested):
Changes looks amazing!

evm/src/main/java/io/horizen/evm/TraceOptions.java
JcBernack: It was never used and already removed on the go side in the original PR. On the SDK side it was kept accidentally.

evm/src/main/java/com/horizen/evm/interop/TracerConfig.java
JcBernack: Good point. In case of an error we log a warning and continue without a tracer, Geth would throw an error. I will change this.

Currently, the only native tracer that has any arguments is the `callTracer`, correct. Theoretically though, they all can have different parameters, which is why we don't parse `TraceOptions.tracerConfig` anymore and just pass it along as raw JSON.

evm/src/main/java/io/horizen/evm/LibEvmCallback.java
JcBernack: True, not possible from the "inside" though, is it? I assume we would need some kind of benchmark and run it under a memory profiler.

i-Alex: correct, let's keep it in mind after libevm separation

evm/src/main/java/io/horizen/evm/TraceOptions.java
i-Alex: ok, I see.
was confused with the outdated reference: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#traceconfig

evm/src/main/java/com/horizen/evm/interop/TracerConfig.java
JcBernack: done

JcBernack: test case for an invalid tracer added to `sc_evm_test_debug_methods.py`

i-Alex (approved):

PR #750, opened 2023-02-23
Check required fields for getBlock... and getTransactionByHash

akrckn:
Delete author in EthereumBlockView - the rest mentioned in [612](https://horizenlabs.atlassian.net/jira/software/c/projects/EVM/boards/12?modal=detail&selectedIssue=EVM-612&assignee=624db49aad6b7e006aa91a3c) are needed and the behavior is the same to what ethereum does.

### DIFF
--- a/sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
+++ b/sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
@@ -26,8 +26,11 @@ public class EthereumTransactionView {
     public final byte[] input;
     public final BigInteger gas;
     public final BigInteger gasPrice;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     public final BigInteger maxPriorityFeePerGas;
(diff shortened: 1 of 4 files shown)


akrckn:
Edit: Might be - due to some compiling issues earlier - that I tested with a version before the latest rpc changes - I'll redo the tests and re-assign the Review then!

JcBernack (approved):

sdk/src/main/java/com/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -26,8 +26,11 @@ public class EthereumTransactionView {
     public final byte[] input;
     public final BigInteger gas;
     public final BigInteger gasPrice;
+    @JsonInclude(JsonInclude.Include.NON_NULL)
```
i-Alex, line 29: let's extend one of our Py or Unit tests and check these 3 fields are included only for eip-1559 tx.

i-Alex (changes requested):

i-Alex (approved):

PR #752, opened 2023-02-27
Updated Bcrypt library

MarcoOl94:
## Description
Update Bcrypt library.

## Jira Ticket
[756](https://horizenlabs.atlassian.net/jira/software/c/projects/SDK/boards/14?modal=detail&selectedIssue=SDK-756&assignee=615f31c32f6aed0068c67291)

## Changes

- Updated Bcrypt library inside Bootstrapping tool

## Breaking Changes
- No breaking changes, the new library allows to use older version of algorithm

## Checks
- [✔] Project Builds
- [✔] Updated documentation accordingly

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -168,7 +168,7 @@ HTTP request:
 Config file:
 
         restApi {
-            "apiKeyHash": "$2a$10$6CUHuJzctrPAsxWO5K4Pi.mx3OV0TIb2bSaq4ZBZQsR0MV3SVB6rC"
+            "apiKeyHash": "$2y$12$vga1LEzU1jiLYI766CIeVOi1A9QwFBqYgjbAsD.2t8Z7SFP6ff4Eq"
         }
 
 **Sidechain-related RPC commands in Mainchain**
--- a/examples/README.md
+++ b/examples/README.md
@@ -168,7 +168,7 @@ HTTP request:
 Config file:
 
         restApi {
-            "apiKeyHash": "$2y$12$vga1LEzU1jiLYI766CIeVOi1A9QwFBqYgjbAsD.2t8Z7SFP6ff4Eq"
+            "apiKeyHash": "$2y$08$sB9Zrn4S2/YYfJ8/lhNxXO/Ku7uciazel3hx4bvjMs7nrqcLqMS0y"
         }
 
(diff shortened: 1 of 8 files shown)


tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
```diff
@@ -624,7 +624,8 @@ private void processEncodeString(JsonNode json ) {
             return;
         }
         String toEncode = json.get("string").asText();
-        String encoded = BCrypt.hashpw(toEncode, BCrypt.gensalt());
+        String encoded = BCrypt.with(BCrypt.Version.VERSION_2Y).hashToString(12, toEncode.toCharArray());
```
i-Alex, line 627: Consider the performance of the algorithm for the factor 12.
Discuss with Cronic the suitable value of `factor` and ask if we need to let user choose the factor or/and the salt values

i-Alex (changes requested):

sdk/src/test/scala/com/horizen/account/api/http/AccountEthRpcRouteMock.scala
```diff
@@ -52,7 +52,7 @@ abstract class AccountEthRpcRouteMock extends AnyWordSpec with Matchers with Sca
 
   val credentials = HttpCredentials.createBasicHttpCredentials("username","password")
   val badCredentials = HttpCredentials.createBasicHttpCredentials("username","wrong_password")
-  val apiKeyHash = BCrypt.hashpw(credentials.password(), BCrypt.gensalt())
+  val apiKeyHash = BCrypt.`with`(BCrypt.Version.VERSION_2Y).hashToString(12, credentials.password().toCharArray)
```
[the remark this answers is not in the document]
MarcoOl94: Good point, I moved to a separate fixture

tools/sctool/src/main/java/com/horizen/ScBootstrappingToolCommandProcessor.java
MarcoOl94: I've talked with cronic and we agreed to use 8 as a algorithm cost factor

PR #756, opened 2023-03-01
Added try/match blocks for ABI decode() calls in native smart contrac…

alsala:
…ts for throwing ExecutionRevertedException; added some more UT


sdk/src/main/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessor.scala
```diff
@@ -147,7 +148,17 @@ case class CertificateKeyRotationMsgProcessor(params: NetworkParams) extends Nat
   private def execSubmitKeyRotation(msg: Message, view: BaseAccountStateView, currentEpochNum: Int): Array[Byte] = {
     //verify
     checkMessageValidity(msg)
-    val inputData = SubmitKeyRotationCmdInputDecoder.decode(getArgumentsFromData(msg.getData))
+
+    val inputData : SubmitKeyRotationCmdInput = Try {
```
i-Alex, line 152: check the method below `checkKeyRotationProofValidity`.
in case the `keyRotationProof.newKey` is not a valid key then
```
    val newKeyAsMessage = keyRotationProof.keyType match {
      case SigningKeyRotationProofType => CryptoLibProvider.thresholdSignatureCircuitWithKeyRotation
        .getMsgToSignForSigningKeyUpdate(keyRotationProof.newKey.pubKeyBytes(), currentEpochNum, params.sidechainId)
      case MasterKeyRotationProofType => CryptoLibProvider.thresholdSignatureCircuitWithKeyRotation
        .getMsgToSignForMasterKeyUpdate(keyRotationProof.newKey.pubKeyBytes(), currentEpochNum, params.sidechainId)
    }
``` may throw a run time exception.

Let's wrap the method inside with `Try(fn).recoverWith {` and check in case method throws exception not of type `ExecutionFailedException` -> wrap it with `ExecutionRevertedException` and throw outside.

In general we must be careful with sc_cryptolib objects in msg processors.
We may discuss if needed.

**Review by i-Alex — CHANGES_REQUESTED** [review]

File: sdk/src/main/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessor.scala (hunk 1 above)
Comment (alsala) [reply sdk/src/main/scala/com/horizen/account/state/CertificateKeyRotationMsgProcessor.scala]: Done. I also added some more check to `ThresholdSignatureCircuitWithKeyRotationImplZendoo.java` in order to avoid sending bad data to sc crypto lib which might lead to a rust panic event

File: sdk/src/main/scala/io/horizen/account/state/CertificateKeyRotationMsgProcessor.scala [hunk 2]
```diff
@@ -143,6 +143,10 @@ case class CertificateKeyRotationMsgProcessor(params: NetworkParams) extends Nat
 
     if (!newKeySignature.isValid(keyRotationProof.newKey, newKeyAsMessage))
       throw new ExecutionRevertedException(s"Key rotation proof - self signature is invalid: $index")
+
+  } recoverWith {
+     case t : Exception =>
+     Failure(t)
```
i-Alex, line 149: `recoverWith` is not needed anymore. Here we just throw the same exception.

alsala: Right, done

i-Alex (approved):

sdk/src/main/scala/io/horizen/account/state/ForgerStakeMsgProcessor.scala
```diff
@@ -354,7 +389,7 @@ case class ForgerStakeMsgProcessor(params: NetworkParams) extends NativeSmartCon
       case AddNewStakeCmd => doAddNewStakeCmd(msg, gasView)
       case RemoveStakeCmd => doRemoveStakeCmd(msg, gasView)
       case OpenStakeForgerListCmd => doOpenStakeForgerListCmd(msg, gasView)
-      case opCodeHex => throw new ExecutionRevertedException(s"op code $opCodeHex not supported")
+      case opCodeHex => throw new ExecutionRevertedException(s"op code not supported: $opCodeHex")
```
saratnt, line 392: I have a doubt: is this the same exception thrown by EVM when an non existing function is called on a smart contract? It may have sense to have the same message thrown by a real smart contract for a native smart contract.

sdk/src/main/scala/io/horizen/account/state/WithdrawalMsgProcessor.scala
```diff
@@ -104,7 +113,14 @@ object WithdrawalMsgProcessor extends NativeSmartContractMsgProcessor with Withd
     val nextNumOfWithdrawalReqs: Int = numOfWithdrawalReqs + 1
     setWithdrawalEpochCounter(view, currentEpochNum, nextNumOfWithdrawalReqs)
 
-    val inputParams = AddWithdrawalRequestCmdInputDecoder.decode(getArgumentsFromData(msg.getData))
+    val inputParams : AddWithdrawalRequestCmdInput = Try {
```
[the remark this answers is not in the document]
alsala: Done, and thanks for the hint :)

sdk/src/main/scala/io/horizen/account/state/ForgerStakeMsgProcessor.scala
alsala: It is not clear to me where this error msg should be found. For the time being I propose to leave it as it is.

saratnt: For me it's ok

PR #757, opened 2023-03-02
Sdk 747: add redeem message side [Account] 

paologalligit:
This pr adds:

- AbstractCrossChainRedeemMessageProcessor + unit test
- storage of scTxCommitmentTreeHash for validation purposes


sdk/src/main/scala/com/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,181 @@
...
+   * </ul>
+   *
+   * @param msg          message to apply to the state
+   * @param view         state view
+   * @param gas          available gas for the execution
+   * @param blockContext contextual information accessible during execution
+   * @return return data on successful execution
+   */
+  @throws(classOf[ExecutionFailedException])
+  override def process(msg: Message, view: BaseAccountStateView, gas: GasPool, blockContext: BlockContext): Array[Byte] = {
```
paolocappelletti, line 39: the process() method is the main entry point of the fake smart contract
inside it we shooud check if the requested function in msg is one defined by our smart contract and redirect to proper action etc...
since we are defining just an abstract class we can leave it unimplemented and delegate subclasses to implement it.
see what has been done in the sender part:
https://github.com/HorizenOfficial/Sidechains-SDK/blob/pc/sc2sc/sdk/src/test/scala/com/horizen/account/sc2sc/CrossChainMessageProcessorFixture.scala#L29

the code here we can be in a protected method processRedeemMesssage(msg: AccountCrossChainRedeemMessage)

paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -359,7 +357,7 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
     // If sidechain has ceased set the flag
     hasCeasedOpt.foreach(_ => updateList.add(new JPair(ceasingStateKey, new ByteArrayWrapper(Array.emptyByteArray))))
 
-    scTxCommitmentTreeRootHash.foreach(hash => updateList.add(new JPair(getSidechainTxCommitmentTreeHashKey(hash), hash)))
+    scTxCommitmentTreeRootHashesSeq.foreach(hash => updateList.add(new JPair(getSidechainTxCommitmentTreeHashKey(hash), hash)))
```
paolocappelletti, line 360: check the method getSidechainTxCommitmentTreeHashKey(...)
must generate a different key for every hash

sdk/src/main/scala/com/horizen/account/state/AccountStateView.scala
```diff
@@ -97,10 +97,4 @@ class AccountStateView(
   }
 
   override def getAccountStateRoot: Array[Byte] = metadataStorageView.getAccountStateRoot
-
-  override def doesScTxCommitmentTreeRootExist(hash: Array[Byte]): Boolean =
```
paolocappelletti: you need this, to be used in the message processor

sdk/src/main/scala/com/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -87,11 +89,11 @@ abstract class AbstractCrossChainRedeemMessageProcessor(
                                             scCommitmentTreeRoot: Array[Byte],
                                             nextScCommitmentTreeRoot: Array[Byte],
                                             view: BaseAccountStateView): Unit = {
-    if (!view.doesScTxCommitmentTreeRootExist(scCommitmentTreeRoot)) {
+    if (view.getAccountStorage(contractAddress, scCommitmentTreeRoot).isEmpty) {
```
paolocappelletti, line 92: the treeroot is not saved in the account storage.
must be taken from the metadatastorage (throug the accountstateview that is the implementation of the BaseAccountStateView hierarchy)

paolocappelletti (changes requested):

sdk/src/main/scala/com/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -86,14 +86,18 @@ abstract class AbstractCrossChainRedeemMessageProcessor(
     }
   }
 
+  override def addScTxCommitmentTreeRootHash(hash: Array[Byte], view: BaseAccountStateView): Unit =
+    view.updateAccountStorage(contractAddress, hash, Array.emptyByteArray)
```
paolocappelletti, line 90: let's prepend a prefix to the key hash, to avoid collisions with the message hash stored in line 125

paolocappelletti (approved):

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  private def validateScCommitmentTreeRoot(
+                                            scCommitmentTreeRoot: Array[Byte],
+                                            nextScCommitmentTreeRoot: Array[Byte],
+                                            view: BaseAccountStateView): Unit = {
+    if (view.getAccountStorage(contractAddress, getScTxCommitmentTreeRootHashKey(scCommitmentTreeRoot)).isEmpty) {
+      throw new IllegalArgumentException(s"Sidechain commitment tree root `${BytesUtils.toHexString(scCommitmentTreeRoot)}` does not exist")
+    }
+
+    if (view.getAccountStorage(contractAddress, getScTxCommitmentTreeRootHashKey(nextScCommitmentTreeRoot)).isEmpty) {
```
saratnt, line 101: For checking if a key is defined in the stateDb, this is the check that needs to be done:
`getAccountStorage(.....).sameElements(NULL_HEX_STRING_32)`

saratnt (changes requested):
I was wondering if it is possible to create a separate MessageProcessor for storing the SCCommitmentTreehashroot, so each RedeemMessageProcessor won't need to store this data but it will just ask the "SCCommitmentTree" MessageProcessor for it. What do you think about it?

saratnt (changes requested):

saratnt (approved):

sdk/src/test/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessorTest.scala
```diff
@@ -0,0 +1,233 @@
...
+
+class CrossChainRedeemMessageProcessorImpl(networkParams: NetworkParams, sc2scCircuit: Sc2scCircuit, scTxCommTreeRootProvider: ScTxCommitmentTreeRootHashMessageProvider)
+  extends AbstractCrossChainRedeemMessageProcessor(networkParams, sc2scCircuit, scTxCommTreeRootProvider) {
+
+  override val contractAddress: Address = CrossChainRedeemMessageProcessorImpl.contractAddress
+  override val contractCode: Array[Byte] = CrossChainRedeemMessageProcessorImpl.contractCode
+
+  override protected def getAccountCrossChainRedeemMessageFromMessage(msg: Message): AccountCrossChainRedeemMessage = {
+    val accountCrossChainMessage = AccountCrossChainMessage(
+      messageType = 1,
```
alsala, line 187: What is the meaning of the `messageType` field? Is '1' the only supported value? Consider using an enum for instance in the `AccountCrossChainMessage` class.

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * Extract the AccountCrossChainRedeemMessage from the message
+   *
+   * @param msg message to apply to the state
+   * @return the AccountCrossChainRedeemMessage built from the arguments in the message
+   */
+  protected def getAccountCrossChainRedeemMessageFromMessage(msg: Message): AccountCrossChainRedeemMessage
+
+  private def validateRedeemMsg(ccRedeemMgs: AccountCrossChainRedeemMessage, view: BaseAccountStateView): Unit = {
+    try {
```
alsala, line 60: Does it make sense to check also `messageType` value of `accountCcMsg`?

Moreover, I understand that `sender/receiver` have a generic byte array format therefore are not checked on purpose, correct?

But however consider checking them not to be null or zero-length.
Finally, does it make sense for them to have a maximum size in order to prevent malicious behaviour?

sdk/src/main/scala/io/horizen/account/sc2sc/AccountCrossChainRedeemMessage.scala
```diff
@@ -0,0 +1,68 @@
...
+  nextCertificateDataHash: Array[Byte],
+  scCommitmentTreeRoot: Array[Byte],
+  nextScCommitmentTreeRoot: Array[Byte],
+  proof: Array[Byte]
+) extends BytesSerializable with ABIEncodable[StaticStruct] {
+  override type M = AccountCrossChainRedeemMessage
+
+  override def serializer: SparkzSerializer[AccountCrossChainRedeemMessage] = AccountCrossChainRedeemMessageSerializer
+
+  override def asABIType(): StaticStruct =
```
alsala, line 21: Many of these parameters are hashes could we use Byte32 ABI type for them?

sdk/src/main/scala/io/horizen/account/sc2sc/AccountCrossChainRedeemMessage.scala
```diff
@@ -0,0 +1,68 @@
...
+    new StaticStruct(
+      new DynamicBytes(accountCrossChainMessage.bytes),
+      new DynamicBytes(certificateDataHash),
+      new DynamicBytes(nextCertificateDataHash),
+      new DynamicBytes(scCommitmentTreeRoot),
+      new DynamicBytes(nextScCommitmentTreeRoot)
```
alsala, line 27: Why we do not include `proof` here?

sdk/src/main/scala/io/horizen/account/sc2sc/ScTxCommitmentTreeRootHashMessageProcessor.scala
```diff
@@ -0,0 +1,40 @@
...
+  override def process(msg: Message, view: BaseAccountStateView, gas: GasPool, blockContext: BlockContext): Array[Byte] =
+    throw new ExecutionRevertedException("Cannot call ScTxCommitmentTreeRootHashMessageProcessor directly")
+
+  override def addScTxCommitmentTreeRootHash(hash: Array[Byte], view: BaseAccountStateView): Unit =
+    view.updateAccountStorage(contractAddress, getScTxCommitmentTreeRootHashKey(hash), Array.emptyByteArray)
+
+  override def doesScTxCommitmentTreeRootHashExist(hash: Array[Byte], view: BaseAccountStateView): Boolean =
+    !view.getAccountStorage(contractAddress, getScTxCommitmentTreeRootHashKey(hash)).sameElements(NULL_HEX_STRING_32)
+
+  private[sc2sc] def getScTxCommitmentTreeRootHashKey(hash: Array[Byte]): Array[Byte] =
```
alsala, line 35: Cosmetic: code may be indentedhere and below

sdk/src/main/scala/io/horizen/account/state/events/AddCrossChainRedeemMessage.scala
```diff
@@ -0,0 +1,35 @@
...
+             accountCrossChainMessage: AccountCrossChainMessage,
+             certificateDataHash: Array[Byte],
+             nextCertificateDataHash: Array[Byte],
+             scCommitmentTreeRoot: Array[Byte],
+             nextScCommitmentTreeRoot: Array[Byte],
+             proof: Array[Byte],
+           ): AddCrossChainRedeemMessage =
+    new AddCrossChainRedeemMessage(
+      new DynamicBytes(AccountCrossChainMessageSerializer.toBytes(accountCrossChainMessage)),
+      new DynamicBytes(certificateDataHash),
```
alsala, line 29: Some of these fields are hashes, could we use Byte32 ABI type for them?

alsala (changes requested):

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
paologalligit: We don't enforce any check on `messageType` at this level, this is something that the implementation will do.
Correct about the `sender/receiver`, at this level we're just interested into checking the receiving sidechain is correct; however we could check whether they're null, empty or too large.

sdk/src/main/scala/io/horizen/account/sc2sc/AccountCrossChainRedeemMessage.scala
paologalligit: Because I forgot it 😅
I'm adding it

alsala (approved):

PR #759, opened 2023-03-02
Modified semanticValidity() avoiding doing a check on signature twice

alsala:


sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -474,8 +476,11 @@ public String toString() {
     }
 
     @Override
-    public byte[] messageToSign() {
-       return encode(false);
+    public synchronized byte[] messageToSign() {
+        if (this.messageToSign == null) {
+            this.messageToSign = encode(false);
+        }
+        return this.messageToSign;
```
saratnt, line 483: Maybe it would be better to return a copy of `this.messageToSign` because the content of an array is always modifiable and someone could change the contained values.

alsala: That is true, but copying would be expensive. Let's discuss. We are also exposing `getData()` accessor method which behaves similarly.

saratnt: ok

i-Alex (approved):

saratnt (approved):

PR #760, opened 2023-03-02
Added log for tracing block reward payments

alsala:


sdk/src/main/scala/io/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -324,6 +324,19 @@ class AccountForgeMessageBuilder(
       logsBloom
     )
 
+    if (isWithdrawalEpochLastBlock) {
```
i-Alex, line 327: I'm thinking to put this info in the AccountState.
Currently this log data will be available only in a single Forger node.
What do you think?

i-Alex (approved):

PR #762, opened 2023-03-03
Is/consensus protocol change

i-skrypnyk:
websocket_server.py is failing, but it is also failing on dev_evm branch, so that is not related


sdk/pom.xml
```diff
@@ -56,7 +56,7 @@
       <groupId>io.horizen</groupId>
       <artifactId>zendoo-sc-cryptolib</artifactId>
       <!-- 0.6.0-SNAPSHOT2 tag -->
```
i-Alex, line 58: can be removed now

i-Alex (commented):
Do we need to update `sc_settings.conf` and `sc_evm_settings.conf` genesis block data or is it fine even now?

i-skrypnyk:
@i-Alex 
> Do we need to update sc_settings.conf and sc_evm_settings.conf genesis block data or is it fine even now?

It was able to start, responded to several API calls, forged blocks without any changes

i-Alex (approved):

PR #763, opened 2023-03-03
sc_evm_block_size_limit.py fixes

i-Alex:


qa/sc_evm_block_size_limit.py
```diff
@@ -82,9 +82,9 @@ def run_test(self):
         evm_address_sc1 = remove_0x_prefix(self.evm_address)
         evm_address_sc2 = sc_node_2.wallet_createPrivateKeySecp256k1()["result"]["proposition"]["address"]
 
-        # create huge FTs for filling up the MC ref data. This constant allows us to have tx with size ~96K, a little
-        # below the 100K MC limit
-        outputs_in_ft = 1050
+        # create huge FTs for filling up the MC ref data. This constant allows us to have tx with size ~90K, a little
+        # below the 100K MC limit and moreover stay within the max FT number CommTree limit of <4096 FT per sidechain
```
alsala, line 86: Not <= 4096 because we must not use all tree leaves, correct? BTW: I tested with `outputs_in_ft=1024` and just 3 out of 4 FTs are included in the MC block, as expected.

alsala (approved):

qa/sc_evm_block_size_limit.py
i-Alex: ok good, i've tried as well. Works as expected.

qa/sc_evm_block_size_limit.py
```diff
@@ -178,7 +178,7 @@ def run_test(self):
                 sc_best_block['mainchainBlockReferencesData'][i]['sidechainRelatedAggregatedTransaction']['size'])
             assert_equal(len(
                 sc_best_block['mainchainBlockReferencesData'][i]['sidechainRelatedAggregatedTransaction']['newBoxes']),
-                         outputs_in_ft * ft_in_block)  # = 1050 * 8
+                         outputs_in_ft * ft_in_block)  # = 1000 * 8
```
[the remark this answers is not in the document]
i-Alex: fixed

PR #764, opened 2023-03-06
Merge sender and receiver sc2sc validators

paologalligit:
Jira reference [ticket](https://horizenlabs.atlassian.net/browse/SDK-763)


sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -11,50 +11,59 @@ import com.horizen.certnative.BackwardTransfer
...
     with ModifierValidation[SidechainBlock]
     with SidechainTypes
     with NodeState
     with SparkzLogging
     with UtxoMerkleTreeView
-    with NetworkParamsUtils
-{
+    with NetworkParamsUtils {
 
+  protected val crossChainValidatorContainer: CrossChainValidatorContainer = new CrossChainValidatorContainer(
```
paolocappelletti, line 60: i would keep it simpler: we are defining a key-value structure but anyone wil be able to add more entries from outside the state
We can instead use a similar structure as AbstractSidechainNodeViewHolder.semanticBlockValidators :  

- define a trait Sc2ScBlockValidator similar to com.horizen.validation.SemanticBlockValidator with a method validate(block: PMOD)
-  here initialize a Seq of the above trait, with two implementations: one for output crosschainmessages and one for input

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -181,18 +191,30 @@ class SidechainState private[horizen] (stateStorage: SidechainStateStorage,
...
+      validate(tx, consensusEpochNumber, modWithdrawalEpochInfo.epoch).get
+      tx match {
+        case transaction: AbstractCrossChainRedeemTransaction =>
+          crossChainValidatorContainer.validateReceivingCrossChain(
+            new CrossChainRedeemMessageBodyToValidate(transaction)
+          )
+        case _ =>
+      }
+    })
+    crossChainValidatorContainer.validateSenderCrossChain(
```
paolocappelletti, line 213: with the above suggestion this piece becomes something like:
sc2scBlockValidators.foreach ( val.validate(mod))

sdk/src/main/scala/com/horizen/SidechainState.scala
paologalligit: That's a good point, gonna fix it

sdk/src/main/scala/com/horizen/SidechainState.scala
```diff
@@ -11,52 +11,58 @@ import com.horizen.certnative.BackwardTransfer
...
     with UtxoMerkleTreeView
-    with NetworkParamsUtils
-{
-
-
+    with NetworkParamsUtils {
   override type NVCT = SidechainState
+  type PMOD = SidechainBlock
+
+  protected val crossChainValidators: Seq[CrossChainValidator[PMOD]] = Seq(
```
MarcoOl94, line 62: The idea is to pass the custom validators from the Application side. I think we need to add an additional injected object inside the **SidechainAppModule** like "Sc2ScMessageValidators" which is a `Seq[CrossChainValidator]` and pass it to the **SidechainState** constructor.

sdk/src/main/scala/com/horizen/validation/crosschain/sender/AbstractCrossChainMessageValidator.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.horizen.params.NetworkParams
+import com.horizen.sc2sc.{CrossChainMessageHash, Sc2ScConfigurator}
+import com.horizen.storage.SidechainStateStorage
+import com.horizen.transaction.MC2SCAggregatedTransaction
+import com.horizen.validation.crosschain.CrossChainValidator
+import com.horizen.{SidechainState, SidechainTypes}
+
+import scala.collection.JavaConverters._
+
+class AbstractCrossChainMessageValidator(
```
MarcoOl94, line 15: This class is extended by the custom message validator from the application side. As we discussed we can add a **protected applicationValidate** that is called inside the validate and it is implemented by the custom validator.

MarcoOl94 (changes requested):

paolocappelletti (approved):

PR #766, opened 2023-03-06
Fix RPC parameter types

JcBernack:
- refactored `TransactionArgs` and `FilterQuery` to be immutable as much as possible
- added more validation to both types
- this also fixes a `NullPointerException` caused by optional arguments being omitted, e.g. omitting `value` from an RPC call to `eth_sendTransaction`

### DIFF
--- a/sdk/src/main/java/io/horizen/account/api/rpc/types/TransactionArgs.java
+++ b/sdk/src/main/java/io/horizen/account/api/rpc/types/TransactionArgs.java
@@ -1,6 +1,7 @@
 package io.horizen.account.api.rpc.types;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import io.horizen.account.api.rpc.handler.RpcException;
 import io.horizen.account.api.rpc.utils.RpcCode;
 import io.horizen.account.api.rpc.utils.RpcError;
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/io/horizen/account/api/rpc/types/TransactionArgs.java
```diff
@@ -63,54 +108,45 @@ public EthereumTransaction toTransaction(NetworkParams params) throws RpcExcepti
                 String.format("invalid chainID: got %d, want %d", chainId, saneChainId)
             ));
         }
+        if (nonce == null) {
+            // TODO: should automatically use current nonce from "latest" state
```
i-Alex, line 112: the `TODOs` in this method are not very clear to me.
Do we need this to be changed now? how critical is it?

JcBernack: I added the Todos while going through the equivalent implementation in GETH.

I don't think this is critical at all because it is only relevant when the node is signing a transaction, i.e. `eth_signTransaction` or `eth_sendTransaction`, which no enduser will ever use. Also, most tools will fill all these parameters anyway so we don't need to auto-fill them, e.g. a tool will likely call `eth_estimateGas` itself and then put the `gas` value, instead of omitting `gas` and expecting the node to estimate gas usage.

JcBernack: I just added the checks and exceptions to prevent `NullPointerExceptions` down the road, because e.g. `EthereumTransaction.gasLimit` is `null`. The deployment issue in YUMA currently is caused by this.

i-Alex: thanks, now it is clear.

i-Alex (approved):

PR #768, opened 2023-03-07
Added NewExecTransactionsEvent management

saratnt:


sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
```diff
@@ -245,4 +253,4 @@ object AccountNodeViewHolderRef {
 
 }
 
-case class MempoolReAddedTransactions[T <: SidechainTypes#SCAT](readdedTxs: Seq[T]) extends NodeViewHolderEvent
\ No newline at end of file
+case class NewExecTransactionsEvent(newExecTxs: Iterable[SidechainTypes#SCAT]) extends NodeViewHolderEvent
```
i-Alex, line 256: let's put this class as a part of `object AccountSidechainNodeViewHolder`

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -334,34 +337,31 @@ class MempoolMap(
...
 
-    if (readdedTxs.size > this.size)
-      // If the number of the transactions to be readded exceed the mempool size, we just return the mempool
-      // because the mempool now contains only transactions taken from the removed blocks because are newer than
-      // the transactions previously contained
-      txCache.values.toSeq
-    else
-      readdedTxs
+    //newExecTxs is not changed after checkMempoolSize because the event is sent even if the corresponding tx is evicted
+    newExecTxs
```
i-Alex, line 358: So, if transaction reentered the meempool from the reverted block or was promoted (was present before as non-executable) we will return it in any case?
But what if it was then removed from the mempool?

sdk/src/main/scala/io/horizen/account/state/AccountEventNotifier.scala
```diff
@@ -0,0 +1,13 @@
+package io.horizen.account.state
+
+import io.horizen.SidechainTypes
+
+
+trait AccountEventNotifier {
+  def sendNewExecTxsEvent(listOfNewExecTxs: Iterable[SidechainTypes#SCAT])
```
i-Alex, line 7:
```suggestion
  def sendNewExecTxsEvent(listOfNewExecTxs: Iterable[SidechainTypes#SCAT]): Unit
```

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServer.scala
```diff
@@ -47,7 +47,7 @@ class WebSocketAccountServer(wsPort: Int)
       websocket.onSuccessfulTransaction(tx)
     case ChangedVault(_) =>
       websocket.onChangedVault()
-    case MempoolReAddedTransactions(readdedTxs: Seq[EthereumTransaction]) =>
+    case NewExecTransactionsEvent(readdedTxs: Seq[EthereumTransaction]) =>
       websocket.onMempoolReaddedTransaction(readdedTxs)
```
i-Alex, line 51: I think, the websocket part must be renamed as well.
@MarcoOl94

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
saratnt: In theory, txs evicted are are like any other valid txs added to the mempool and later evicted because the mempool is too big, there are no events for this. In this case this is almost at the same time, but it is consistent

sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
saratnt: Done

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServer.scala
saratnt: Done

sdk/src/main/scala/io/horizen/account/state/AccountEventNotifier.scala
saratnt: Done

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
i-Alex: ok

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServer.scala
i-Alex: i see the same as before.
My expectation to rename `websocket.onMempoolReaddedTransaction` -> `websocket.onNewExecTransactions`
does it have sense?

saratnt: Sorry, I misunderstood the comment. Now it should be correct.

i-Alex (approved):

MarcoOl94 (approved):

PR #770, opened 2023-03-08
St/evm 787

saratnt:


sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -915,21 +915,21 @@ class EthService(
   }
 
   @RpcMethod("zen_getForwardTransfers")
-  def getForwardTransfers(hash: Hash): ForwardTransfersView = {
+  def getForwardTransfers(tag: String): ForwardTransfersView = {
     applyOnAccountView { nodeView =>
       nodeView.history
-        .getStorageBlockById(bytesToId(hash.toBytes))
+        .getStorageBlockById(getBlockIdByHashOrTag(nodeView, tag))
```
i-Alex, line 921: actually we don't want to use `tags`, only hash or height
same for zen_getFeePayments

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
saratnt: Done

sdk/src/main/resources/account/rpc/api/eth_openrpc.json
```diff
@@ -357,7 +343,7 @@
 					"name": "Block",
 					"required": true,
 					"schema": {
-						"title": "Block number or tag",
+						"title": "Block number, block hash or tag",
```
i-Alex, line 346: update the description, remove `tag`
check other places

i-Alex (changes requested):

sdk/src/main/resources/account/rpc/api/eth_openrpc.json
saratnt: Done

i-Alex (approved):

PR #772, opened 2023-03-09
EVM-549 - txpool _contentFrom and txpool _inspect

davehorizen:


sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -125,6 +140,23 @@ class EthService(
     )
   }
 
+  @RpcMethod("txpool_contentFrom")
+  def txpoolContent(from: String): TxPoolContent = applyOnAccountView { nodeView =>
```
i-Alex, line 144: we can use `Address` type directly.
In case the input is invalid (by structure or length) it will throw exception during parsing

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -291,6 +291,24 @@ class MempoolMap(
     txsMap
   }
 
+  def mempoolTransactionsMapFrom(executable: Boolean, requestedFrom: AddressProposition): TrieMap[SidechainTypes#SCP, TxByNonceMap] = {
```
i-Alex, line 294: why do we return map here?
i have a feeling the we do something very inefficient.
Please, contact @saratnt

i-Alex (changes requested):
In general looks good, but I think that we may get data from mempool in more efficient way.

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -117,6 +115,23 @@ class EthService(
       }
     )
 
+  private def convertMempoolMapInspect(poolMap: TrieMap[SidechainTypes#SCP, MempoolMap#TxByNonceMap]) =
+    CollectionConverters.mapAsJavaMap(
+      poolMap.map { case (proposition, txByNonce) =>
+        new Address(proposition.bytes()) -> CollectionConverters.mapAsJavaMap(
```
saratnt, line 121: I think it is more efficient casting `proposition` to Addressproposition and calling directly `address()`

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -291,6 +291,24 @@ class MempoolMap(
...
+    val txsMap = TrieMap.empty[SidechainTypes#SCP, TxByNonceMap]
+    val mempoolIdsMap = if (executable)
+      executableTxs
+    else
+      nonExecutableTxs
+    for ((from, nonceIdsMap) <- mempoolIdsMap) {
+      if(from.equals(requestedFrom)) {
+        val nonceTxsMap: mutable.TreeMap[BigInteger, SidechainTypes#SCAT] = new mutable.TreeMap[BigInteger, SidechainTypes#SCAT]()
+        for ((txNonce, txId) <- nonceIdsMap) {
+          nonceTxsMap.put(txNonce, getTransaction(txId).get)
```
saratnt, line 304: I wonder if the function transforming EthereumTransaction to TxPoolTransaction cannot be passed here, to avoid iterating twice on the map. For txs belonging to a single account it is no much of a difference but when all the mempool txs are request it may be important.

saratnt (changes requested):

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
davehorizen: `from` parameter type change to `Address`
I also added in the python test an rpc call to `txpool_contentFrom` with an invalid input to check the exception expected

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
davehorizen: Method updated, as reported in the `go-ethereum` implementation (https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L190) the `txpool_contentFrom` has a different interface than the `txpool_content` method.
`txpool_contentFrom` expose the data grouping them by nonce and transaction details (and not address/nonce/transaction)

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -291,6 +291,24 @@ class MempoolMap(
     txsMap
   }
 
+  def mempoolTransactionsMapFrom(executable: Boolean, requestedFrom: AddressProposition): TrieMap[SidechainTypes#SCP, TxByNonceMap] = {
+    val txsMap = TrieMap.empty[SidechainTypes#SCP, TxByNonceMap]
+    val mempoolIdsMap = if (executable)
+      executableTxs
+    else
+      nonExecutableTxs
+    for ((from, nonceIdsMap) <- mempoolIdsMap) {
```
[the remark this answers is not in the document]
davehorizen: method update using the `getOrElse` method instead of iterating on the mempool ids map

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
davehorizen: outdated because the method was removed by a code refactor

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
davehorizen: Yes, good point! I proceed to do an extensive refactor on these methods:
- declaring the map with the TxPoolTransactions in the `MempoolMap` class
- using Scala object as return objects of the RPC methods
in this way no double iterations are needed to cast the object from Scala to Java or to declare the correct object for the transaction data.

The same was done for the `txpool_inspect` method moving the response generation in the `MempoolMap` class
Unit tests updated as well

sdk/src/main/scala/io/horizen/account/api/rpc/types/TxPoolContent.scala
```diff
@@ -0,0 +1,14 @@
...
+import io.horizen.evm.Address
+import io.horizen.json.Views
+
+import java.math.BigInteger
+import scala.collection.concurrent.TrieMap
+import scala.collection.mutable
+
+@JsonView(Array(classOf[Views.Default]))
+class TxPoolContent(
+    val pending: TrieMap[Address, mutable.SortedMap[BigInteger, TxPoolTransaction]],
```
i-Alex, line 13: In the doc https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool I see that nonce is serialized as decimal. In our case we serialize it as hex (BigInteger is always represented as hex)

sdk/src/main/scala/io/horizen/account/api/rpc/types/TxPoolContentFrom.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import com.fasterxml.jackson.annotation.JsonView
+import io.horizen.json.Views
+
+import java.math.BigInteger
+import scala.collection.mutable
+
+@JsonView(Array(classOf[Views.Default]))
+class TxPoolContentFrom(
+    val pending: mutable.SortedMap[BigInteger, TxPoolTransaction],
```
i-Alex, line 11: maybe here as well: nonce is decimal.
Note: I can't find in the docs proper example.

sdk/src/main/scala/io/horizen/account/api/rpc/types/TxPoolInspect.scala
```diff
@@ -0,0 +1,14 @@
...
+import io.horizen.evm.Address
+import io.horizen.json.Views
+
+import java.math.BigInteger
+import scala.collection.concurrent.TrieMap
+import scala.collection.mutable
+
+@JsonView(Array(classOf[Views.Default]))
+class TxPoolInspect(
+    val pending: TrieMap[Address, mutable.SortedMap[BigInteger, String]],
```
i-Alex, line 13: be sure that the map is serialzied like `563662: "0x958c1fa64b34db746925c6f8a3dd81128e40355e: 1051546810000000000 wei + 90000 gas × 20000000000 wei",`
without additional object `{}`

saratnt (approved):

sdk/src/main/scala/io/horizen/account/api/rpc/types/TxPoolContentFrom.scala
davehorizen: Same as `txpool_content` as reported in the comment above

sdk/src/main/scala/io/horizen/account/api/rpc/types/TxPoolInspect.scala
davehorizen: checked, it serialized the response without additional object `{}`
The nonce is serialized as string with decimal representation

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -275,22 +276,71 @@ class MempoolMap(
...
+      txsMap.put(from.asInstanceOf[AddressProposition].address(), retrieveTxPoolByNonceMap(nonceIdsMap))
+    }
+    txsMap
+  }
+
+  // method used by the txpool namespace rpc methods, it retrieve a map of executable or non-executable TxPoolTransaction
+  // ordered by nonce filtering by from address
+  def mempoolTransactionsMapFrom(executable: Boolean, fromAddress: Address): mutable.SortedMap[BigInteger, TxPoolTransaction] = {
+    val mempoolIdByNonceMapFrom = if (executable)
+      executableTxs.getOrElse(new AddressProposition(fromAddress), mutable.SortedMap.empty[BigInteger, ModifierId])
```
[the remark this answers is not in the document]
davehorizen: Even if more readable unfortunately this implementation can't be used because the `nonEmpty` method can't be applied on the `mempoolIdByNonceMapFrom` because it is an Option of object and not the direct object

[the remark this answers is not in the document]
i-Alex: @saratnt we can't relay on the fact that `executableTxs` or `nonExecutableTxs` have a record for `fromAddress`
So the suggested approach may throw an unexpected exception.
Am I right?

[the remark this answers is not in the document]
saratnt: @davehorizen Option actually has nonEmpty method, but if you prefer you can use `isDefined` instead, it is the same
@i-Alex  nonExecutableTxs.get(new AddressProposition(fromAddress)) returns an Option that, if notEmpty == true , using get won't throw an exception.
Maybe it is more clear this way:

```
    val mempoolIdByNonceMapFrom = if (executable)
      executableTxs.get(new AddressProposition(fromAddress))
    else
      nonExecutableTxs.get(new AddressProposition(fromAddress))

    mempoolIdByNonceMapFrom match {
      case Some(map) => retrieveTxPoolByNonceMap(map)
      case None => mutable.SortedMap.empty[BigInteger, TxPoolTransaction]
    }

```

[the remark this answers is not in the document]
i-Alex: ok, clear.
@davehorizen let's do as Sara suggested.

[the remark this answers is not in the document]
i-Alex: looks fine

PR #774, opened 2023-03-09
EVM-755 - stop endpoint does not work properly in EVM app

i-skrypnyk:


qa/sc_node_termination_during_sync.py
```diff
@@ -0,0 +1,151 @@
...
+    Create 997 blocks (this number is not important)
+    Connect sc node 2 to sc node 1
+    When sync process started, stop sc node 2
+    Verify node is terminated
+    Restart sc node 2, reconnect it to sc node 2
+    Verify that node is able to sync all the remaining blocks
+"""
+
+
+class BigBlockTest(SidechainTestFramework):
```
i-Alex, line 34: let's rename the class.
Moreover add this test to the utxo list in `run_sc_tests.sh`

sdk/src/main/scala/io/horizen/AbstractSidechainApp.scala
```diff
@@ -70,6 +70,7 @@ abstract class AbstractSidechainApp
 
   private val closableResourceList = mutable.ListBuffer[AutoCloseable]()
   protected val sidechainTransactionsCompanion: DynamicTypedSerializer[TX, TransactionSerializer[TX]]
+  protected val terminationTimeout: FiniteDuration = Duration(30, TimeUnit.SECONDS)
```
i-Alex, line 73: I think in the most of the cases 30seconds must be ok.
But what if the node owner set in the config file the `settings.sparkzSettings.restApi.timeout` duration bigger than 30.
If I'm not wrong, a lot of `await` cases depend on that parameter (forger, submitter, etc)

To be even more safe we may set `terminationTimeout` as a `Min(restApi.timeout, 30s)`
What do you think?

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/AbstractSidechainApp.scala
i-skrypnyk: After some more discussion, we came to the conclusion that there is no healthy situation where we want to wait for node termination for more than 30 seconds. Even if the rest timeout is bigger, the fact that some actor has to wait until it's expiration indicates some problematic behavior.

i-Alex (approved):

PR #776, opened 2023-03-09
ApplicationApiRoute bound to Utxo model

alsala:
`ApplicationApiRoute` and `ApplicationApiGroup` classes are strictly bound to Utxo model, but used both in Utxo and Account models. This class is made generic and an account model specific implementation is provided as well.

https://horizenlabs.atlassian.net/browse/EVM-679


sdk/src/main/scala/io/horizen/account/AccountSidechainApp.scala
```diff
@@ -148,6 +146,8 @@ class AccountSidechainApp @Inject()
   // Init Sync Status actor
   val syncStatusActorRef: ActorRef = SyncStatusActorRef("SyncStatus", sidechainSettings, nodeViewHolderRef, params, timeProvider)
 
+  override lazy val applicationApiRoutes: Seq[AccountApplicationApiRoute] = customApiGroups.asScala.map(apiRoute => AccountApplicationApiRoute(settings.restApi, apiRoute, nodeViewHolderRef))
```
i-Alex, line 149:
```suggestion
  override lazy val applicationApiRoutes: Seq[ApiRoute] = customApiGroups.asScala.map(apiRoute => AccountApplicationApiRoute(settings.restApi, apiRoute, nodeViewHolderRef))
```
I assume that Seq type must remain the same as in the base class.

sdk/src/main/scala/io/horizen/account/api/http/route/AccountApplicationApiRoute.scala
```diff
@@ -0,0 +1,75 @@
...
+  private def sendMessageToSidechainNodeView[T, R](messageToSend: T): R = {
+    try {
+      val res = (sidechainNodeViewHolderRef ? messageToSend).asInstanceOf[Future[R]]
+      val result = Await.result[R](res, settings.timeout)
+      result
+    }
+    catch {
+      case e: Exception => throw new Exception(e)
+    }
+  }
```
i-Alex, line 74: can we put all this methods to the base class?
Why do we have the same code here and in utxo class `SidechainApplicationApiRoute`?

sdk/src/main/scala/io/horizen/utxo/SidechainApp.scala
```diff
@@ -214,6 +213,8 @@ class SidechainApp @Inject()
 
   val boxIterator: BoxIterator = backupStorage.getBoxIterator
 
+  override lazy val applicationApiRoutes: Seq[SidechainApplicationApiRoute] = customApiGroups.asScala.map(apiRoute => SidechainApplicationApiRoute(settings.restApi, apiRoute, nodeViewHolderRef))
```
i-Alex, line 216:
```suggestion
  override lazy val applicationApiRoutes: Seq[ApiRoute] = customApiGroups.asScala.map(apiRoute => SidechainApplicationApiRoute(settings.restApi, apiRoute, nodeViewHolderRef))
```
same as in Account class

sdk/src/test/java/io/horizen/utxo/api/http/SimpleCustomApi.java
```diff
@@ -34,7 +34,7 @@ public List<Route> getRoutes() {
     private ApiResponse getNSecretsFunction(GetSecretRequest ent) {
         try {
             RespAllSecret resp = new RespAllSecret();
-            List<Secret> res = new ArrayList<>(getFunctionsApplierOnSidechainNodeView().applyFunctionOnSidechainNodeView(v -> v.getNodeWallet().allSecrets()));
+            List<io.horizen.secret.Secret> res = new ArrayList<>(getFunctionsApplierOnSidechainNodeView().applyFunctionOnSidechainNodeView(v -> (v).getNodeWallet().allSecrets()));
```
i-Alex, line 37: why do we need any changes in this line?

i-Alex (changes requested):
Looks good, minor points left

sdk/src/main/scala/io/horizen/account/AccountSidechainApp.scala
alsala: Right, done

sdk/src/main/scala/io/horizen/utxo/SidechainApp.scala
alsala: Done

sdk/src/test/java/io/horizen/utxo/api/http/SimpleCustomApi.java
alsala: No need indeed

i-Alex (approved):

saratnt (approved):

PR #777, opened 2023-03-09
RPC transaction autofill

JcBernack:
Automatically fill omitted transaction parameters when using the RPC methods `eth_signTransaction` or `eth_sendTransaction`:
- [x] `type` if only `gasPrice` is set create a legacy tx, otherwise dynamic-fee tx
- [x] `maxPriorityFeePerGas` use gas tip suggestion
- [x] `maxFeePerGas` use `2 * baseFee + maxPriorityFeePerGas`
- [x] `nonce` use latest nonce from mempool or state
- [x] `gas` use gas estimation

Not ideal, but out of scope here:
- fields of `TransactionArgs` should be immutable, but can't because of gas estimation
- separation of `EthService` / `Backend` is messy, but requires a bigger refactoring

See [EVM-777](https://horizenlabs.atlassian.net/browse/EVM-777)

[EVM-777]: https://horizenlabs.atlassian.net/browse/EVM-777?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sdk/src/main/java/io/horizen/account/api/rpc/types/TransactionArgs.java
```diff
@@ -93,66 +93,101 @@ public TransactionArgs(
...
+            ? EthereumTransactionType.LegacyTxType
+            : EthereumTransactionType.DynamicFeeTxType;
+
+        if (saneType == EthereumTransactionType.DynamicFeeTxType) {
+            // if omitted use suggested tip cap
+            if (maxPriorityFeePerGas == null) maxPriorityFeePerGas = Backend.suggestTipCap(history);
+            // if omitted use 2 * baseFee + maxPriorityFeePerGas
+            if (maxFeePerGas == null) {
+                var baseFee = history.bestBlock().header().baseFee();
+                maxFeePerGas = BigInteger.TWO.multiply(baseFee).add(maxPriorityFeePerGas);
```
i-Alex, line 157: we may get value from AccountState using `state.getNextBaseFee()` seems it is lighter than get the full block from History ledger.
Beside that do we need any other calculations?

i-Alex (approved):

PR #785, opened 2023-03-15
EVM-657 - EIP1559 transactions manage by RPC txpool namespace

davehorizen:
details reported in the task EMV-657


sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -79,5 +79,38 @@ public EthereumTransactionView(EthereumTransaction tx, EthereumReceipt receipt,
         blockNumber = BigInteger.valueOf(receipt.blockNumber());
         transactionIndex = BigInteger.valueOf(receipt.transactionIndex());
     }
+
+    // constructor used for the pending transactions
+    public EthereumTransactionView(EthereumTransaction tx) {
+        blockHash = Hash.ZERO;
```
i-Alex, line 85: are you sure that blockHash for mempool tx must be `Hash.ZERO`, but not `null` as for other fields?
I see in eth docs `0000.0000` result, but in other docs it is `null`.
Moreover. in Geth it seems to return `null`:
```
      if blockHash != (common.Hash{}) {
		result.BlockHash = &blockHash
		result.BlockNumber = (*hexutil.Big)(new(big.Int).SetUint64(blockNumber))
		result.TransactionIndex = (*hexutil.Uint64)(&index)
	}
```

sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -79,5 +79,38 @@ public EthereumTransactionView(EthereumTransaction tx, EthereumReceipt receipt,
         blockNumber = BigInteger.valueOf(receipt.blockNumber());
         transactionIndex = BigInteger.valueOf(receipt.transactionIndex());
     }
+
+    // constructor used for the pending transactions
+    public EthereumTransactionView(EthereumTransaction tx) {
```
i-Alex, line 84: I see a lot of code duplication between this constructor and the one above for inchain txs.
the difference is only in `EthereumReceipt receipt, BigInteger baseFee`.
Let's unify the code of this 2 constructors.

sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumTransactionView.java
```diff
@@ -79,5 +79,38 @@ public EthereumTransactionView(EthereumTransaction tx, EthereumReceipt receipt,
...
+        blockNumber = null;
+        transactionIndex = null;
+        type = BigInteger.valueOf(tx.version());
+        nonce = tx.getNonce();
+        from = tx.getFromAddress();
+        to = tx.getToAddress();
+        value = tx.getValue();
+        input = tx.getData();
+        gas = tx.getGasLimit();
+        gasPrice = tx.getGasPrice();
```
i-Alex, line 95: do we need to account `baseFee` here:
`gasPrice = tx.getEffectiveGasPrice(baseFee);`
?

i-Alex (changes requested):

saratnt (approved):
I agree with Sasha's comments, a part from the one related tot the gasPrice. For the rest, LGTM.

sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumTransactionView.java
i-Alex: let's check it. Geth may get the pending block base fee (it is known)

JcBernack: Shouldn't the transactions look exactly like what was submitted?
- `gasPrice` only for legacy TX
- omit for EIP1559 and only set priority fee and fee cap?

Should check the behavior of GETH though.

davehorizen: In the GETH implementation the transaction `gasFeeCap` value is used for the `gasPrice` field of EIP1559 transactions: https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1353
The `gasFeeCap` corresponds to our `EthereumTransaction.maxFeePerGas`, containing information on the maximum price per unit of gas that the user is willing to pay for his transaction.

Following the flow of the `txpool_content` method, starting from [here](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L163), we will see that for every pending transaction an `RPCTransaction` is retrieved with the function [NewRPCPendingTransaction](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1359)
`NewRPCPendingTransaction` internally calls the function `newRPCTransaction` that will finally assign, for the pending transactions, the `gasPrice` field with the `gasFeeCap` value.

Moreover, in the `NewRPCPendingTransaction` function the block number and the base fee are retrieved from the current (latest) blockchain block.
These values are passed to the `newRPCTransaction` but they are never used because the block hash in our case is only initialized (ex. [here](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1348))
It seems that the gas price for pending transactions is expected to be calculated based on the last block but in reality is not.

PR #786, opened 2023-03-16
Added support for all RPC commands inside the websocket

MarcoOl94:


sdk/src/main/scala/io/horizen/account/api/rpc/service/RpcProcessor.scala
```diff
@@ -0,0 +1,63 @@
...
+import io.horizen.account.api.rpc.response.RpcResponseError
+import io.horizen.account.api.rpc.utils.{RpcCode, RpcError}
+import io.horizen.account.serialization.EthJsonMapper
+import sparkz.util.SparkzLogging
+
+import scala.jdk.CollectionConverters.asScalaIteratorConverter
+import scala.util.{Failure, Success, Try}
+
+
+object RpcProcessor extends SparkzLogging {
```
i-Alex, line 15: I still vote to keep `RpcProcessor` as a class, not as an object.
In the AccountSidechainApp create an instance of this class and then pass as an argument of:
1) `AccountEthRpcRoute` constructor, to let this class use RpcProcessor explicitly. Implicit dependencies are not good to have.
2) `WebSocketAccountServerRef` constructor, and then inside do the same trick as for NVHactor.

sdk/src/main/scala/io/horizen/account/api/rpc/service/RpcProcessor.scala
```diff
@@ -0,0 +1,63 @@
...
+      EthJsonMapper.serialize(responses)
+    } else {
+      EthJsonMapper.serialize(responses.head)
+    }
+
+    log.trace(s"RPC message response << $json")
+    json
+  }
+
+  def getClientVersion: String = {
```
i-Alex, line 51: do we really need to have this method here?
Let's move it to a separate utility class inside `package io.horizen.account.api.rpc.service`

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServerEndpoint.scala
```diff
@@ -307,4 +309,13 @@ private object WebSocketAccountServerEndpoint extends SparkzLogging {
     })
   }
 
+  def sendRpc(rpcResponse: Object, session: Session): Unit = {
```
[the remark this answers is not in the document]
i-Alex: beside that, let's extend py tests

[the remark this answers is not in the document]
MarcoOl94: Done

sdk/src/main/scala/io/horizen/account/api/rpc/service/RpcProcessor.scala
MarcoOl94: Done

sdk/src/main/scala/io/horizen/account/api/rpc/service/RpcProcessor.scala
MarcoOl94: Done

i-Alex (approved):

PR #787, opened 2023-03-16
As/issue evm823

alsala:
Using `BigInteger` for signature `v, r, s` objects instead of `byte []` type. Allowing r and s to be shorter that 32 bytes. Added validity checks for `v, r, s` similarly to what geth does.


sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
```diff
@@ -2,86 +2,129 @@
...
-import io.horizen.utils.BytesUtils;
+import java.math.BigInteger;
+
 
 @JsonView(Views.Default.class)
 public final class SignatureSecp256k1 implements ProofOfKnowledge<PrivateKeySecp256k1, AddressProposition> {
 
     @JsonProperty("v")
-    private final byte[] v;
+    @JsonSerialize(using = HexNoPrefixBigIntegerSerializer.class)
```
i-Alex, line 21: As was discussed @JsonSerialize can be removed from this class

sdk/src/main/java/io/horizen/account/utils/EthereumTransactionEncoder.java
```diff
@@ -64,9 +61,9 @@ private static RlpList LegacyTransaction2RlpList(EthereumTransaction tx, boolean
             result.add(RlpString.create(s));
         } else {
             if (tx.isEIP155()) {
-                result.add(RlpString.create(EthereumTransactionUtils.convertToBytes(tx.getChainId())));
-                result.add(RlpString.create(new byte[] {}));
-                result.add(RlpString.create(new byte[] {}));
+                result.add(RlpString.create(BigInteger.valueOf(tx.getChainId())));
+                result.add(RlpString.create(BigInteger.ZERO));
```
i-Alex, line 65: just a double check: are we sure that ZERO byte representation is empty array?

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -251,7 +252,10 @@ class EthService(
     applyOnAccountView { nodeView =>
       val secret = getFittingSecret(nodeView.vault, nodeView.state, sender, BigInteger.ZERO)
       val signature = secret.sign(messageToSign)
-      signature.getR ++ signature.getS ++ signature.getV
+      // TODO check order: why V is the last contribution?
```
i-Alex, line 255: I think it is because of historical reasons first of all. In geth it is the same.
Moreover, I think we have an issue here. Because R and S must be padded to 32 bytes to make the result predictable, otherwise how to understand how to parse?

In geth:
```
// Serialize the signature
	rbytes, sbytes := sigdata.Signature.R.Bytes(), sigdata.Signature.S.Bytes()
	sig := make([]byte, 65)
	copy(sig[32-len(rbytes):32], rbytes)
	copy(sig[64-len(sbytes):64], sbytes)

	// Recover the V value.
	sig, err = makeRecoverableSignature(hash, sig, sigdata.PublicKey)
	if err != nil {
		return nil, err
	}
```

sdk/src/test/scala/io/horizen/account/state/ForgerStakeMsgProcessorTest.scala
```diff
@@ -912,7 +912,7 @@ class ForgerStakeMsgProcessorTest
       // bytes are not considered when decoding
       val decodingOk = RemoveStakeCmdInputDecoder.decode(data)
       val decodingBad = RemoveStakeCmdInputDecoder.decode(badData2)
-      assertArrayEquals(decodingOk.signature.getV, decodingBad.signature.getV)
+      assertArrayEquals(decodingOk.signature.getV.toByteArray, decodingBad.signature.getV.toByteArray)
```
i-Alex, line 915:
```suggestion
      assertEquals(decodingOk.signature.getV, decodingBad.signature.getV)
```

i-Alex (changes requested):

sdk/src/test/java/io/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
```diff
@@ -351,4 +352,42 @@ public void checkUnsignedTx() {
...
+    }
+
+    @Test
+    public void checkIssueEVM823()  {
+        // The tx has a signature r value whose byte array representation is 31 bytes long (not 32), and this should be supported
+        byte[] b = BytesUtils.fromHexString("f8878214920783011170941050072833849ff9bf49a55ff15d40f6d89651e880a440d097c30000000000000000000000000f6cf5090c089dfbd2c37eb58a70abad83dd79d2820d209f52759db97387354cd02079a382b1a165bac472be1da47a15a748039c6faf18a032bff10a83396a6a4a89b2f18880d6d73f7b6af9952b3f06cacc1283c86f0fa8");
+        Reader reader = new VLQByteBufferReader(ByteBuffer.wrap(b));
+        EthereumTransaction ethTx = EthereumTransactionDecoder.decode(reader);
+        try {
+            ethTx.semanticValidity();
```
i-Alex, line 382: since we don't expect any exception here: let's add a `fail(...)` inside `catch` block

sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
```diff
@@ -2,86 +2,130 @@
...
+
+        if (v.signum() < 0 || r.signum() < 0 || s.signum() < 0) {
+            throw new IllegalArgumentException("Negative v/r/s obj passed in signature data");
+        }
+        // reject upper range of s values (ECDSA malleability)
+        if (s.compareTo(secp256k1halfN) > 0) {
+            throw new IllegalArgumentException("Invalid s obj passed in signature data");
+        }
+
+        if (r.compareTo(secp256k1N) >= 0 || s.compareTo(secp256k1N) >= 0) {
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
```diff
@@ -2,86 +2,130 @@
...
+            throw new IllegalArgumentException("Null obj passed in signature data: " +
+                    "v=" + ((v == null) ? "NULL" : v.toString(16)) +
+            ", r=" + ((r == null) ? "NULL" : r.toString(16)) +
+            ", s=" + ((s == null) ? "NULL" : s.toString(16))  );
         }
-    }
 
-    public SignatureSecp256k1(byte[] v, byte[] r, byte[] s) {
-        verifySignatureData(v, r, s);
+        if (v.bitLength()/8 > 1)
```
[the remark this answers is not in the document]
alsala: Right, fixed. I also added a UT inspired by the test in geth.

sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
```diff
@@ -2,86 +2,130 @@
...
 import io.horizen.account.utils.Secp256k1;
 import io.horizen.evm.Address;
+import io.horizen.json.serializer.HexBigIntegerSerializer;
 import io.horizen.proof.ProofOfKnowledge;
 import io.horizen.proof.ProofSerializer;
 import io.horizen.json.Views;
-import io.horizen.utils.BytesUtils;
+import java.math.BigInteger;
+
+import static io.horizen.account.utils.BigIntegerUInt256.getUnsignedByteArray;
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -241,6 +241,9 @@ public void semanticValidity() throws TransactionSemanticValidityException {
         if (getNonce().signum() < 0)
             throw new TransactionSemanticValidityException(String.format("Transaction [%s] is semantically invalid: " +
                     "negative nonce", id()));
+        if (!BigIntegerUtil.isUint64(getNonce()))
```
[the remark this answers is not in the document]
alsala: No, added.

sdk/src/main/java/io/horizen/account/utils/BigIntegerUInt256.java
```diff
@@ -23,4 +25,22 @@ public BigIntegerUInt256(byte[] bytes) {
...
+        if(barr[0] == 0x00) {
+            return Arrays.copyOfRange(barr, 1, barr.length);
+        } else {
+            return barr;
+        }
+    }
+    // alternative way, but much costly
+    public static byte[] getUnsignedByteArray2(BigInteger bi) {
+        return Numeric.hexStringToByteArray(bi.toString(16));
+    }
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/java/io/horizen/account/utils/EthereumTransactionDecoder.java
```diff
@@ -207,16 +196,15 @@ private static EthereumTransaction RlpList2LegacyTransaction(RlpList rlpList) {
...
         if (v == LOWER_REAL_V || v == (LOWER_REAL_V + 1)) {
             return 0L;
         }
         return (v - CHAIN_ID_INC) / 2;
     }
 
-    private static byte[] getVFromRecId(int recId) {
-        return new byte[]{(byte)(LOWER_REAL_V + recId)};
+    private static BigInteger getVFromRecId(BigInteger recId) {
+        return BigInteger.valueOf(LOWER_REAL_V + recId.intValueExact());
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
alsala: I tried adding this in `ApplicationJsonSerializer` but it overrides the default behavior for other cases, for instance we have `value` and `nonce` of the `BigInteger` type in `EthereumTransaction`, therefore for the time being I left it as it is.

sdk/src/main/java/io/horizen/account/utils/EthereumTransactionEncoder.java
alsala: Yes, in that case the `create() `method returns `new RlpString(EMPTY)`

sdk/src/main/java/io/horizen/account/utils/EthereumTransactionUtils.java
```diff
@@ -16,37 +16,23 @@ private EthereumTransactionUtils() {
...
         BigInteger v = BigInteger.valueOf(x);
-        return v.toByteArray();
+        return BigIntegerUInt256.getUnsignedByteArray(v);
     }
 
-    // w3j way for converting bytes, it works also with generic byte contents (not only Long.BYTES byte arrays)
-    public static long convertToLong(byte[] bytes) {
-        BigInteger bi = Numeric.toBigInt(bytes);
-        return bi.longValueExact();
+    public static long convertToLong(BigInteger val) {
```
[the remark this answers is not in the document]
alsala: Ok, done

sdk/src/main/scala/io/horizen/account/state/ForgerStakeMsgProcessorData.scala
```diff
@@ -178,7 +179,10 @@ case class RemoveStakeCmdInput(
   extends ABIEncodable[StaticStruct] {
 
   override def asABIType(): StaticStruct = {
-    val listOfParams: util.List[Type[_]] = util.Arrays.asList(new Bytes32(stakeId), new Bytes1(signature.getV), new Bytes32(signature.getR), new Bytes32(signature.getS))
+    val t_v = getUnsignedByteArray(signature.getV)
+    val t_r = getUnsignedByteArray(signature.getR)
+    val t_s = getUnsignedByteArray(signature.getS)
```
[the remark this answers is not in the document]
alsala: Fixed, I also added a specific UT for this case. We use both `encode() `and `decode()` methods for this class, for instance we encode it in the http api.

sdk/src/test/scala/io/horizen/account/state/ForgerStakeMsgProcessorTest.scala
alsala: Done

sdk/src/test/java/io/horizen/account/transaction/EthereumTransactionRlpStreamCodecTest.java
alsala: Ok, fixed

sdk/src/main/java/io/horizen/account/proof/SignatureSecp256k1.java
i-Alex: good point, we for sure want other BigInteger values to be serialized in decimal form.

i-Alex (approved):

PR #792, opened 2023-03-20
fix invalid block RPC responses

JcBernack:
The following methods were giving wrong results when the given block was not found, now they return an error instead:
- eth_getBalance
- eth_getTransactionCount
- eth_getCode

Also update and fix unit and python tests.

See [EVM-772](https://horizenlabs.atlassian.net/browse/EVM-772)

[EVM-772]: https://horizenlabs.atlassian.net/browse/EVM-772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
+++ b/sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
@@ -447,42 +447,6 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
   }
 
   /**
-   * Helper for constructing an rpc request. Takes up to two Arrays containing the parameter names, values and the
-   * method name
-   * @param params
-   *   default is empty Array
-   * @param method
-   *   RPC method, default is calling method name
-   * @return
-   *   RpcRequest instance
-   */
-  private[this] def getRpcRequest(
-      params: Array[String] = null,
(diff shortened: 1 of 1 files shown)


sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
```diff
@@ -910,7 +915,12 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
       ("0x1234567890123456789012345678901234567890", "latest", "\"0x\"")
     )
 
-    val invalidCases = Table(("Address", "Tag"), ("0x", "latest"), ("0x1234567890123456789012345678901234567890", ""))
+    val invalidCases = Table(
+      ("Address", "Tag"),
+      ("0x", "latest"),
+      ("0x1234567890123456789012345678901234567890", ""),
+      ("0x1234567891011121314151617181920212223242", "0x1337")
```
i-Alex, line 922: let's add comments for each case explaining why it is considered to be invalid.
Same in other places.

i-Alex (changes requested):

i-Alex (approved):
Looks good!

PR #793, opened 2023-03-21
fix SidechainSettings logging

JcBernack:
- add `SensitiveStringer` trait which omits values from `toString` calls that are annotated with `SensitiveString`
- use `SensitiveString` on all settings classes


i-Alex (commented):
Looks good!
Let's double check that in the places where sensitive part is deserialized we do not return the full string representation in case of any errors (like deserialization failure)

i-Alex (approved):

JcBernack:
found another instance of logging sensitive data and fixed tests

i-Alex (approved):

PR #796, opened 2023-03-23
EVM 685: missing rpc methods - web3_sha3

ddrvar:
This PR includes

- web3_sha3 RPC method
- Unit test for it
- Integration test for it


qa/sc_evm_rpc_web3_sha3.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+from SidechainTestFramework.account.ac_chain_setup import AccountChainSetup
+from test_framework.util import assert_equal, assert_false, assert_true
+
+
+class SCEvmRpcWeb3Sha3(AccountChainSetup):
```
saratnt, line 6: This test should be added in run_sc_tests.sh

saratnt (changes requested):
LGTM! 
My only comment is that I wouldn't create a separate python test but I would have put the same test inside sc_evm_rpc_web3_methods.py.

qa/sc_evm_rpc_web3_sha3.py
i-Alex: As you've suggested, let's put the checks as a part of `sc_evm_rpc_web3_methods`.

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1035,4 +1037,16 @@ class EthService(
       }
     }
   }
+
+  @RpcMethod("web3_sha3")
+  def getSHA3(data: String): String = {
```
i-Alex, line 1042:
```suggestion
  def getSHA3(data: Array[Byte]): String = {
```
we may expect byte array type here, json unmarshaller will automatiically apply `EthByteDeserializer` to this type and process `0x` properly. Note: it allows hex string with or without `0x`.

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
saratnt: I haven't found an explicit requirement about the 0x prefix in input param, but trying web3_sha3 on an ethereum node with an hex string without 0x returns an error ("cannot unmarshal hex string without 0x prefix into Go value of type hexutil.Bytes"). 
I don't know if we should be as "strict" or not, to be honest.

qa/sc_evm_rpc_web3_sha3.py
david-drvar: Done
Integration test moved to `sc_evm_rpc_web3_methods.py` and deleted `sc_evm_rpc_web3_sha3_py`

qa/sc_evm_rpc_web3_methods.py
```diff
@@ -36,6 +42,20 @@ def run_test(self):
...
+        self.validate(sc_node_1.rpc_web3_sha3("68656c6c6f20776f726c64"),
+                      "0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad")
+        self.validate(sc_node_1.rpc_web3_sha3("0xZZXX68656c6c6f20776f726c64"),
+                      "0x8bc0488d7a81d4c07855cf17f63c41b5f6cc6c0d15c764ce43ab668c9e6809ba")
+        self.validate(sc_node_1.rpc_web3_sha3("0x"),
+                      "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
+        self.validate(sc_node_1.rpc_web3_sha3(""),
+                      "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
+        self.validate(sc_node_1.rpc_web3_sha3("text"),
+                      "0x2c7a9a0b269b5b740e242917d5b704ce4329a174526cd76ba1f042dfd88795bb")
```
JcBernack, line 57: These tests should fail.

sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
```diff
@@ -1329,4 +1329,25 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
...
+      ("68656c6c6f20776f726c64", "0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad"),
+      ("ZZXX68656c6c6f20776f726c64", "0x8bc0488d7a81d4c07855cf17f63c41b5f6cc6c0d15c764ce43ab668c9e6809ba"),
+      ("0x", "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"),
+      ("", "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"),
+      ("text", "0x2c7a9a0b269b5b740e242917d5b704ce4329a174526cd76ba1f042dfd88795bb")
+    )
+
+    forAll(validCases) { (input, expectedOutput) =>
+      assert(
+        expectedOutput == ethService.execute(getRpcRequest(paramValues = Array(input), method = method))
```
JcBernack, line 1348: Ok, now I see the issue. You need to merge in `dev`, there have been a lot of changes in the RPC api  and its tests.

This line here should simplify to: `assertJsonEquals(expectedOutput, rpc("web3_sha3", input))` and you need to updated the expected output to include the quotation marks, i.e. `"\"0x0000000000000000000000000000000000000000000000000000000000000000\""`

JcBernack: After merging `dev` all cases without the `0x` prefix will fail. We could use some of them as tests for invalid cases.

JcBernack (changes requested):

sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
david-drvar: Thanks for the info and clearing that up! After the merge with `dev`, the tests are modified and work as expected.

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1027,4 +1029,14 @@ class EthService(
       }
     }
   }
+
+  @RpcMethod("web3_sha3")
+  def getSHA3(data: Array[Byte]): Hash = {
+    try {
+      new Hash(Keccak256.hash(data))
+    } catch {
+      case e: IllegalArgumentException => throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, e.getCause.getMessage))
```
JcBernack, line 1038: Hm, I don't think there is any reasonable way this can fail. I don't think we should catch anything here, if it really does throw it will get turned into an internal server error - which would be correct here.

JcBernack (approved):

qa/sc_evm_rpc_web3_methods.py
```diff
@@ -36,6 +46,20 @@ def run_test(self):
...
 
+        # Check web3_sha3 method
+        self.validate(sc_node_1.rpc_web3_sha3("0x68656c6c6f20776f726c64"),
+                      "0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad")
+        self.validate(sc_node_1.rpc_web3_sha3("0x68656C6C6F20776F726C64"),
+                      "0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad")
+        self.validate(sc_node_1.rpc_web3_sha3("0x"),
+                      "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
+        self.validate(sc_node_1.rpc_web3_sha3("0X68656C6C6F20776F726C64"))
+        self.validate(sc_node_1.rpc_web3_sha3("0x68656c6c6f20776f726c642"))
```
saratnt, line 57: It would be nice to have some comments, explaining why some cases are not valid, at least for non obvious ones.

sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
```diff
@@ -1292,4 +1292,33 @@ class EthServiceTest extends JUnitSuite with MockitoSugar with ReceiptFixture wi
...
+  @Test
+  def eth_getSHA3(): Unit = {
+    val validCases = Table(
+      ("Input", "Expected output"),
+      ("0x68656c6c6f20776f726c64", "\"0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad\""),
+      ("0x68656C6C6F20776F726C64", "\"0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad\""),
+      ("0x", "\"0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470\""),
+    )
+
+    val invalidCases = Table("Input",
```
saratnt, line 1305: It would be nice to have some comments, explaining why some cases are not valid, at least for non obvious ones.

saratnt (approved):
For me it's OK. I have just a couple of really minor comments.

qa/sc_evm_rpc_web3_methods.py
ddrvar: Done

sdk/src/test/scala/io/horizen/account/api/rpc/service/EthServiceTest.scala
ddrvar: Done

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
ddrvar: Agree, error handling removed

i-Alex (approved):

PR #802, opened 2023-03-29
Evm 757 seed length

VolodymyrBabaiev:
Introduced maximum seed length for a key generation of all types.
Introduced minimum seed length for generatedKey and generateAccountKey.

It's safe to use generateVrfKey and generateCertificateSignerKey with an empty seed.


MarcoOl94 (approved):
Do we have this sort of limitation also on the config file property **wallet.seed**?

i-Alex (approved):

PR #803, opened 2023-03-30
Mo/sdk 812

MarcoOl94:
Add an IP whitelist for the account websocket.

### DIFF
--- a/sdk/src/main/scala/io/horizen/SidechainSettings.scala
+++ b/sdk/src/main/scala/io/horizen/SidechainSettings.scala
@@ -26,8 +26,9 @@ case class WebSocketClientSettings(
 
 case class WebSocketServerSettings(
     wsServer: Boolean = false,
-    wsServerPort: Int = 8025
-) extends SensitiveStringer
+    wsServerPort: Int = 8025,
+    wsServerAllowedOrigins: Seq[String] = Seq("127.0.0.1")
+)
 
 case class GenesisDataSettings(
     scGenesisBlockHex: String,
--- a/sdk/src/main/scala/io/horizen/SidechainSettings.scala
+++ b/sdk/src/main/scala/io/horizen/SidechainSettings.scala
@@ -27,7 +27,7 @@ case class WebSocketClientSettings(
 case class WebSocketServerSettings(
     wsServer: Boolean = false,
     wsServerPort: Int = 8025,
-    wsServerAllowedOrigins: Seq[String] = Seq("127.0.0.1")
+    wsServerAllowedOrigins: Seq[String] = Seq()
 )
 
 case class GenesisDataSettings(
--- a/sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountConfigurator.scala
+++ b/sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountConfigurator.scala
@@ -11,7 +11,8 @@ class WebSocketAccountConfigurator() extends ServerEndpointConfig.Configurator w
 
   override def modifyHandshake(config: ServerEndpointConfig, request: HandshakeRequest, response: HandshakeResponse): Unit = {
     val remoteAddress = InetAddress.getByName(request.asInstanceOf[RequestContext].getRemoteAddr)
-    if (!remoteAddress.isAnyLocalAddress && !remoteAddress.isLoopbackAddress && !remoteAddress.isSiteLocalAddress && !WebSocketAccountConfigurator.allowedIPs.contains(remoteAddress.getHostAddress))
+    if ((WebSocketAccountConfigurator.allowedIPs.isEmpty && !remoteAddress.isAnyLocalAddress && !remoteAddress.isLoopbackAddress && !remoteAddress.isSiteLocalAddress) ||
+      (WebSocketAccountConfigurator.allowedIPs.nonEmpty && !WebSocketAccountConfigurator.allowedIPs.contains(remoteAddress.getHostAddress)))
       throw new SecurityException(s"IP ${remoteAddress} is not allowed to connect to the WebSocket")
   }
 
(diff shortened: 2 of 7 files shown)


sdk/src/test/scala/io/horizen/account/websocket/WebSocketAccountServerEndpointTest.scala
```diff
@@ -106,6 +106,19 @@ class WebSocketAccountServerEndpointTest extends JUnitSuite with MockitoSugar wi
...
   }
 
+  @Test
+  def allowedOriginsTest(): Unit = {
+    server = WebSocketAccountServerRef(mockedSidechainNodeViewHolderRef, rpcProcessor, 9035, Seq("127.0.0.2"))
+
+    val cec = ClientEndpointConfig.Builder.create.build
+    val client = ClientManager.createClient
+
+    val countDownController: CountDownLatchController = new CountDownLatchController(1)
```
i-Alex, line 116: why do we need `CountDownLatchController` here?

sdk/src/main/scala/io/horizen/SidechainSettings.scala
```diff
@@ -26,8 +26,9 @@ case class WebSocketClientSettings(
 
 case class WebSocketServerSettings(
     wsServer: Boolean = false,
-    wsServerPort: Int = 8025
-) extends SensitiveStringer
+    wsServerPort: Int = 8025,
+    wsServerAllowedOrigins: Seq[String] = Seq("127.0.0.1")
```
i-Alex, line 30: are we sure about that default ipv4 localhost is enough?
what if the server has only ipv6 configuration or other loopback operations?

maybe we may make default value as `Seq()` and inside `WebSocketAccountConfigurator` class detect this case. Then check it in a more comprehensive way.
For example use InetAddress class methods: `address.isSiteLocalAddress || address.isLoopbackAddress || address.isAnyLocalAddress`

i-Alex (changes requested):

sdk/src/test/scala/io/horizen/account/websocket/WebSocketAccountServerEndpointTest.scala
MarcoOl94: Because we need it to create a WsEndpoint instance, but maybe I miss something?

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountConfigurator.scala
```diff
@@ -5,20 +5,20 @@ import jakarta.websocket.server.{HandshakeRequest, ServerEndpointConfig}
...
 
+import java.net.InetAddress
 
 class WebSocketAccountConfigurator() extends ServerEndpointConfig.Configurator with SparkzLogging {
 
   override def modifyHandshake(config: ServerEndpointConfig, request: HandshakeRequest, response: HandshakeResponse): Unit = {
-    val remoteAddress = request.asInstanceOf[RequestContext].getRemoteAddr
-    if (!WebSocketAccountConfigurator.allowedIPs.contains(remoteAddress)) {
+    val remoteAddress = InetAddress.getByName(request.asInstanceOf[RequestContext].getRemoteAddr)
+    if (!remoteAddress.isAnyLocalAddress && !remoteAddress.isLoopbackAddress && !remoteAddress.isSiteLocalAddress && !WebSocketAccountConfigurator.allowedIPs.contains(remoteAddress.getHostAddress))
```
i-Alex, line 14: It seems a bit different logic now.
I think we need to have `if allowedIps.isEmpty -> chech localhost` -> else check `WebSocketAccountConfigurator.allowedIPs.contains(remoteAddress.getHostAddress))`

Because now, we can't reject localhost connections, they are always accepted

sdk/src/test/scala/io/horizen/account/websocket/WebSocketAccountServerEndpointTest.scala
```diff
@@ -106,6 +114,23 @@ class WebSocketAccountServerEndpointTest extends JUnitSuite with MockitoSugar wi
     null
   }
 
+  @Test
+  def allowedOriginsTest(): Unit = {
+    server = WebSocketAccountServerRef(mockedSidechainNodeViewHolderRef, rpcProcessor, mockedWebsocketServerSettingsWithExternalOrigin)
```
i-Alex, line 119: let's have 2 different tests:
1) for `mockedWebsocketServerSettings` when we have a default behavior (accept localhost only)
2) for `mockedWebsocketServerSettingsWithExternalOrigin` when we expect only remote one `185.123.0.12`, so localhost connection is refused.

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountConfigurator.scala
MarcoOl94: Ok sorry I missunderstood, I thought we wanted to allow localhost by default.

i-Alex (approved):

sdk/src/main/scala/io/horizen/SidechainSettings.scala
```diff
@@ -26,8 +26,9 @@ case class WebSocketClientSettings(
 
 case class WebSocketServerSettings(
     wsServer: Boolean = false,
-    wsServerPort: Int = 8025
-) extends SensitiveStringer
+    wsServerPort: Int = 8025,
+    wsServerAllowedOrigins: Seq[String] = Seq()
```
paolocappelletti, line 30: since the allowed list is implemented only in the account app, 
let's raise an exception if this property is detected and not empty  inside SidechainApp (utxo one)

paolocappelletti (changes requested):

paolocappelletti (approved):

PR #804, opened 2023-03-30
[SDK-827] StateDbAccountStateView: failed transaction log level fixes

i-Alex:
Before:
```
[ERROR] 2023-03-30 19:24:22:674 +0300 [StateDbAccountStateView.scala:211 ] [2-Hop-akka.actor.default-dispatcher-16] io.horizen.account.state.AccountStateView - applying message failed, tx.id=3bc1b2459b8ed7608b2b50ac2db8810dc4e56d46df14a9039510b27b73a995b4
io.horizen.account.state.ExecutionRevertedException: Key Rotation Proof is invalid: Key rotation proof - master signature is invalid: 0
	at io.horizen.account.state.CertificateKeyRotationMsgProcessor.execSubmitKeyRotation(CertificateKeyRotationMsgProcessor.scala:159) ~[sidechains-sdk-0.7.0-SNAPSHOT.jar:0.7.0-SNAPSHOT]
	at io.horizen.account.state.CertificateKeyRotationMsgProcessor.process(CertificateKeyRotationMsgProcessor.scala:44) ~[sidechains-sdk-0.7.0-SNAPSHOT.jar:0.7.0-SNAPSHOT]
	at io.horizen.account.state.StateTransition.transition(StateTransition.scala:54) ~[sidechains-sdk-0.7.0-SNAPSHOT.jar:0.7.0-SNAPSHOT]
	at io.horizen.account.state.StateDbAccountStateView.applyMessage(StateDbAccountStateView.scala:163) ~[sidechains-sdk-0.7.0-SNAPSHOT.jar:0.7.0-SNAPSHOT]
	at io.horizen.account.state.StateDbAccountStateView.$anonfun$applyTransaction$1(StateDbAccountStateView.scala:206) ~[sidechains-sdk-0.7.0-SNAPSHOT.jar:0.7.0-SNAPSHOT]
```

After:
```
[DEBUG] 2023-03-30 19:31:22:503 +0300 [StateDbAccountStateView.scala:211 ] [2-Hop-akka.actor.default-dispatcher-11] io.horizen.account.state.AccountStateView - applying message failed, tx id: 8b78e73788c38a9fbe51c561fa0282b767da38c63b827412651a48b2d6ef8837, reason: Key Rotation Proof is invalid: Key rotation proof - signing signature is invalid: 0
```

### DIFF
--- a/sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
+++ b/sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
@@ -1,23 +1,25 @@
 package io.horizen.account.state;
 
+import org.web3j.utils.Numeric;
+
 /**
  * Message processing failed, also revert-and-keep-gas-left.
  */
 public class ExecutionRevertedException extends ExecutionFailedException {
-    public final byte[] revertReason;
+    public final byte[] returnData;
 
-    public ExecutionRevertedException(byte[] revertReason) {
-        super("execution reverted");
-        this.revertReason = revertReason;
+    public ExecutionRevertedException(byte[] returnData) {
+        super(String.format("execution reverted with return data \"%s\"", Numeric.toHexString(returnData)));
+        this.returnData = returnData;
     }
 
     public ExecutionRevertedException(String message) {
         super(message);
-        this.revertReason = new byte[0];
+        this.returnData = new byte[0];
     }
 
     public ExecutionRevertedException(String message, Throwable cause) {
         super(message, cause);
-        this.revertReason = new byte[0];
+        this.returnData = new byte[0];
     }
 }
--- a/sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
+++ b/sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
@@ -1,6 +1,7 @@
 package io.horizen.account.state;
 
 import org.web3j.utils.Numeric;
+import scala.Array;
 
 /**
  * Message processing failed, also revert-and-keep-gas-left.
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
```diff
@@ -1,23 +1,25 @@
 package io.horizen.account.state;
 
+import org.web3j.utils.Numeric;
+
 /**
  * Message processing failed, also revert-and-keep-gas-left.
  */
 public class ExecutionRevertedException extends ExecutionFailedException {
-    public final byte[] revertReason;
+    public final byte[] returnData;
```
saratnt, line 9: Considering how it is used, maybe it is better to save directly the hex string instead of byte[]

sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
```diff
@@ -1,23 +1,25 @@
...
     public ExecutionRevertedException(String message) {
         super(message);
-        this.revertReason = new byte[0];
+        this.returnData = new byte[0];
     }
 
     public ExecutionRevertedException(String message, Throwable cause) {
         super(message, cause);
-        this.revertReason = new byte[0];
+        this.returnData = new byte[0];
```
saratnt, line 23: Maybe a constant can be defined, instead of creating a new object every time (eg `static final byte[] EMPTY_RETURN_DATA = new byte[0];`)

saratnt (approved):

paolocappelletti (approved):

sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
i-Alex: done

sdk/src/main/java/io/horizen/account/state/ExecutionRevertedException.java
i-Alex: For me the current usage is just a particular case to "print" the field, but in general it is expected to be in a raw format and can be used by native/evm smart contract processors.
For example, later we will introduce the interoperability between the contracts, so the caller-contract may require to process the `returnData` field from the delegated-contract exception.

saratnt: OK, that makes sense.

saratnt (approved):

PR #805, opened 2023-03-31
Return JSON error in case of malformed request body json in RPC API

i-skrypnyk:


i-skrypnyk:
@MarcoOl94 @JcBernack please review, you were suggested by github :)

JcBernack (approved):
LGTM

sdk/src/main/scala/io/horizen/account/api/http/route/AccountEthRpcRejectionHandler.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  def rejectionHandler: RejectionHandler =
+    RejectionHandler.newBuilder()
+      .handle {
+        case MalformedRequestContentRejection(msg, _) =>
+          SidechainApiResponse(
+            EthJsonMapper.serialize(
+              new RpcResponseError(
+                new RpcId(),
+                RpcError.fromCode(RpcCode.ParseError, msg)
```
i-Alex, line 20: here we return error code `ParseError(-32700, "Parse error")`
infura example from Cronic returns `{"jsonrpc":"2.0","error":{"code":-32600,"message":"invalid json request"}}` which in our case is `InvalidRequest(-32600, "Invalid request")`

Let's double check what is actually expected.

i-Alex (changes requested):

sdk/src/main/scala/io/horizen/account/api/http/route/AccountEthRpcRejectionHandler.scala
i-skrypnyk: Decided to go with -32700, as it's description fully represent this case. Notified Cronicc about it.
image

i-Alex (approved):

paolocappelletti (approved):

PR #810, opened 2023-04-11
End of withdrawal epoch not correctly handled by Account State

alsala:
Fix for method `getWithdrawalEpochInfo` for handling the case of start from end of epoch and jump directly to the next end of epoch with the maximum number of mc ref blocks


i-Alex (approved):
Looks good!

PR #811, opened 2023-04-11
Sc2Sc base protocol

paologalligit:
Introducing the cross-chain message protocol between sidechains. Here the [document](https://docs.google.com/document/d/1yQZiQShZnv9BgIc6Q8XiO1jQQxvzF0bc5xp018_00ck/edit#heading=h.3ucor72iy7nn)


sdk/src/main/scala/io/horizen/utxo/SidechainApp.scala
```diff
@@ -180,55 +183,64 @@ class SidechainApp @Inject()
...
       Transaction.ModifierTypeId -> sidechainTransactionsCompanion)
 
   override val nodeViewSynchronizer: ActorRef =
     actorSystem.actorOf(SidechainNodeViewSynchronizer.props(networkControllerRef, nodeViewHolderRef,
-        SidechainSyncInfoMessageSpec, settings.network, timeProvider, modifierSerializers))
+      SidechainSyncInfoMessageSpec, settings.network, timeProvider, modifierSerializers))
 
   // Init Forger with a proper web socket client
-  val sidechainBlockForgerActorRef: ActorRef = forge.ForgerRef("Forger", sidechainSettings, nodeViewHolderRef,  mainchainSynchronizer, sidechainTransactionsCompanion, timeProvider, params)
+  val sidechainBlockForgerActorRef: ActorRef = ForgerRef("Forger", sidechainSettings, nodeViewHolderRef, mainchainSynchronizer, sidechainTransactionsCompanion, timeProvider, sc2scConfigurator, params)
```
i-Alex, line 202: A side comment to NetworkParams class.
You may see in this file a lot of places where we pass `sc2scConfigurator` to a constructor altogether with `params`. Can be simplified a lot by putting it inside `params` once.

sdk/src/main/scala/io/horizen/utxo/SidechainApp.scala
```diff
@@ -180,55 +183,64 @@ class SidechainApp @Inject()
...
   val cswManager: Option[ActorRef] = if (isCSWEnabled) Some(CswManagerRef(sidechainSettings, params, nodeViewHolderRef)) else None
 
   //Websocket server for the Explorer
-  if(sidechainSettings.websocketServer.wsServer) {
-    val webSocketServerActor: ActorRef = WebSocketServerRef(nodeViewHolderRef,sidechainSettings.websocketServer.wsServerPort)
+  if (sidechainSettings.websocketServer.wsServer) {
+    val webSocketServerActor: ActorRef = WebSocketServerRef(nodeViewHolderRef, sidechainSettings.websocketServer.wsServerPort)
   }
 
+  var sc2scProverRef: Option[ActorRef] = if (sc2scConfigurator.canSendMessages) Some(Sc2ScProverRef(sidechainSettings, nodeViewHolderRef, params)) else None
```
i-Alex, line 221: Prover instance can be declared exactly where we create `Sc2scApiRoute` below.
```
if (sc2scConfigurator.canSendMessages) {
 val prover = Sc2ScProverRef(sidechainSettings, nodeViewHolderRef, params)
 ret = ret :+ Sc2scApiRoute(settings.restApi, nodeViewHolderRef, prover)
```

sdk/src/main/scala/io/horizen/utxo/SidechainApp.scala
```diff
@@ -180,55 +183,64 @@ class SidechainApp @Inject()
...
-    SidechainBlockApiRoute(settings.restApi, nodeViewHolderRef, sidechainBlockActorRef, sidechainTransactionsCompanion, sidechainBlockForgerActorRef, params),
-    SidechainNodeApiRoute(peerManagerRef, networkControllerRef, timeProvider, settings.restApi, nodeViewHolderRef, this, params),
-    SidechainTransactionApiRoute(settings.restApi, nodeViewHolderRef, sidechainTransactionActorRef, sidechainTransactionsCompanion, params, circuitType),
-    SidechainWalletApiRoute(settings.restApi, nodeViewHolderRef, sidechainSecretsCompanion),
-    SidechainSubmitterApiRoute(settings.restApi, params, certificateSubmitterRef, nodeViewHolderRef, circuitType),
-    SidechainCswApiRoute(settings.restApi, nodeViewHolderRef, cswManager, params),
-    SidechainBackupApiRoute(settings.restApi, nodeViewHolderRef, boxIterator, params)
-  )
+  override lazy val coreApiRoutes: Seq[ApiRoute] = {
+    var ret = Seq(
```
i-Alex, line 227: let's choose more contextual name for `ret`

sdk/src/main/scala/io/horizen/utxo/state/SidechainBackup.scala
```diff
@@ -26,7 +26,7 @@ class SidechainBackup @Inject()
    @Named("Params") val params : NetworkParams
   ) extends SparkzLogging
   {
-    protected val sidechainBoxesCompanion: SidechainBoxesCompanion =  SidechainBoxesCompanion(customBoxSerializers)
+    protected val sidechainBoxesCompanion: SidechainBoxesCompanion =  SidechainBoxesCompanion(customBoxSerializers, true)
```
i-Alex, line 29: why don't we account if `sc2sc` is really supported?

sdk/src/main/java/io/horizen/sc2sc/Sc2ScConfigurator.scala
```diff
@@ -0,0 +1,4 @@
+package io.horizen.sc2sc
+
+case class Sc2ScConfigurator(canSendMessages: Boolean, canReceiveMessages: Boolean)
```
i-Alex, line 3: see my comments in AbstractSidechainApp/SidechainApp classes.
We must care about backward compatibility.
`canSendMessages` and `canReceiveMessages` in general can be activated after genesis block and even separately, optionally:
1) A newly created sidechain may activate both sender and receiver at the genesis phase
2) EON will do it after genesis
3) other chain may decide to skip this feature.
It is a for sure a new mechanism for our SDK and ForkManager.

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/CircuitStrategy.scala
```diff
@@ -25,7 +26,7 @@ abstract class CircuitStrategy[
   PM <: SidechainBlockBase[TX, H],
   HIS <: AbstractHistory[TX, H, PM, _, _, _],
   MS <: AbstractState[TX, H, PM, MS],
-  T <: CertificateData](settings: SidechainSettings, params: NetworkParams) extends SparkzLogging{
+  T <: CertificateData](settings: SidechainSettings, sc2scConfig: Sc2ScConfigurator, params: NetworkParams) extends SparkzLogging with Sc2ScUtils[TX, H, PM, MS, HIS] {
```
i-Alex, line 29: no need to extend `Sc2ScUtils` here.
Must be used only by CircuitWithKeyRotation

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/WithKeyRotationCircuitStrategy.scala
```diff
@@ -75,6 +77,13 @@ class WithKeyRotationCircuitStrategy[
 
     val schnorrKeysSignatures = getSchnorrKeysSignatures(state, status.referencedEpoch)
 
+    val sc2ScDataForCertificate: Option[Sc2ScDataForCertificate] =
+      sc2scConfig.canSendMessages match {
+        case true => Some(getDataForCertificateCreation(status.referencedEpoch, state, history, params))
```
i-Alex, line 82: this method is and must be used only in this class (2 times: in message to sign and proof creation).
`Sc2ScDataForCertificate` result object contains `previousTopQualityCertificateHash` that is also used by this type of circuit. See `previousCertificateOption` object couple of lines above.

No sense to duplication this logic.

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/WithKeyRotationCircuitStrategy.scala
```diff
@@ -85,6 +94,7 @@ class WithKeyRotationCircuitStrategy[
       sidechainId,
       backwardTransfers,
       endEpochCumCommTreeHash,
+      sc2ScDataForCertificate,
```
i-Alex, line 97: `CertificateDataWithKeyRotation` creation must be refactored.
This code in general will not work.
Couple of suggestions:
1) `Sc2ScDataForCertificate` can be removed at all
2) `Sc2ScDataForCertificate.previousTopQualityCertificateHash` can be retrieved from `previousCertificateOption`
3) `CertificateDataWithKeyRotation` custom fields must be constructed using circuit specific custom fields appended with sc2sc if needed (if feature is enabled AFTER fork activation only).

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/WithKeyRotationCircuitStrategy.scala
```diff
@@ -108,6 +118,11 @@ class WithKeyRotationCircuitStrategy[
     val keysAndSignatures: SchnorrKeysSignatures = getSchnorrKeysSignatures(state, referencedWithdrawalEpochNumber)
     val keysRootHash: Array[Byte] = circuit.getSchnorrKeysHash(keysAndSignatures)
 
+    val sc2ScDataForCertificate: Option[Sc2ScDataForCertificate] = sc2scConfig.canSendMessages match {
```
i-Alex, line 121: not used at all at the moment.
Must be used exactly like in `buildCertificateData` method above.

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/WithoutKeyRotationCircuitStrategy.scala
```diff
@@ -22,9 +23,10 @@ class WithoutKeyRotationCircuitStrategy[
   PM <: SidechainBlockBase[TX, H],
   HIS <: AbstractHistory[TX, H, PM, _, _, _],
   MS <: AbstractState[TX, H, PM, MS]](settings: SidechainSettings,
+                                      sc2scConfig: Sc2ScConfigurator,
                                       params: NetworkParams,
                                       circuit: ThresholdSignatureCircuit)
-  extends CircuitStrategy[TX, H, PM, HIS, MS, CertificateDataWithoutKeyRotation](settings, params) {
+  extends CircuitStrategy[TX, H, PM, HIS, MS, CertificateDataWithoutKeyRotation](settings, sc2scConfig, params) {
```
i-Alex, line 29: revert all the changes in this class because it doesn't support sc2sc.
Circuit WITHOUT key rotation has no "free" custom fields for sc2sc needs.

sdk/src/main/scala/io/horizen/certificatesubmitter/dataproof/CertificateDataWithKeyRotation.scala
```diff
@@ -6,28 +6,29 @@ import com.horizen.certnative.BackwardTransfer
...
                                           override val schnorrKeyPairs: Seq[(SchnorrProposition, Option[SchnorrProof])],
                                           schnorrKeysSignatures: SchnorrKeysSignatures,
                                           previousCertificateOption: Option[WithdrawalEpochCertificate],
                                           genesisKeysRootHash: Array[Byte]
                                          )
-  extends CertificateData(referencedEpochNumber, sidechainId, backwardTransfers, endEpochCumCommTreeHash, btrFee, ftMinAmount, schnorrKeyPairs) {
+  extends CertificateData(referencedEpochNumber, sidechainId, backwardTransfers, endEpochCumCommTreeHash, sc2ScDataForCertificate, btrFee, ftMinAmount, schnorrKeyPairs) {
 
   override def getCustomFields: Seq[Array[Byte]] = {
+    //TODO: use sc2ScDataForCertificate to add custom fields for sc2sc
```
i-Alex, line 30: here we should create a seq of 32 elements, where the first element is used for `keysRootHash` and others 31 for whatever needs. In out case second and thirst is used for sc2sc (if enabled and activated)

sdk/src/main/scala/io/horizen/certificatesubmitter/dataproof/CertificateDataWithoutKeyRotation.scala
```diff
@@ -13,13 +14,15 @@ case class CertificateDataWithoutKeyRotation(override val referencedEpochNumber:
                                              override val sidechainId: Array[Byte],
                                              override val backwardTransfers: Seq[BackwardTransfer],
                                              override val endEpochCumCommTreeHash: Array[Byte],
+                                             override val sc2ScDataForCertificate: Option[Sc2ScDataForCertificate],
```
i-Alex, line 17: revert all the changes in this class. this circuit doesn't support sc2sc

tools/sctool/src/main/java/io/horizen/ScBootstrappingToolCommandProcessor.java
i-Alex, line 1: Account model sidechain should keep in mind the support of sc2sc and its impact on genesis block statedb root.
sc2sc specific message processors may impact the root and lead to an inconsistent `genesisBlock.header.stateRoot` value.

i-Alex (commented):
CR is in progress.
First batch of the comments related to the circuits part and sc2sc initialization.

i-Alex (commented):
CR is in progress.
First batch of the comments related to the circuits part and sc2sc initialization.

sdk/src/main/java/io/horizen/utxo/box/CrossChainMessageBox.java
```diff
@@ -0,0 +1,49 @@
+package io.horizen.utxo.box;
+
+import io.horizen.proposition.PublicKey25519Proposition;
+import io.horizen.sc2sc.CrossChainProtocolVersion;
+import io.horizen.utxo.box.data.CrossChainMessageBoxData;
+
+import static io.horizen.utxo.box.CoreBoxesIdsEnum.CrossChainMessageBoxId;
+
+public final class CrossChainMessageBox
```
i-Alex, line 9: During the review of sc2sc `CrossChainMessage` part and `CrossChainMessageBox` I saw the following ideas:
1. `CrossChainMessage ` is an interface with one particular implementation `CrossChainMessageImpl`. All the validation code was relying on that interface. So, theoretically application developers may define their own implementations. I guess the main reason was to have a contextual `payload` type (not raw bytes) in the ancestor if needed.
2. `CrossChainMessageBox` is the final class. So, we force developers to use it if they want to have sc2sc. `payload` is always raw (bytes), no context from the core perspective. But on application side developer must know how to decode this raw data to something that have sense -> then to validate it according to the application rules.

In the comments for `CrossChainMessage` I put some arguments of making it as a `public final class` instead of an `interface`.

Now I'm thinking about the benefits of having `CrossChainMessageBox` as `public final`. We may keep it as is, but it seems that we should do an opposite:

1. Make it as a `public abstract`. The same for `CrossChainMessageBoxData`
2. Add a `final` methods implementation of `getCrossChainMessage` and several others.
3. In all the core validation rules deal with these abstract classes without additional context, but allow developers to extend them and define the exact payload data structure.
4. Keep undefined the boxId of `CrossChainMessageBox` and remove `CrossChainMessageBox` from core boxes and core box serializes.

The very same logic must be applied to `CrossChainRedeemMessageBox`

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainMessageProcessor.scala
```diff
@@ -0,0 +1,173 @@
...
+    calculateKey(Bytes.concat("crossChainMsgEpochCounter".getBytes, Ints.toByteArray(withdrawalEpoch)))
+  }
+
+  private[horizen] def getMessageKey(withdrawalEpoch: Int, counter: Int): Array[Byte] = {
+    calculateKey(Bytes.concat("crossChainMsg".getBytes, Ints.toByteArray(withdrawalEpoch), Ints.toByteArray(counter)))
+  }
+}
+
+object AbstractCrossChainMessageProcessor {
+  private[horizen] def buildCrosschainMessageFromAccount(data: AccountCrossChainMessage, params: NetworkParams): CrossChainMessage = {
```
i-Alex, line 136: same as for UTXO model: let's move this method to `AccountCrossChainMessage` class.

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainMessageProcessor.scala
```diff
@@ -0,0 +1,173 @@
...
+
+import java.math.BigInteger
+import java.util
+import scala.collection.JavaConverters.seqAsJavaListConverter
+
+trait CrossChainMessageProvider {
+  private[horizen] def getCrossChainMessages(epochNum: Int, view: BaseAccountStateView): Seq[CrossChainMessage]
+  private[horizen] def getCrossChainMessageHashEpoch(msgHash: CrossChainMessageHash, view: BaseAccountStateView): Option[Int]
+}
+abstract class AbstractCrossChainMessageProcessor(networkParams: NetworkParams) extends NativeSmartContractMsgProcessor with CrossChainMessageProvider {
```
i-Alex, line 28: Consider that this class will be used/extended by application developers => must be Java class.
Same for `AbstractCrossChainRedeemMessageProcessor`

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainMessageProcessor.scala
```diff
@@ -0,0 +1,173 @@
...
+    if (numOfReqs >= MaxCrosschainMessagesPerEpoch) {
+      throw new ExecutionRevertedException("Reached maximum number of CrosschainMessages per epoch: request is invalid")
+    }
+
+    val request = AccountCrossChainMessage(messageType, sender.toBytes, receiverSidechain, receiver, payload)
+    val messageHash = CryptoLibProvider.sc2scCircuitFunctions.getCrossChainMessageHash(AbstractCrossChainMessageProcessor.buildCrosschainMessageFromAccount(request, networkParams))
+
+    //check for duplicates in this and other message processor, in any epoch
+    if (view.getCrossChainMessageHashEpoch(messageHash).nonEmpty) {
+      throw new ExecutionRevertedException("Dupicate crosschain message")
```
i-Alex, line 87: type in message

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainMessageProcessor.scala
```diff
@@ -0,0 +1,173 @@
...
+    val data = view.getAccountStorage(contractAddress, messageHash.bytes)
+    if (data.sameElements(NULL_HEX_STRING_32)) {
+      Option.empty
+    } else {
+      Some(Ints.fromByteArray(data))
+    }
+  }
+
+
+  protected def addCrossChahinMessage(messageType: Int,
```
i-Alex, line 70:
```suggestion
  protected def addCrossChainMessage(messageType: Int,
```
Beside the name method may become final.
Same for some others methods, we for sure don't want storage related methods to be overridden. It may break the class.

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,150 @@
...
+      accCcRedeemMessage.scCommitmentTreeRoot,
+      accCcRedeemMessage.nextScCommitmentTreeRoot,
+      accCcRedeemMessage.proof
+    )
+    val evmLog = getEthereumConsensusDataLog(event)
+    view.addLog(evmLog)
+  }
+
+  private def addCrossChainMessageToView(view: BaseAccountStateView, accCcRedeemMessage: AccountCrossChainRedeemMessage): Unit = {
+    val messageHash = CryptoLibProvider.sc2scCircuitFunctions.getCrossChainMessageHash(
```
i-Alex, line 47:
```suggestion
    val messageHash = sc2scCircuit.getCrossChainMessageHash(
```
don't use `CryptoLibProvider.sc2scCircuitFunctions`.
And in general the same comment as in UTXO model: let's use OOP style: `accCcRedeemMessage.getAccountCrooschainMessage.getHash`

sdk/src/main/scala/io/horizen/account/state/events/AddCrossChainMessage.scala
```diff
@@ -0,0 +1,34 @@
...
+import io.horizen.account.state.events.annotation.{Indexed, Parameter}
+import io.horizen.evm.Address
+import org.web3j.abi.datatypes.DynamicBytes
+import org.web3j.abi.datatypes.generated.Uint32
+
+import scala.annotation.meta.getter
+
+case class AddCrossChainMessage(@(Parameter@getter)(1) @(Indexed@getter) sender: Address,
+                                @(Parameter@getter)(2) @(Indexed@getter) messageType: Uint32,
+                                @(Parameter@getter)(3) receiverSidechain: DynamicBytes,
```
i-Alex, line 12:
```suggestion
                                @(Parameter@getter)(3) receiverSidechain: Bytes32,
```
use as much precise types as it is possible

sdk/src/main/scala/io/horizen/account/state/events/AddCrossChainRedeemMessage.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import io.horizen.account.sc2sc.{AccountCrossChainMessage, AccountCrossChainMessageSerializer}
+import io.horizen.account.state.events.annotation.{Indexed, Parameter}
+import org.web3j.abi.datatypes.DynamicBytes
+import org.web3j.abi.datatypes.generated.Bytes32
+
+import scala.annotation.meta.getter
+
+case class AddCrossChainRedeemMessage(
+                                       @(Parameter @getter)(1) @(Indexed @getter) accountCrossChainMessage: DynamicBytes,
```
i-Alex, line 11: I'm not sure about the benefit of having `accountCrossChainMessage bytes` indexed.
What can be useful is to keep `accountCrossChainMessage` in a context form and index `receiver` and sender sidechain id fields.
Then the user can easily filter the log by his address to see the Redeems for his account, for example. 

Other fields, like `proof`, have no sense to be put into the Log.

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,150 @@
...
+    } catch {
+      case exception: IllegalArgumentException =>
+        throw new ExecutionRevertedException("Error while validating redeem message", exception)
+    }
+  }
+
+  private def validateSemantics(byteArraysToValidate: Array[Byte]*): Unit = {
+    val byteArrayMaxLength = 64
+    byteArraysToValidate.foreach(arrayToValidate => {
+      if (arrayToValidate.isEmpty || arrayToValidate.length > byteArrayMaxLength) {
```
i-Alex, line 88: aren't we sure about the exact expected lengths?
like: sidechain id is strictly 32, receiver is exactly 20, etc.

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -0,0 +1,150 @@
...
+    if (!scTxMsgProc.doesScTxCommitmentTreeRootHashExist(scCommitmentTreeRoot, view)) {
+      throw new IllegalArgumentException(s"Sidechain commitment tree root `${BytesUtils.toHexString(scCommitmentTreeRoot)}` does not exist")
+    }
+
+    if (!scTxMsgProc.doesScTxCommitmentTreeRootHashExist(nextScCommitmentTreeRoot, view)) {
+      throw new IllegalArgumentException(s"Sidechain next commitment tree root `${BytesUtils.toHexString(nextScCommitmentTreeRoot)}` does not exist")
+    }
+  }
+
+  private def validateProof(ccRedeemMessage: AccountCrossChainRedeemMessage): Unit = {
```
i-Alex, line 121: snark proof validation is very resource consuming operation, we need to evaluate the proper gas cost for it.
Otherwise, it can be a big bottleneck or even a source of attack.

i-Alex (commented):
Batch of comments for account model.
CR is finished.

sdk/src/main/java/io/horizen/utxo/box/data/CrossChainMessageBoxData.java
```diff
@@ -0,0 +1,75 @@
...
+    private CrossChainProtocolVersion protocolVersion;
+    private final Integer messageType;
+    private final byte[]  receiverSidechain;
+    private final byte[]  receiverAddress;
+    private final byte[]  payload;
+
+    public CrossChainMessageBoxData(PublicKey25519Proposition proposition,
+                                    CrossChainProtocolVersion protocolVersion,
+                                    Integer messageType,
+                                    byte[] receiverSidechain,
```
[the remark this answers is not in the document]
paolocappelletti: this box should be created in upper protocols layer only as a result of another tx, so I expect additional checks there.
But maybe as an additional security we can add it anyway

sdk/src/main/java/io/horizen/validation/crosschain/receiver/CrossChainRedeemMessageValidator.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private void validateMsgDoubleRedeem(CrossChainMessage ccMsg) throws Exception {
+        CrossChainMessageHash currentMsgHash = sc2scCircuit.getCrossChainMessageHash(ccMsg);
+        boolean ccMsgFromRedeemAlreadyExists = scStateStorage.doesCrossChainMessageHashFromRedeemMessageExist(currentMsgHash);
+        if (ccMsgFromRedeemAlreadyExists) {
+            throw new IllegalArgumentException(String.format("The message `%s` has already been redeemed", ccMsg));
+        }
+    }
+
+    private void verifyCommitmentTreeRootAndNextCommitmentTreeRoot(byte[] scCommitmentTreeRoot, byte[] nextScCommitmentTreeRoot) {
```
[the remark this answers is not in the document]
paolocappelletti: the circuit already vefies that in nextScCommitmentTreeRoot there is a certificate pointing to the certificate in scCommitmentTreeRoot, so this should implicitly guarantee that it is following

PR #812, opened 2023-04-14
Fixed websocket server actor restart on exception

MarcoOl94:


sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServerEndpoint.scala
```diff
@@ -249,7 +258,10 @@ private object WebSocketAccountServerEndpoint extends SparkzLogging {
   }
 
   def onVaultChanged(): Unit = {
-    walletAddresses = webSocketAccountChannelImpl.getWalletAddresses
+    webSocketAccountChannelImpl.getWalletAddresses match {
+      case Success(addresses) => walletAddresses = addresses
+      case Failure(exception) => log.debug("Websocket failed to update walletAddresses "+exception.getMessage)
```
i-Alex, line 263: Currently, in case wallet has been changed (new secret added), but this method failed because of a timeout -> we will never leave `walletAddresses` as before.
Do we want instead to try again until we succeed?

i-Alex (commented):
Looks good. Now websocket must not crash, but we may miss some events notification.
Seems to be another issue.

sdk/src/main/scala/io/horizen/account/websocket/WebSocketAccountServerEndpoint.scala
MarcoOl94: I suppose that if we are not able to access the wallet, adding a retry would just decrease more tha node performance. We may consider to add some sort of queue that keeps track of the failed requests and try to re-executed them in a second moment.

paolocappelletti (approved):

PR #814, opened 2023-04-19
Contract interop

JcBernack:
- See [SDK-931](https://horizenlabs.atlassian.net/browse/SDK-931)
- SDK https://github.com/HorizenOfficial/Sidechains-SDK/pull/814
- libevm https://github.com/HorizenOfficial/libevm/pull/9
- geth fork https://github.com/HorizenOfficial/go-ethereum/pull/1

[SDK-931]: https://horizenlabs.atlassian.net/browse/SDK-931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


saratnt (approved):

sdk/src/main/scala/io/horizen/account/state/StateTransition.scala
```diff
@@ -146,4 +128,55 @@ class StateTransition(
     // return remaining gas to the gasPool of the current block so it is available for the next transaction
     blockGasPool.addGas(gas.getGas)
   }
+
+  @throws(classOf[InvalidMessageException])
+  @throws(classOf[ExecutionFailedException])
+  def execute(invocation: Invocation): Array[Byte] = {
+    // limit call depth to 1024
+    if (invocationStack.length >= 1024) throw new ExecutionRevertedException("Maximum depth of call stack reached")
```
davehorizen, line 136: Reading from the PR specs, this stack depth value taken from the current geth implementation.
If the default in the EVM is to allocate 63/64th of the currently available gas to the nested call and if we consider the current block gas limit this call stack will not be reached, right? The tx gas will play the upper limit in this case.

qa/sc_evm_native_interop.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python3
```
davehorizen, line 1: Not directly related to the logic of this test and given the fact that we do not support `DelegateCall`, do you think it is necessary to add a test that verifies the correct throwing of an exception in the case of a contract A that executes `DelegateCall` on the contract B or would it not make sense as there is no risk of unexpected state change?

davehorizen (changes requested):

sdk/src/main/scala/io/horizen/account/state/StateTransition.scala
JcBernack: Possible, yes. It all depends on the gas parameters, how much gas we consume for what, etc.

qa/sc_evm_native_interop.py
JcBernack: No, you're right. We should add a test verifying correct behavior in case someone tries DelegateCall from the EVM to a native contract.

sdk/src/main/java/io/horizen/account/state/EvmMessageProcessor.java
```diff
@@ -86,4 +98,36 @@ protected Hash getBlockHash(BigInteger blockNumber) {
...
+                        gasPool,
+                        invocation.readOnly
+                    )
+                );
+                return new InvocationResult(returnData, gasPool.getGas(), "", false, null);
+            } catch (ExecutionRevertedException e) {
+                // forward the revert reason if any
+                return new InvocationResult(e.returnData, gasPool.getGas(), e.getMessage(), true, null);
+            } catch (Exception e) {
+                return new InvocationResult(new byte[0], gasPool.getGas(), e.getMessage(), false, null);
```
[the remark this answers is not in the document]
JcBernack: Good catch! I must have written too much Go prior to this part, Go errors always have a message.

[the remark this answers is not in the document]
JcBernack: fixed

qa/sc_evm_native_interop.py
JcBernack: done

qa/SidechainTestFramework/account/smart_contract_resources/contracts/ForgerStakes.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
```
alsala, line 1: Actually this is exactly the same file as `./qa/truffle/contracts/ForgerStakes.sol`, which is currently useful only as reference for manual tests to be executed outside the test framework. Consider removing it.

qa/SidechainTestFramework/account/smart_contract_resources/contracts/WithdrawalRequests.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
```
alsala, line 1: The solidity interface seems not compatible with the native smart contract in the SDK. In this case the file `qa/truffle/contracts/WithdrawalRequests.sol` seems to be the correct one.

paolocappelletti:
already included in st/interop_final

PR #817, opened 2023-04-24
SDK-916 : Error logs when certificate submission fails

ddrvar:
[Jira ticket](https://horizenlabs.atlassian.net/browse/SDK-916)

### DIFF
--- a/qa/sc_multiple_cert_submitters.py
+++ b/qa/sc_multiple_cert_submitters.py
@@ -108,10 +108,9 @@ class SCMultipleCertSubmitters(SidechainTestFramework):
         with open(os.path.join(self.options.tmpdir + "/sc_node0/log", "debugLog.txt")) as node0, open(os.path.join(self.options.tmpdir + "/sc_node1/log", "debugLog.txt")) as node1:
             log0 = node0.read()
             log1 = node1.read()
-            submission_error = "[ERROR]"
+            submission_not_needed_msg = "Submission not needed. Certificate of equal or higher quality already present in epoch 0"
             submission_successful = "Backward transfer certificate response had been received"
-            assert_true((submission_error not in log0 and submission_successful in log1) or (submission_error not in log1 and submission_successful in log0))
-
+            assert_true((submission_not_needed_msg in log0 and submission_successful in log1) or (submission_not_needed_msg in log1 and submission_successful in log0))
 
 if __name__ == "__main__":
     SCMultipleCertSubmitters().main()
--- a/sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/CeasingSidechain.scala
+++ b/sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/CeasingSidechain.scala
@@ -32,6 +32,8 @@ class CeasingSidechain(mainchainChannel: MainchainNodeChannel, params: NetworkPa
         case Success(currentCertificateTopQuality) =>
           if (status.knownSigs.size > currentCertificateTopQuality)
             return true
+          else
+            log.info(s"Submission not needed. Certificate of equal or higher quality already present in epoch " + status.referencedEpoch)
         case Failure(e) => e match {
           // May happen if there is a bug on MC side or the SDK code is inconsistent to the MC one.
           case ex: WebsocketErrorResponseException =>
--- a/sdk/src/test/scala/io/horizen/certificatesubmitter/strategy/NonCeasingSidechainTest.scala
+++ b/sdk/src/test/scala/io/horizen/certificatesubmitter/strategy/NonCeasingSidechainTest.scala
@@ -129,7 +129,7 @@ class NonCeasingSidechainTest extends JUnitSuite
         )
       }
     })
-    assertFalse("QQuality check must fail.", nonCeasingSidechainStrategy.checkQuality(status))
+    assertFalse("Quality check must fail.", nonCeasingSidechainStrategy.checkQuality(status))
 
     //test - higher quality certificate present
     Mockito.when(mainchainChannel.getTopQualityCertificates(ArgumentMatchers.any())).thenAnswer(_ => {


JcBernack:
A bit off-topic: Please check your git config, somehow your name is set to `=` :wink: 
image

qa/sc_evm_multiple_cert_submitters.py
```diff
@@ -0,0 +1,263 @@
+#!/usr/bin/env python3
```
MarcoOl94, line 1: I think we can simplify a lof the test. We just need to have 2 SC nodes and generate 1 certificate. After that we can check the nodes logs and verify that we have the expected error messages.

MarcoOl94 (changes requested):

sdk/src/main/scala/io/horizen/certificatesubmitter/AbstractCertificateSubmitter.scala
```diff
@@ -413,7 +414,10 @@ abstract class AbstractCertificateSubmitter[
                     log.info(s"Backward transfer certificate response had been received. Cert hash = " + BytesUtils.toHexString(certificate.certificateId))
 
                   case Failure(ex) =>
-                    log.error("Creation of backward transfer certificate had been failed.", ex)
+                    if (ex.isInstanceOf[CertificateAlreadyPresentException])
+                      log.info(s"Submission failed. Certificate already present in epoch " + dataForProofGeneration.referencedEpochNumber)
```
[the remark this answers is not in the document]
ddrvar: Done

sdk/src/main/scala/io/horizen/websocket/client/MainchainNodeChannelImpl.scala
```diff
@@ -156,6 +157,14 @@ class MainchainNodeChannelImpl(client: CommunicationClient, params: NetworkParam
         certificateRequest.fieldElementCertificateFields.map(BytesUtils.toHexString),
         certificateRequest.bitVectorCertificateFields.map(BytesUtils.toHexString))
 
+    // check if there are already certificates present in mempool for current or higher epoch
```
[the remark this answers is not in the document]
ddrvar: Done

sdk/src/main/scala/io/horizen/websocket/client/MainchainNodeChannelImpl.scala
```diff
@@ -156,6 +157,14 @@ class MainchainNodeChannelImpl(client: CommunicationClient, params: NetworkParam
...
+    getTopQualityCertificates(BytesUtils.toHexString(certificateRequest.sidechainId)) match {
+      case Success(mcRefTry) =>
+        if ((mcRefTry.mempoolCertInfo.get.epoch >= certificateRequest.epochNumber && mcRefTry.mempoolCertInfo.get.certHash != null) || (mcRefTry.chainCertInfo.get.epoch >= certificateRequest.epochNumber  && mcRefTry.chainCertInfo.get.certHash != null))
```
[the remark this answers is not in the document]
ddrvar: Done

qa/sc_evm_multiple_cert_submitters.py
ddrvar: Done

sdk/src/main/scala/io/horizen/certificatesubmitter/AbstractCertificateSubmitter.scala
```diff
@@ -408,13 +409,30 @@ abstract class AbstractCertificateSubmitter[
...
-                  case Success(certificate) =>
-                    log.info(s"Backward transfer certificate response had been received. Cert hash = " + BytesUtils.toHexString(certificate.certificateId))
-
-                  case Failure(ex) =>
-                    log.error("Creation of backward transfer certificate had been failed.", ex)
+                // check if there are already certificates present in mempool for current or higher epoch
+                var isCertificateAlreadyPresent = false
+                mainchainChannel.getTopQualityCertificates(BytesUtils.toHexString(BytesUtils.reverseBytes(params.sidechainId))) match {
+                  case Success(mcRefTry) =>
+                    (mcRefTry.mempoolCertInfo, mcRefTry.chainCertInfo) match {
```
MarcoOl94, line 416: Here we are supposing to have both field (mempoolCertInfo and chainCertInfo) populated, are we sure that this is always the case?

qa/sc_evm_multiple_cert_submitters.py
```diff
@@ -0,0 +1,141 @@
...
+    - creates new forward transfer to sidechain
+    - verifies that the receiving account balance is changed
+    - generate MC and SC blocks to reach the end of the Withdrawal epoch 0
+    - generate one more MC and SC block accordingly and await for certificate submission to MC node mempool
+    - check epoch 0 certificate with not backward transfers in the MC mempool
+    - check that only one SC node forged certificate successfuly, and the other one contains 'submission not needed message'
+"""
+
+
+class SCEvmBackwardTransfer(AccountChainSetup):
```
MarcoOl94, line 43: Please rename the class in **SCEvmMultipleCertSubmittersTest**

MarcoOl94 (changes requested):
Looks good, please add a UnitTest for the new **tryToGenerateCertificate** behavior

qa/sc_evm_multiple_cert_submitters.py
ddrvar: Done

sdk/src/main/scala/io/horizen/certificatesubmitter/AbstractCertificateSubmitter.scala
ddrvar: Match cases expanded to include those when only mempoolCertInfo or chainCertInfo is populated
Upon closer look, I'm not sure if it's always the case. Looking at [this ](https://github.com/HorizenOfficial/Sidechains-SDK/blob/2196d647be114599c1c035c41cb3b3b03b69d3a5/sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/CeasingSidechain.scala#L64)call of getTopQaulityChainCertificates, there are cases when only one of them is taken into consideration

JcBernack: Match clauses also throw an exception if none of the branches match. Hence the compiler warning if a match clause is "non-exhaustive". We should always cover all cases, even if it is a "catch all" with `_ => ..` to throw an exception. This will remove the warning and make the error explicit. Improves readability a lot, imho.

JcBernack (changes requested):

sdk/src/main/scala/io/horizen/websocket/client/MainchainNodeChannelImpl.scala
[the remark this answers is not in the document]
ddrvar: There is already an integration test that checks the case

MarcoOl94 (approved):

JcBernack (approved):

paolocappelletti (changes requested):

qa/sc_multiple_cert_submitters.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # wait till both SC nodes stop generating certificate
+        while sc_node1.submitter_isCertGenerationActive()["result"]["state"] or sc_node2.submitter_isCertGenerationActive()["result"]["state"]:
+            time.sleep(2)
+
+        # Assert that proper log messages are present in both log files - only one node submitted certificate and the other one faced no errors
+        with open(os.path.join(self.options.tmpdir + "/sc_node0/log", "debugLog.txt")) as node0, open(os.path.join(self.options.tmpdir + "/sc_node1/log", "debugLog.txt")) as node1:
+            log0 = node0.read()
+            log1 = node1.read()
+            submission_error = "[ERROR]"
```
MarcoOl94, line 111: Can we test a more specific string? "ERROR" can be also found for other reasons

MarcoOl94 (changes requested):

qa/sc_multiple_cert_submitters.py
ddrvar: Instead of doing blocklisting of errors, I added a log message in CeasingSidechain when a certificate of equal or higher quality is already present in the current epoch, thus making the logging behavior of the two strategies as similar as possible.
I changed this python test to follow that approach, and now it checks if the two appropriate log messages are present, instead of checking if no error exists.

MarcoOl94 (approved):

paolocappelletti (approved):

PR #820, opened 2023-04-28
SDK-394 - chainAfter and chainSince methods

davehorizen:
Old `chainAfter` method converted in the new `chainSince` method that return the list of IDs including the input starting one, new `chainAfter` method added.
Details in task: https://horizenlabs.atlassian.net/browse/SDK-394
Unit tests updated


MarcoOl94 (approved):

i-skrypnyk (approved):
:+1:

PR #821, opened 2023-04-28
SDK-1027 - extraData and sha3Uncles format fix

davehorizen:
`extraData` and `sha3Uncles` format fixed, now they are returned with pattern 0x followed by 64 0s
Unit test updated

### DIFF
--- a/sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumBlockView.java
+++ b/sdk/src/main/java/io/horizen/account/api/rpc/types/EthereumBlockView.java
@@ -41,9 +41,9 @@ public class EthereumBlockView {
     // no PoW, no nonce, but we explicity set it to all zeroes as some RPC clients are very strict here (e.g. GETH)
     public final String nonce = "0x0000000000000000";
     // we do not have uncles
-    public final String sha3Uncles = "0x";
+    public final String sha3Uncles = "0x0000000000000000000000000000000000000000000000000000000000000000";
     // we do not have extraData in the block
-    public final String extraData = "0x";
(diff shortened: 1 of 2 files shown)


mikelodder7:
While these are good defaults, are there checks for when these values are less than 32 bytes = 64 hexits?

davehorizen:
> While these are good defaults, are there checks for when these values are less than 32 bytes = 64 hexits?



> While these are good defaults, are there checks for when these values are less than 32 bytes = 64 hexits?

I don't think checks are needed because these fields exist only at RPC response level, they are not used for internal business logic and their length and format are defined on the SDK side and not on the client side.
In our case we retrieve the block data from the sidechain and use the `EthereumBlockView` class to parse them in the format that the RPC specs required (adding `sha3Uncles` and `extraData` fields, etc.)

mikelodder7:
Okay, I guess I'm not being clear. What I mean is since you are converting the data into that format what if you receive 31 bytes and need to prepend leading 0's? Is there code for this or is it unnecessary?

davehorizen:
> Okay, I guess I'm not being clear. What I mean is since you are converting the data into that format what if you receive 31 bytes and need to prepend leading 0's? Is there code for this or is it unnecessary?

If we are only talking about the `extraData` and `sha3Uncles` fields, you can see that these fields are not managed by the constructor of the `EthereumBlockView` class, consequently no internal or external client can pass a new value or modify its format which is now fixed at 64 0s.
Moreover this class is used only by read-only RPC methods.

mikelodder7:
Okay thanks for the explanation

paolocappelletti (approved):

PR #823, opened 2023-05-09
BootstrappingTool custom message provider injection

paologalligit:
The bootstrapping tool now allow the application layer to customise logic such as injecting custom message processors


MarcoOl94 (approved):

examples/mc_sc_workflow_example.md
paolocappelletti, line 1: Step 2 needs an update here (boostraptool jar to be launched)

examples/utxo/utxoapp_sctool/src/main/java/io/horizen/UTXOSimpleAppBootstrappingTool.java
paolocappelletti, line 1: Let's add a class similar to this one in the tools/sctool module (maybe named "DefaultUTXOBootstrappingTool") and another one for account model.
Can be useful if an app wants just to use the default functionalities without any addition (does not have to definie a whole new full module)

paolocappelletti (changes requested):

paolocappelletti (approved):

PR #831, opened 2023-05-24
Seeder nodes - St/sdk 1090

saratnt:
This PR requires Sparkz as in Sparkz project PR st/SDK_987 (https://github.com/HorizenOfficial/Sparkz/pull/52) to be built


sdk/src/main/scala/io/horizen/utxo/SidechainNodeViewHolder.scala
```diff
@@ -247,8 +247,44 @@ object SidechainNodeViewHolderRef {
...
+                                   params: NetworkParams, timeProvider: NetworkTimeProvider,
+                                   applicationWallet: ApplicationWallet, applicationState: ApplicationState,
+                                   genesisBlock: SidechainBlock) = {
+    if (params.isHandlingTransactionsEnabled)
+      new SidechainNodeViewHolder(sidechainSettings, historyStorage, consensusDataStorage, stateStorage, forgerBoxStorage, utxoMerkleTreeProvider, walletBoxStorage, secretStorage,
+      walletTransactionStorage, forgingBoxesInfoStorage, cswDataProvider, backupStorage, params, timeProvider, applicationWallet, applicationState, genesisBlock)
+    else
+      new SidechainNodeViewHolder(sidechainSettings, historyStorage, consensusDataStorage, stateStorage, forgerBoxStorage, utxoMerkleTreeProvider, walletBoxStorage, secretStorage,
+        walletTransactionStorage, forgingBoxesInfoStorage, cswDataProvider, backupStorage, params, timeProvider, applicationWallet, applicationState, genesisBlock){
+        override protected def updateMemPool(removedBlocks: Seq[SidechainBlock], appliedBlocks: Seq[SidechainBlock],
```
i-skrypnyk, line 271: This looks tricky to me. It looks clean, but for someone who's not aware of this code, they will have a hard time debugging a potential problem, not knowing that the methods were overwritten here.
Also would be tricky to add some new conditions

Do you think having a check in the SidechainNodeViewHolder itself is worse? 
```
  override protected def updateMemPool(removedBlocks: Seq[SidechainBlock], appliedBlocks: Seq[SidechainBlock],
                                       memPool: SidechainMemoryPool, state: SidechainState): SidechainMemoryPool = {
    if (params.isHandlingTransactionsEnabled)
      super.updateMemPool(removedBlocks, appliedBlocks, memPool, state)
    else
      memPool
  }
```
same for applyLocallyGeneratedTransactions

sdk/src/main/scala/io/horizen/api/http/route/WalletBaseApiRoute.scala
```diff
@@ -44,9 +45,10 @@ abstract class WalletBaseApiRoute[
                                    sidechainNodeViewHolderRef: ActorRef,
                                    sidechainSecretsCompanion: SidechainSecretsCompanion
                         )(implicit val context: ActorRefFactory, override val ec: ExecutionContext, override val tag: ClassTag[NV])
-  extends SidechainApiRoute[TX, H, PM, FPI, NH, NS, NW, NP, NV] {
+  extends SidechainApiRoute[TX, H, PM, FPI, NH, NS, NW, NP, NV] with DisableApiRoute {
 
 
+  val myPathPrefix: String = "wallet"
```
i-skrypnyk, line 51: just a suggestion, discard if you think that's bad :)
rename myPathPrefix -> walletPathPrefix \ transactionPathPrefix \ blockPathPrefix \ etc.

sdk/src/main/scala/io/horizen/utxo/SidechainNodeViewHolder.scala
saratnt: This is a good point. For sure having the check in the SidechainNodeViewHolder itself is clearer, I didn't do it because I wanted to avoid repeating a check that cannot change during run-time. What if, instead of using an anonymous class, I use a regular class (e.g.  SidechainNodeViewHolderForSeederNode)? Do you think it would be clearer?

sdk/src/main/scala/io/horizen/api/http/route/WalletBaseApiRoute.scala
saratnt: Ok for me, but what about something more generic, like routePathPrefix? Maybe in the future these fields will be moved to a base class.

sdk/src/main/scala/io/horizen/utxo/SidechainNodeViewHolder.scala
i-skrypnyk: Yes, sounds good

sdk/src/main/scala/io/horizen/api/http/route/WalletBaseApiRoute.scala
i-skrypnyk: that's probably a different idea. What I was thinking is that this field is effectively a constant inside the actor, and adding some context into a variable name helps with readability - f.e. you would know right away what the value of walletPathPrefix is, without looking up where it is set.

saratnt: Ok, I see. I'll do as you suggested then.

sdk/src/main/scala/io/horizen/utxo/SidechainNodeViewHolder.scala
saratnt: Ok, I'll try this way then.

davehorizen (changes requested):
lgtm but if it's not a cumbersome process and if the goal of these seeder nodes is just store blocks I will proceed to disable the following rpc methods:
- `eth_call` and `eth_estimateGas` to avoid some computationally heavy operations
- the entire `txpool` namespace (`txpool_status`, `txpool_content`, `txpool_contentFrom` and `txpool_inspect`) because they will always display an empty mempool

sdk/src/main/scala/io/horizen/api/http/route/WalletBaseApiRoute.scala
saratnt: Done

i-skrypnyk:
There are some compilation error in the CI, other than that LGTM!

i-skrypnyk (approved):

saratnt:
@davehorizen Done

davehorizen (approved):

PR #833, opened 2023-05-24
SDK-1111

MarcoOl94:
Fixed vulnerability on ListSerializer.parse() method.

### DIFF
--- a/sdk/src/main/java/io/horizen/utils/ListSerializer.java
+++ b/sdk/src/main/java/io/horizen/utils/ListSerializer.java
@@ -41,7 +41,15 @@ public class ListSerializer<T extends BytesSerializable> implements SparkzSerial
         if(maxListLength > 0 && objectsCount > maxListLength)
             throw new IllegalArgumentException("Input data contains to many elements - " + objectsCount);
 
-        List<T> objectsList = new ArrayList<>(objectsCount);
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/io/horizen/utils/ListSerializer.java
```diff
@@ -41,7 +41,16 @@ public List<T> parse(Reader reader) {
         if(maxListLength > 0 && objectsCount > maxListLength)
             throw new IllegalArgumentException("Input data contains to many elements - " + objectsCount);
 
-        List<T> objectsList = new ArrayList<>(objectsCount);
+        //To avoid someone to create a malicious object which has a really big objectCount that can cause an Heap out of memory error
+        //we want to don't preallocate the ArrayList size except if this number of objects is less than 1000
```
paolocappelletti, line 45: a couple of typos :-)
we want to don't -> we don't want
we decided to lowered -> we decided to lower

paolocappelletti (approved):

alsala (approved):
For further testing we might add some lines in an existing py test such as:
`        try:`
`           decodeTransaction(sc_node, payload="0201fafbd0a00f")`
`        except Exception as e:`
`           ...`
Which without fix triggers the issue

PR #834, opened 2023-05-26
fix block priority in the NodeViewHolder mailbox

i-skrypnyk:


paolocappelletti:
not merged - already included in pr #836

PR #835, opened 2023-05-26
eth_getLogs limit bug fix

ddrvar:
[Jira ticket](https://horizenlabs.atlassian.net/browse/SDK-1117?atlOrigin=eyJpIjoiNDk3MGY5MjFiMWYwNGVlM2FkNjZkOThhNzIwZTYwZWUiLCJwIjoiaiJ9)

After looking at how different EVM-compatible blockchains put limits on eth_getLogs, we decided on the following:

- max 10K results can be returned by a single query

- query duration must not exceed 10 seconds

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
+++ b/sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
@@ -50,7 +50,7 @@ import scala.collection.JavaConverters.seqAsJavaListConverter
 import scala.collection.convert.ImplicitConversions.`collection AsScalaIterable`
 import scala.collection.mutable.ListBuffer
 import scala.concurrent.duration.FiniteDuration
-import scala.concurrent.{Await, Future}
+import scala.concurrent.{Await, Future, TimeoutException}
 import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
 import scala.concurrent.duration._
@@ -1070,7 +1070,7 @@ class EthService(
             throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, "invalid block range"))
           }
 
-          val timeout = 10.seconds
+          val timeout = networkParams.getLogsQueryTimeout
           var resultCount = 0
           // get the logs from all blocks in the range into one flat list
           val result: Try[Seq[EthereumLogView]] = Try {
@@ -1084,8 +1084,8 @@ class EthService(
                   .get
 
                 resultCount += logs.length
-                if (resultCount > 10000)
-                  throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, "Log response size exceeded. You can make eth_getLogs requests with up to a 10K response size. Limit some parameters and try again."))
+                if (resultCount > networkParams.getLogsSizeLimit)
+                  throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, "Log response size exceeded. You can make eth_getLogs requests with up to a " + networkParams.getLogsSizeLimit +  " response size. Limit some parameters and try again."))
 
                 logs
               })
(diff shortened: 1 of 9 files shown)


sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1068,15 +1070,41 @@ class EthService(
...
+                val logs = nodeView.history
+                  .blockIdByHeight(blockNumber)
+                  .map(ModifierId(_))
+                  .flatMap(nodeView.history.getStorageBlockById)
+                  .map(RpcFilter.getBlockLogs(stateView, _, query))
+                  .get
+
+                resultCount += logs.length
+                if (resultCount > networkParams.getLogsSizeLimit)
+                  throw new RpcException(RpcError.fromCode(RpcCode.InvalidParams, "Log response size exceeded. You can make eth_getLogs requests with up to a " + networkParams.getLogsSizeLimit +  " response size. Limit some parameters and try again."))
```
MarcoOl94, line 1089: Are these errors taken from other EVM chains?

sdk/src/main/scala/io/horizen/params/MainNetParams.scala
```diff
@@ -40,7 +42,9 @@ case class MainNetParams(
                           override val sidechainCreationVersion: SidechainCreationVersion = SidechainCreationVersion1,
                           override val chainId: Long = 33333333,
                           override val isCSWEnabled: Boolean = true,
-                          override val isNonCeasing: Boolean = false
+                          override val isNonCeasing: Boolean = false,
```
MarcoOl94, line 45: In this particular case I don't like to much to add network settings for just one RPC call. Can we verify if is it needed from the product side?

MarcoOl94 (changes requested):

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
davehorizen: Calling the `eth_getLogs` rpc method for an extremely large number of blocks using Infura results in the following message:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32005,
        "data": {
            "from": "0x10114DC",
            "limit": 10000,
            "to": "0x10114F8"
        },
        "message": "query returned more than 10000 results. Try with this block range [0x10114DC, 0x10114F8]."
    }
}
```
BNB chain is much less verbose:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32005,
        "message": "limit exceeded"
    }
}
```
The code used (`32005`) seems to be a non-standard rpc code and it is reported in the [EIP-1474](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md) but not directly present on _geth_. Let's discuss if necessary how to proceed in this case, personally I will keep the `32602` Invalid params error code.

ddrvar: Current error messages are taken partly from Alchemy [[source](https://docs.alchemy.com/reference/eth-getlogs)]

sdk/src/main/scala/io/horizen/params/MainNetParams.scala
ddrvar: I checked with the product side and @paolocappelletti and we agreed to put these limits in config file, precisely in `EthServiceSettings`
I created two unit tests that check response size and timeout limitations

sdk/src/main/scala/io/horizen/params/TestNetParams.scala
```diff
@@ -10,6 +10,7 @@ import io.horizen.proposition.{PublicKey25519Proposition, SchnorrProposition, Vr
 import sparkz.core.block.Block
 import sparkz.util.ModifierId
 import sparkz.util.bytesToId
+import scala.concurrent.duration._
```
MarcoOl94, line 13: Is it needed?

MarcoOl94 (approved):
LGTM, please just remove the unnecessary changes (e.g. blank line, import unused....)

davehorizen (approved):

PR #837, opened 2023-05-30
SDK-1105 - SyncStatusActor corner case

davehorizen:
Implementation of the solution reported in task [SDK-1105](https://horizenlabs.atlassian.net/browse/SDK-1105) encountered with forger nodes restarted and not quickly connected to the rest of the network.

Note: the python test result can be non-deterministic because the sync process can get stuck from time to time due to known bug  encountered in small node network.

[SDK-1105]: https://horizenlabs.atlassian.net/browse/SDK-1105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


qa/sc_evm_sync_status_forger_node_corner_case.py
```diff
@@ -0,0 +1,234 @@
...
+        assert_equal(1, status, "Make forger stake with native smart contract as owner should create a failed tx")
+
+        self.block_id = generate_next_block(sc_node0, "node 0", force_switch_to_next_epoch=True)
+        self.sc_sync_all()
+
+        self.block_id = generate_next_block(sc_node0, "node 0", force_switch_to_next_epoch=True)
+        self.sc_sync_all()
+
+        # -------------------------------------------------------------------------------------
+        # Test
```
MarcoOl94, line 161: I think this test can be simplified:

1. Node0 generate 10 blocks
2. Node 1 shut down
3. Node 1 restart and generate 1 block
4. Node0-Node1 reconnection
5. Node0 generate 1 block
6. Node1 doesn't find the parentid of this block in his list (it will only contain the newly generated block) and raise the exception

qa/sc_evm_sync_status_forger_node_corner_case.py
```diff
@@ -0,0 +1,234 @@
+#!/usr/bin/env python3
```
MarcoOl94, line 1: Is there a specific reason why we didn't include this test in the **run_sc_test.sh** script?

MarcoOl94 (changes requested):

qa/sc_evm_sync_status_forger_node_corner_case.py
davehorizen: In that workflow execution of steps 3 and 4 will ensure that the generated block will be shared and recognized by both nodes as the new tip. After which the block generated in step 5 will find correlation between those present in the `SyncStatusActor` buffer of `sc_node1`.

In the current python test implementation the `sc_node1` is stopped, restarted and then it create a couple of blocks on its own before being connected to the rest of the network, this simulate the node restart performed by the infra team.
I modified the current test flow removing the fork creation on `sc_node1` because it wasn't necessary and I also lowered the number of blocks created on `sc_node0` to 500 to 200.

qa/sc_evm_sync_status_forger_node_corner_case.py
davehorizen: No specific reason, it was an oversight. Test added to the `run_sc_test.sh` script.
I also change the python test name to a more generic `sc_evm_sync_status_forger_nodes.py` in case we need to add more test cases in the future that involve more forger nodes.

MarcoOl94 (approved):

paolocappelletti (approved):

PR #840, opened 2023-06-06
SDK-1150 - eth_feeHistory pending block tag fix

ddrvar:
[Jira ticket](https://horizenlabs.atlassian.net/browse/SDK-1150?atlOrigin=eyJpIjoiYWViZTAwOWUwNDlhNGMxNmFjYjM1ZDkxNTM5YTBlOGMiLCJwIjoiaiJ9)

I discovered that the error happens in [this](https://github.com/HorizenOfficial/Sidechains-SDK/blob/04676955c2e918ab3f0f08c5fa74b47d58c1b251/sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala#LL1020C15-L1020C48) line when the height parameter is greater than the current SC height which always happens in the last iteration when input block tag is pending.  As a result, a block is null and `NullPointerException` happens when we try to access the block header. My proposed solution includes block null check and in the last iteration uses previously computed `requestedBlock` which contains information on pending block.

### DIFF
--- a/qa/sc_evm_feeHistory.py
+++ b/qa/sc_evm_feeHistory.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+import json
+import logging
+from decimal import Decimal
+
+from SidechainTestFramework.account.ac_chain_setup import AccountChainSetup
+from SidechainTestFramework.account.ac_utils import ac_makeForgerStake
+from SidechainTestFramework.account.httpCalls.transaction.createEIP1559Transaction import createEIP1559Transaction
+from SidechainTestFramework.account.utils import convertZenToZennies
+from SidechainTestFramework.scutil import generate_next_block, \
+    assert_equal
+from test_framework.util import forward_transfer_to_sidechain, fail
+
+"""
+Check eth_feeHistory pending block tag works properly
+
+Configuration:
+    - 2 SC nodes connected with each other
+    - 1 MC node
+
+Test:
+    - Add some transactions to the SC nodes. 
+    - Get node 1 and node 2 fee histories with pending block tag
+    - Verify that they are equal and return results with valid lengths w.r.t. blocks mined and transactions
+    - Get node 1 and node 2 fee histories with latest block tag
+    - Verify that they are equal and return results with valid lengths w.r.t. blocks mined and transactions
+"""
+
+
+class SCEvmFeeHistory(AccountChainSetup):
+
+    def __init__(self):
+        super().__init__(number_of_sidechain_nodes=2)
+
+    def run_test(self):
+        mc_node = self.nodes[0]
+        sc_node_1 = self.sc_nodes[0]
+        sc_node_2 = self.sc_nodes[1]
+
+        # transfer some fund from MC to SC1 at a new evm address, then mine mc block
+        evm_address_sc1 = sc_node_1.wallet_createPrivateKeySecp256k1()["result"]["proposition"]["address"]
+        evm_address_sc2 = sc_node_2.wallet_createPrivateKeySecp256k1()["result"]["proposition"]["address"]
+
+        ft_amount_in_zen = Decimal('500.0')
+
+        forward_transfer_to_sidechain(self.sc_nodes_bootstrap_info.sidechain_id,
+                                      mc_node,
+                                      evm_address_sc1,
+                                      ft_amount_in_zen,
+                                      mc_return_address=mc_node.getnewaddress(),
+                                      generate_block=True)
+        forward_transfer_to_sidechain(self.sc_nodes_bootstrap_info.sidechain_id,
+                                      mc_node,
+                                      evm_address_sc2,
+                                      ft_amount_in_zen,
+                                      mc_return_address=mc_node.getnewaddress(),
+                                      generate_block=True)
+        self.sync_all()
+
+        generate_next_block(sc_node_1, "first node")
+        self.sc_sync_all()
+
+        nonce_addr_1 = 0
+        nonce_addr_2 = 0
+
+        sc2_blockSignPubKey = sc_node_2.wallet_createPrivateKey25519()["result"]["proposition"]["publicKey"]
+        sc2_vrfPubKey = sc_node_2.wallet_createVrfSecret()["result"]["proposition"]["publicKey"]
+
+        forgerStake_amount = 300  # Zen
+
+        makeForgerStakeJsonRes = ac_makeForgerStake(sc_node_1, evm_address_sc1, sc2_blockSignPubKey,
+                                                    sc2_vrfPubKey,
+                                                    convertZenToZennies(forgerStake_amount))
+        nonce_addr_1 += 1
+        if "result" not in makeForgerStakeJsonRes:
+            fail("make forger stake failed: " + json.dumps(makeForgerStakeJsonRes))
+        else:
+            logging.info("Forger stake created: " + json.dumps(makeForgerStakeJsonRes))
+        self.sc_sync_all()
+
+        # Generate SC blocks on SC node to make node 2 a forger
+        generate_next_block(sc_node_1, "first node", force_switch_to_next_epoch=True)
+        self.sc_sync_all()
+        generate_next_block(sc_node_1, "first node", force_switch_to_next_epoch=True)
+        self.sc_sync_all()
+        generate_next_block(sc_node_1, "first node", force_switch_to_next_epoch=True)
+        self.sc_sync_all()
+
+        common_tx_list = []
+        for i in range(10):
+            common_tx_list.append(
+                createEIP1559Transaction(sc_node_1, fromAddress=evm_address_sc1, toAddress=evm_address_sc2,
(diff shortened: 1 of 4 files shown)


ddrvar:
@saratnt Done, feeHistory returns default value now (`new EthereumFeeHistoryView()`) on input block pending
Maybe we could solve the issue then by not including the pending block in the rewards array, and keep the rest till that block

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -1003,7 +1003,7 @@ class EthService(
       // limit the range of blocks by the number of available blocks and cap at 1024
       val blocks = blockCount.intValueExact().min(requestedBlockInfo.height).min(1024)
       // geth comment: returning with no data and no error means there are no retrievable blocks
-      if (blocks < 1) {
+      if (blocks < 1 || newestBlock == "pending") {
```
saratnt, line 1006: That's OK but it is better to move the check on "pending" before calling `getBlockByTag` (line 1002). This because `getBlockByTag` will create a pending block and it is an expensive operation that it is not needed anymore.

qa/sc_evm_feeHistory.py
```diff
@@ -0,0 +1,137 @@
...
+Test:
+    - Add some transactions to the SC nodes. 
+    - Get node 1 and node 2 fee histories with pending block tag
+    - Verify that they are equal and return results with valid lengths w.r.t. blocks mined and transactions
+    - Get node 1 and node 2 fee histories with latest block tag
+    - Verify that they are equal and return results with valid lengths w.r.t. blocks mined and transactions
+"""
+
+
+class SCEvmFeeHistory(AccountChainSetup):
```
saratnt, line 30: This test is OK. It should be updated also EthServiceTest.scala with a test on "pending" tag.

saratnt (changes requested):

saratnt:
@ddrvar About your comment "Maybe we could solve the issue then by not including the pending block in the rewards array, and keep the rest till that block", it is a possibility but I think it is clearer for the user that we cannot provide info for "pending" if we don't return nothing. In addition, I think the same result can be obtained requesting "latest", so it is not like the user doesn't have other way to obtain the same info. But this is just my opinion. Maybe we can check what other blockchains do in this case or ask Victor what is his opinion.

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
ddrvar: Done, moved out of `applyOnAccountView`

qa/sc_evm_feeHistory.py
ddrvar: Done

ddrvar:
I checked a couple of other blockchains and from what I understood none of them support pending input block, and compute it as the latest tag.
* Erigon - get pending block and its receipts returns null ([source](https://github.com/maticnetwork/erigon/blob/fd37f7e9bdef4c3740ec168b92cd7c2f7e1261b6/cmd/rpcdaemon/commands/eth_system.go#LL228C1-L228C17)). Also, the tests concerning pending functionality are commented out ([source](https://github.com/maticnetwork/erigon/blob/fd37f7e9bdef4c3740ec168b92cd7c2f7e1261b6/eth/gasprice/feehistory_test.go#L52-L53))
* Geth - it was difficult to determine, but it seems to depend on backend it is connected to, not exactly sure ([source](https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/eth/gasprice/feehistory.go#L211-L212)). One of them returns null for pending blocks and receipts, while others return correct result and include rewards as well ([source1](https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/eth/gasprice/feehistory.go#L172-L180), [source2](https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/les/api_backend.go#L137-L139), [source3](https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/eth/api_backend.go#L197-L199)). Additionally, they have tests that cover pending block tag case ([source](https://github.com/ethereum/go-ethereum/blob/281e8cd5abaac86ed3f37f98250ff147b3c9fe62/eth/gasprice/feehistory_test.go#L51-L52))
* Hyperledger - they compute pending tag as if it latest ([source](https://github.com/hyperledger/besu/blob/c1cfaf462ac6324bc874635e4f1a8dcca41a9452/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistory.java#L80-L84)). Also, the tests don't cover the pending tag at all ([source](https://github.com/hyperledger/besu/blob/c1cfaf462ac6324bc874635e4f1a8dcca41a9452/ethereum/api/src/test/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthFeeHistoryTest.java#L77-L82))

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -998,6 +998,8 @@ class EthService(
       rewardPercentiles: Array[Double]
   ): EthereumFeeHistoryView = {
     val percentiles = sanitizePercentiles(rewardPercentiles)
+    if (newestBlock == "pending")
```
saratnt, line 1001: At this point, I think the check on pending can be moved before `sanitizePercentiles` too, so we totally avoid any needless computation.

saratnt:
@ddrvar Good analysis! Now we can ask Victor his opinion.

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
ddrvar: Done

ddrvar:
As per [Victor's comment](https://horizenlabs.atlassian.net/browse/SDK-1150?focusedCommentId=21385), we decided to return the same result as the latest for the pending block. Tests updated accordingly.
I realize that I have not been clear in my description of how Erigon handles pending block tag, but they do it in the same way as Hyperledger. I updated Victor about it just to be on the same page with the solution which he confirmed.

qa/sc_evm_feeHistory.py
```diff
@@ -127,10 +127,17 @@ def run_test(self):
...
+
         assert_equal(history_pending_sc_node_1['result'], history_pending_sc_node_2['result'], "fee history should be equal among connected nodes")
-        assert_equal(None, history_pending_sc_node_1['result']['baseFeePerGas'])
-        assert_equal(None, history_pending_sc_node_1['result']['gasUsedRatio'])
-        assert_equal(None, history_pending_sc_node_1['result']['reward'])
+        assert_equal(history_latest_sc_node_1['result'], history_latest_sc_node_2['result'], "fee history should be equal among connected nodes")
+        assert_equal(history_pending_sc_node_1['result'], history_latest_sc_node_1['result'], "fee history should be equal for pending and latest nodes")
+
+        assert_equal('0x1', history_pending_sc_node_1['result']['oldestBlock'])
+        assert_equal(9, len(history_pending_sc_node_1['result']['baseFeePerGas']))
```
saratnt, line 138: If it is not too much effort, it would be nice to link the expected length of  _gasUsedRatio_, _reward_ and _baseFeePerGas_ result arrays with the height of the blockchain. I mean, the feeHistory requests the latest 16 blocks fee info but it returns just 8. It would be nice to see  where 8 (and 9) number comes from.

ddrvar: Done

saratnt: OK, just add a comparison between the height and the number of blocks requested, so the expected result length is the minimum between the 2. E.g.
```
blockCount = 16
history_pending_sc_node_1 = sc_node_1.rpc_eth_feeHistory(str(int(blockCount,16)), "pending", [25.0, 75.0])

....
expected_result_length = blockCount 
if (blockCount> height)
       expected_result_length = height
....
assert_equal(expected_result_length, len(history_pending_sc_node_1['result']['gasUsedRatio']))
```
It's just an example

saratnt (changes requested):

qa/sc_evm_feeHistory.py
ddrvar: Done

saratnt (approved):

davehorizen (approved):
Looks good to me!
Just a non blocking comment regarding the `sc_evm_feeHistory` python test, given its generic name I will expect also a call the `eth_feeHistory` rpc method with "standard" request parameters (like a finalized block number) that can be used as a non-regression check. The test with the pending tag can be seen as a corner case for this method. 
Proceed to add such cases only if possible in terms of time, up to you! ;)

ddrvar:
@davehorizen extended python test with specific block number test case, as well as 'earliest' block tag

PR #843, opened 2023-06-08
Sdk 1045 - Account simple app sc2sc message exchange

paologalligit:


tools/sctool/src/main/java/io/horizen/ScBootstrappingToolCommandProcessor.java
```diff
@@ -6,7 +6,32 @@
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.horizen.account.sc2sc.ScTxCommitmentTreeRootHashMessageProcessor;
```
alsala, line 9: lots of unused imports

alsala (changes requested):

sdk/src/main/java/io/horizen/cryptolibprovider/CommonCircuit.java
```diff
@@ -71,8 +71,22 @@ public static WithdrawalCertificate createWithdrawalCertificate(WithdrawalEpochC
         );
     }
 
+    public static WithdrawalCertificate createWithdrawalCertificateWithBtrFreeAndFtMinAmountSwapped(WithdrawalEpochCertificate cert, Enumeration.Value sidechainCreationVersion) {
```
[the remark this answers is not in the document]
paologalligit: Actually it's more general purpose than only sc2sc; the `createWithdrawalCertificate` should not be called every time, but only when we need to swap the `btrFee` and `ftMinAmount` (see [here](https://github.com/HorizenOfficial/Sidechains-SDK/blob/dev/sdk/src/main/scala/io/horizen/block/SidechainCommitmentTree.scala#L74)) otherwise we double swap them ending up with the two parameters in the initial order

sdk/src/main/scala/io/horizen/certificatesubmitter/strategies/WithKeyRotationCircuitStrategy.scala
```diff
@@ -118,13 +119,27 @@ class WithKeyRotationCircuitStrategy[
     val keysAndSignatures: SchnorrKeysSignatures = getSchnorrKeysSignatures(state, referencedWithdrawalEpochNumber)
     val keysRootHash: Array[Byte] = circuit.getSchnorrKeysHash(keysAndSignatures)
 
-    val sc2ScDataForCertificate: Option[Sc2ScDataForCertificate] = sc2scConfig.canSendMessages match {
-      case true => Some(getDataForCertificateCreation(referencedWithdrawalEpochNumber, state, history, params))
-      case false => None
+    val sc2ScDataForCertificate: Option[Sc2ScDataForCertificate] =  if (sc2scConfig.canSendMessages) {
```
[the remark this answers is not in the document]
paologalligit: yes, this is something that is introduced in another PR (sc2sc not inherited from circuitStrategy trait but only related to WithKeyRotation)

sdk/src/main/java/io/horizen/cryptolibprovider/CommonCircuit.java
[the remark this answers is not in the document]
alsala: Ok, it might be called in future by other modules, therefore it make sense to leave it there

qa/cross_chain_evm_to_be_deleted.py
```diff
@@ -0,0 +1,130 @@
...
+            sc2sc_verification_key_file_name="verification"
+        )
+
+    def do_send_raw_tx(self, raw_tx, evm_signer_address):
+        sc_node = self.sc_nodes[0]
+        signed_raw_tx = signTransaction(sc_node, fromAddress=evm_signer_address, payload=raw_tx)
+
+        tx_hash = sendTransaction(sc_node, payload=signed_raw_tx)
+        return tx_hash
+
```
alsala, line 37: Consider adding:

`def sc_setup_chain(self):`
`       # mandatory for this test`
`       self.options.nonceasing = True`
`       super().sc_setup_chain()`


for avoiding to specify this mandatory setting via the `--nonceasing` flag

alsala (approved):

qa/cross_chain_evm_to_be_deleted.py
paologalligit: thanks, done :+1:

examples/account/evmapp/src/main/java/io/horizen/examples/api/VoteController.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public String basePath() {
+        return VOTING_BASE_PATH;
+    }
+
+    @Override
+    public List<Route> getRoutes() {
+        List<Route> routes = new ArrayList<>();
+        routes.add(bindPostRequest("sendVoteMessage", this::sendVoteMessage, SendVoteMessageRequest.class));
+        routes.add(bindPostRequest("redeem", this::redeemVoteMessage, RedeemVoteMessageRequest.class));
```
MarcoOl94, line 32:
```suggestion
        routes.add(bindPostRequest("redeemVoteMessage", this::redeemVoteMessage, RedeemVoteMessageRequest.class));
```

examples/account/evmapp/src/main/java/io/horizen/examples/messageprocessor/VoteMessageProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+import io.horizen.account.sc2sc.AccountCrossChainMessage;
+import io.horizen.account.state.*;
+import io.horizen.evm.Address;
+import io.horizen.examples.messageprocessor.decoder.SendVoteCmdInputDecoder;
+import sparkz.crypto.hash.Keccak256;
+
+import static io.horizen.account.abi.ABIUtil.*;
+
+public class VoteMessageProcessor extends AbstractCrossChainMessageProcessor {
+    public static final Address contractAddress = CustomWellknownAddress.VOTE_SMART_CONTRACT_ADDRESS;
```
MarcoOl94, line 13: Not used

examples/account/evmapp/src/main/java/io/horizen/examples/messageprocessor/VoteRedeemMessageProcessor.java
```diff
@@ -0,0 +1,61 @@
...
+import io.horizen.examples.messageprocessor.decoder.RedeemSendVoteCmdInputDecoder;
+import scala.Option;
+import sparkz.crypto.hash.Keccak256;
+
+import static io.horizen.account.abi.ABIUtil.*;
+
+public class VoteRedeemMessageProcessor extends AbstractCrossChainRedeemMessageProcessor {
+    public final static String REDEEM_SEND_VOTE = getABIMethodId("redeemSendVote(uint32,bytes20,bytes20,bytes20,bytes1,bytes20,bytes20,bytes20,bytes20,bytes)");
+    private SidechainSettings sidechainSettings;
+    private Sc2scCircuit sc2scCircuit;
```
MarcoOl94, line 19: Not used

qa/SidechainTestFramework/account/httpCalls/transaction/simpleapp/decodeShowAllVotes.py
```diff
@@ -0,0 +1,8 @@
+import json
+
+
+def decodeShowAllVotes(sidechain_node, bytes):
```
MarcoOl94, line 4: Sorry I don't find where it is used

qa/httpCalls/transaction/vote/redeem.py
```diff
@@ -0,0 +1,24 @@
+import json
+
+
+def redeem(sidechainNode, proposition, certificateDataHash, nextCertificateDataHash, scCommitmentTreeRoot,
```
MarcoOl94, line 4:
```suggestion
def redeemVoteMessage(sidechainNode, proposition, certificateDataHash, nextCertificateDataHash, scCommitmentTreeRoot,
```

sdk/src/main/scala/io/horizen/account/state/events/AddCrossChainMessage.scala
```diff
@@ -2,16 +2,16 @@ package io.horizen.account.state.events
...
-import org.web3j.abi.datatypes.DynamicBytes
-import org.web3j.abi.datatypes.generated.Uint32
+import org.web3j.abi.datatypes.generated.{Bytes20, Bytes32, Bytes4, Uint32}
+import org.web3j.abi.datatypes.{Address => AbiAddress}
 
 import scala.annotation.meta.getter
 
-case class AddCrossChainMessage(@(Parameter@getter)(1) @(Indexed@getter) sender: Address,
+case class AddCrossChainMessage(@(Parameter@getter)(1) @(Indexed@getter) sender: AbiAddress,
                                 @(Parameter@getter)(2) @(Indexed@getter) messageType: Uint32,
```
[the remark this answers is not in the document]
paologalligit: Yes, that's correct: the `AddCrossChainMessage` has a `Uint32` parameter but its apply method takes a simple Int and map it to `Uint32` (just like byte arrays are mapped to `Bytes32` or `Bytes20`)

sdk/src/main/scala/io/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -368,7 +369,7 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
           case _ => false
         }
         if (isWithdrawalEpochSwitched) {
-          val certEpochNumberToRemove: Int = epochInfo.epoch - 4
+          val certEpochNumberToRemove: Int = epochInfo.epoch - 8
```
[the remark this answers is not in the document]
paologalligit: because we are removing certificates of previous epochs we need for cross chain messages validations, we need to double at least the epoch interval

sdk/src/main/scala/io/horizen/account/sc2sc/AbstractCrossChainRedeemMessageProcessor.scala
```diff
@@ -56,15 +68,13 @@ abstract class AbstractCrossChainRedeemMessageProcessor(
    */
   protected def getAccountCrossChainRedeemMessageFromMessage(msg: Message): AccountCrossChainRedeemMessage
 
-  private def validateRedeemMsg(ccRedeemMgs: AccountCrossChainRedeemMessage, view: BaseAccountStateView): Unit = {
+  private def validateRedeemMsg(ccRedeemMgs: AccountCrossChainRedeemMessage, accountCcMsg: AccountCrossChainMessage, view: BaseAccountStateView): Unit = {
```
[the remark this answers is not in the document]
paologalligit: You're correct, this is stale data, replacing it with CrossChainMessage to validate we're not double redeeming it

sdk/src/test/scala/io/horizen/account/sc2sc/AccountCrossChainRedeemMessageTest.scala
```diff
@@ -0,0 +1,33 @@
+package io.horizen.account.sc2sc
+
+import io.horizen.account.sc2sc.CrossChainRedeemMessageProcessorImpl.receiverSidechain
+import io.horizen.utils.BytesUtils
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class AccountCrossChainRedeemMessageTest {
```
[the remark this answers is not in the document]
paologalligit: If we want to produce a negative test we need to have wrong parameters for the encoder. This is gonna trigger the validator, right after the constructor, that is not injected since we don't allow the possibility to have different validators from the default one. So we should mock the static validator to bypass it and have a malformed byte array. But this is not gonna happen in the real scenario where the validator's always gonna run. So I think it's better not to add such test case

MarcoOl94 (approved):

PR #846, opened 2023-06-09
CI/CD updated for publishing new projects

ysibirski:
I have updated existing CICD pipeline to the better version of it + included two more projects to be published requested by @alsala via modified pom.xml files that includes publishing configuration.

- Currently publishing `'sdk' 'tools/sctool' 'tools/sidechains-sdk-account_sctools'`
- ./ci/devtools/set_version.sh was updated as well. The script will update version in all the places where needed

P.S. `mvn test` command would fail from time to time :shrug: 
https://app.travis-ci.com/github/HorizenOfficial/Sidechains-SDK/builds/263685078


README.md
```diff
@@ -98,5 +98,5 @@ In order to build and use `SNAPSHOT.jar` package version refer to the following
 
 ## Backward Compatibility
 
-Version 0.7.0 is not backward compatible with any previous versions.
-Due to a change in the consensus protocol, you cannot update any existing Sidechain of version <=0.6.0 into a >=0.7.0
\ No newline at end of file
+Version 0.8.0 is not backward compatible with any previous versions.
```
alsala, line 101: We should use something like 'Any version >= 0.7.0 is not backward compatible', and the same below, we should keep the old comment.

ci/publish.sh
```diff
@@ -1,16 +1,35 @@
 #!/bin/bash
 set -eo pipefail
+
 retval=0
+projects_to_publish=('sdk' 'tools/sctool' 'tools/sidechains-sdk-account_sctools')
```
alsala, line 5: As discussed with @paolocappelletti , we should publish also `tools/sidechains-sdk-utxo_sctools`

alsala (changes requested):
Pls add also `tools/sidechains-sdk-utxo_sctools` to the list of jar to be published

paolocappelletti:
hi @ysibirski this PR should depart from dev branch (0.7.0 tags) - can you please rebase it?
the ZenDao branch will be later reabsed over dev once this PR will be merged there

README.md
ysibirski: I have used `/ci/devtools/set_version.sh` script and it only changes VERSION number in various places. The exact wording will still need to be changed manually by someone who knows what exactly it should be

PR #859, opened 2023-06-16
SDK-1221 Reject TXs below baseFee fork value from mempool

JcBernack:
See [SDK-1221](https://horizenlabs.atlassian.net/browse/SDK-1221)

[SDK-1221]: https://horizenlabs.atlassian.net/browse/SDK-1221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
+++ b/sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
@@ -2,9 +2,10 @@ package io.horizen.account.mempool
 
 import io.horizen.account.api.rpc.types.EthereumTransactionView
 import io.horizen.account.block.AccountBlock
+import io.horizen.account.fork.GasFeeFork
 import io.horizen.account.mempool.MempoolMap._
 import io.horizen.account.mempool.TxExecutableStatus.TxExecutableStatus
-import io.horizen.account.mempool.exception.{AccountMemPoolOutOfBoundException, NonceGapTooWideException, TransactionReplaceUnderpricedException, TxOversizedException}
+import io.horizen.account.mempool.exception._
 import io.horizen.account.proposition.AddressProposition
 import io.horizen.account.state.{AccountStateReaderProvider, BaseStateReaderProvider}
 import io.horizen.account.transaction.EthereumTransaction
@@ -59,7 +60,7 @@ class MempoolMap(
     if (!contains(ethTransaction.id)) {
       val account = ethTransaction.getFrom
 
-      //Reject transactions that are too big
+      // Reject transactions that are too big
       val txSize = ethTransaction.size()
       if (txSize > MaxTxSize) {
         log.trace(s"Transaction $ethTransaction size exceeds maximum allowed size: current size $txSize, " +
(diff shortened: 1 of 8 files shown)


sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -59,15 +60,21 @@ class MempoolMap(
...
       if (txSize > MaxTxSize) {
         log.trace(s"Transaction $ethTransaction size exceeds maximum allowed size: current size $txSize, " +
           s"maximum size: $MaxTxSize")
         throw TxOversizedException(account.asInstanceOf[AddressProposition].address(), txSize)
       }
 
-      //Reject transactions with a nonce gap too big
+      // Reject transaction with max fee less than the current minimum base fee
+      val currentEpochNumber = baseStateReaderProvider.getBaseStateReader().getConsensusEpochNumber.getOrElse(0)
+      if (GasFeeFork.get(currentEpochNumber).baseFeeMinimum.compareTo(ethTransaction.getMaxFeePerGas) > 0) {
```
paolocappelletti, line 73: can you double check that with old Legacy (Type 0) txs the check is working properly? (maybe add a unit test for that)

paolocappelletti (changes requested):

sdk/src/main/scala/io/horizen/account/mempool/MempoolMap.scala
```diff
@@ -59,15 +60,21 @@ class MempoolMap(
...
       val txSize = ethTransaction.size()
       if (txSize > MaxTxSize) {
         log.trace(s"Transaction $ethTransaction size exceeds maximum allowed size: current size $txSize, " +
           s"maximum size: $MaxTxSize")
         throw TxOversizedException(account.asInstanceOf[AddressProposition].address(), txSize)
       }
 
-      //Reject transactions with a nonce gap too big
+      // Reject transaction with max fee less than the current minimum base fee
+      val currentEpochNumber = baseStateReaderProvider.getBaseStateReader().getConsensusEpochNumber.getOrElse(0)
```
MarcoOl94, line 72: There is another similar check related to the gas limit of the transaction based on a parameter taken from the ForkManager inside the AccountState (https://github.com/HorizenOfficial/Sidechains-SDK/blob/dev/sdk/src/main/scala/io/horizen/account/state/AccountState.scala#L469). Have you considered to add the additional check here instead of the Mempool?

JcBernack: Good point. Actually, a change in the `validate` method is necessary, otherwise a custom node could still include TXs with lower-than-minimum fees in a block.

sdk/src/test/scala/io/horizen/account/mempool/MempoolMapUpdateTest.scala
```diff
@@ -49,6 +51,8 @@ class MempoolMapUpdateTest extends JUnitSuite with EthereumTransactionFixture wi
     Mockito.reset(rejectedBlock)
     Mockito.reset(appliedBlock)
 
+    Mockito.when(baseStateViewMock.getConsensusEpochNumber).thenReturn(Some(intToConsensusEpochNumber(0)))
```
MarcoOl94, line 54: Is it needed?

MarcoOl94 (approved):

PR #860, opened 2023-06-19
SDK 809 - NCC Audit suggestion: generateNextSecret

ddrvar:
[Jira ticket](https://horizenlabs.atlassian.net/browse/SDK-809?atlOrigin=eyJpIjoiYmFlZmRjYWM4ZDNiNDkxN2FiZGExOTQxNjc0MTMzNmUiLCJwIjoiaiJ9)

### DIFF
--- a/sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
+++ b/sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
@@ -143,25 +143,14 @@ class SidechainSecretStorage(storage: Storage, sidechainSecretsCompanion: Sidech
     }
   }
 
-
-  def storeNonce(nonce: Int, keyTypeSalt: Array[Byte]): Try[SidechainSecretStorage] = Try {
-    require(nonce >= 0, "Nonce must be not negative")
-    require(keyTypeSalt != null, "Key type salt must be NOT NULL")
-    val updateList = new JArrayList[JPair[ByteArrayWrapper, ByteArrayWrapper]]()
-    val removeList = new JArrayList[ByteArrayWrapper]()
-    updateList.add(new JPair(getNonceKey(keyTypeSalt), new ByteArrayWrapper(Ints.toByteArray(nonce))))
-    storage.update(new ByteArrayWrapper(Utils.nextVersion), updateList, removeList)
-    this
-  }
-
   def storeSecretAndNonceAtomic(secret: SidechainTypes#SCS, nonce: Int, keyTypeSalt: Array[Byte]): Try[SidechainSecretStorage] = Try  {
     require(nonce >= 0, "Nonce must be not negative")
     require(keyTypeSalt != null, "Key type salt must be NOT NULL")
     require(secret != null, "Can not add to storage: Secret must be NOT NULL.")
 
     val key = calculateKey(secret.publicImage())
-    val value = new ByteArrayWrapper(sidechainSecretsCompanion.toBytes(secret))
     require(!secrets.contains(key), "Key already exists - " + secret)
+    val value = new ByteArrayWrapper(sidechainSecretsCompanion.toBytes(secret))
 
     val updateList = new JArrayList[JPair[ByteArrayWrapper, ByteArrayWrapper]]()
     val removeList = new JArrayList[ByteArrayWrapper]()
--- a/sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
+++ b/sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
@@ -153,4 +153,25 @@ class SidechainSecretStorage(storage: Storage, sidechainSecretsCompanion: Sidech
     storage.update(new ByteArrayWrapper(Utils.nextVersion), updateList, removeList)
     this
   }
+
+  def storeSecretAndNonceAtomic(secret: SidechainTypes#SCS, nonce: Int, keyTypeSalt: Array[Byte]): Try[SidechainSecretStorage] = Try  {
+    require(nonce >= 0, "Nonce must be not negative")
+    require(keyTypeSalt != null, "Key type salt must be NOT NULL")
+    require(secret != null, "Can not add to storage: Secret must be NOT NULL.")
+
+    val key = calculateKey(secret.publicImage())
+    val value = new ByteArrayWrapper(sidechainSecretsCompanion.toBytes(secret))
+    require(!secrets.contains(key), "Key already exists - " + secret)
+
+    val updateList = new JArrayList[JPair[ByteArrayWrapper, ByteArrayWrapper]]()
+    val removeList = new JArrayList[ByteArrayWrapper]()
+
+    updateList.add(new JPair(key, value))
+    updateList.add(new JPair(getNonceKey(keyTypeSalt), new ByteArrayWrapper(Ints.toByteArray(nonce))))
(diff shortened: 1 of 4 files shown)


sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
```diff
@@ -153,4 +153,25 @@ class SidechainSecretStorage(storage: Storage, sidechainSecretsCompanion: Sidech
     storage.update(new ByteArrayWrapper(Utils.nextVersion), updateList, removeList)
     this
   }
+
+  def storeSecretAndNonceAtomic(secret: SidechainTypes#SCS, nonce: Int, keyTypeSalt: Array[Byte]): Try[SidechainSecretStorage] = Try  {
```
saratnt, line 157: It seems that `storeNonce` method is not used anymore so maybe it is better to remove it.

sdk/src/main/scala/io/horizen/wallet/AbstractWallet.scala
```diff
@@ -116,19 +116,16 @@ abstract class AbstractWallet[
...
-          case Success(_) =>
-            secretStorage.storeNonce(nonce, salt) match {
-              case Success(_) => return Success(this, secret)
-              case Failure(exception) => throw new RuntimeException("Can't store nonce while generating next secret " + exception)
-            }
-          case Failure(exception) => throw new RuntimeException("Can't store secret while generating next secret " + exception)
+        secretStorage.storeSecretAndNonceAtomic(secret, nonce, salt) match {
+          case Success(_) => return Success(this, secret)
+          case Failure(exception) =>
+            throw new RuntimeException("Can't store secret and nonce while generating next secret " + exception)
```
saratnt, line 126: It would be nice to pass exception to the  RuntimeException costructor, e.g. `new RuntimeException("Can't store secret and nonce while generating next secret " + exception, exception)`
This way the exception stack trace is preserved and it can be used for debugging.

sdk/src/test/scala/io/horizen/account/wallet/AccountWalletTest.scala
```diff
@@ -209,8 +209,8 @@ class AccountWalletTest
   @Test
   def testGenerateSecretsOfDifferentDomainsIndependent(): Unit = {
```
saratnt, line 210: At line 262 the assert message is wrong (probably a copy-paste)

sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
```diff
@@ -153,4 +153,25 @@ class SidechainSecretStorage(storage: Storage, sidechainSecretsCompanion: Sidech
...
+    require(nonce >= 0, "Nonce must be not negative")
+    require(keyTypeSalt != null, "Key type salt must be NOT NULL")
+    require(secret != null, "Can not add to storage: Secret must be NOT NULL.")
+
+    val key = calculateKey(secret.publicImage())
+    val value = new ByteArrayWrapper(sidechainSecretsCompanion.toBytes(secret))
+    require(!secrets.contains(key), "Key already exists - " + secret)
```
saratnt, line 164: This check can be moved before calculating value.

sdk/src/test/scala/io/horizen/integration/storage/SidechainSecretStorageTest.scala
```diff
@@ -207,4 +207,61 @@ class SidechainSecretStorageTest
...
+    val s2 = sidechainSecretStorage.get(customSecret.publicImage())
+    assertTrue(result2.isSuccess)
+    assertTrue("Storage must contain added Secret.", s2.isDefined)
+    assertEquals("Secret in storage must be the same as added.", customSecret, s2.get)
+    assertEquals("Storage must contain 2 secrets.", 2, sidechainSecretStorage.getAll.size)
+
+    val n2 = sidechainSecretStorage.getNonce(salt2)
+    assertTrue("Storage must contain nonce.", n2.isDefined)
+    assertEquals("Nonce in storage must be the same as added.", nonce2, n2.get)
+
```
saratnt, line 250: I would also add a test with a salt that is not `PrivateKey25519Creator.getInstance().salt()`, just to check that there can be the same nonce with different salts, if it is not too difficult.

saratnt (changes requested):
For me it's ok :). I have just some super minor comments.

sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
ddrvar: Done

sdk/src/main/scala/io/horizen/wallet/AbstractWallet.scala
ddrvar: Done

sdk/src/main/scala/io/horizen/storage/SidechainSecretStorage.scala
ddrvar: Done

sdk/src/test/scala/io/horizen/integration/storage/SidechainSecretStorageTest.scala
ddrvar: Done
I added a test which stores schnorr secret and salt, with the same nonce

sdk/src/test/scala/io/horizen/account/wallet/AccountWalletTest.scala
saratnt: The message is still wrong because it is not a schnorr key but a 25519 key

saratnt (changes requested):

sdk/src/test/scala/io/horizen/account/wallet/AccountWalletTest.scala
ddrvar: Done

saratnt (approved):

davehorizen (approved):
Looks good to me. I only had doubts regarding the use of the same storage to save different domain information (nonces and secrets), however this issue is external to the purpose of this PR and has not been reported as an issue.

sdk/src/main/scala/io/horizen/wallet/AbstractWallet.scala
```diff
@@ -116,19 +116,16 @@ abstract class AbstractWallet[
       case None => 0
     }
     for (_ <- 0 to secretsNumber) {
+      nonce += 1
```
paolocappelletti, line 119: I think we are introducing a non backward compatible change here
the first secret must be generated with nounce = 0 , while now we are always skipping it.
change this line with:
if (nonce > 0) { nonce += 1}

paolocappelletti (changes requested):

sdk/src/main/scala/io/horizen/wallet/AbstractWallet.scala
ddrvar: Done

paolocappelletti: sorry I realized we are inside a for loop  and if nonce = 0 and  secrtestNumber is > 1 this soultion does not work
let's try to rearrange the code

ddrvar: Done
If condition put before the for loop and nonce incremented at the end of the loop

paolocappelletti (approved):

PR #876, opened 2023-06-28
zendao fork handling

alsala:
Added support for reading epoch number in message processors .
Added support for fork handling in mc address ownership native smart contract (zenDAO). The native smart contract initialization takes place as soon as the fork activation is detected; before that fork point, the message processor can not process any message.


sdk/src/main/java/io/horizen/account/state/MessageProcessor.java
```diff
@@ -14,10 +14,10 @@
 public interface MessageProcessor {
     // Initialization is going to happen only once at genesis State creation.
     // Common pattern: declare a new native smart contract account in the View
-    void init(BaseAccountStateView view) throws MessageProcessorInitializationException;
+    void init(AccountStateView view) throws MessageProcessorInitializationException;
```
saratnt, line 17: MessageProcessors must not access to MetadataStorage. The problem is that MetadataStorage doesn't have snapshots, so in case a transaction fails when a block is applied, there is no way to roll back to the previous snapshot.

saratnt (changes requested):

sdk/src/main/java/io/horizen/account/state/MessageProcessor.java
alsala: You are right, I am now using the consensus epoch number in the message processor interfaces. Consider that `process() ` method already has this info in the block context obj parameter

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
```diff
@@ -38,24 +40,65 @@ case class McAddrOwnershipMsgProcessor(params: NetworkParams) extends NativeSmar
   override val contractAddress: Address = MC_ADDR_OWNERSHIP_SMART_CONTRACT_ADDRESS
   override val contractCode: Array[Byte] = Keccak256.hash("McAddrOwnershipSmartContractCode")
 
-  // ecdsa curve y^2 mod p = (x^3 + 7) mod p
-  val ecParameters: X9ECParameters = SECNamedCurves.getByName("secp256k1")
+  override def init(view: BaseAccountStateView, consensusEpochNumber: Int): Unit = {
+    if (!isForkActive(consensusEpochNumber)) {
+      log.warn(s"Can not perform ${getClass.getName} initialization, fork is not active")
```
paolocappelletti, line 45: I would remove the warn (if a sidechain does never activate the fork would end up with 1 useless warn per block)

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
```diff
@@ -38,24 +40,65 @@ case class McAddrOwnershipMsgProcessor(params: NetworkParams) extends NativeSmar
...
+      return
+    }
+    else {
```
paolocappelletti, line 48: are you sure is needed this else?  I would remove it , should never happen since this init() method can be called:
- at genesys block : the init for sure is not happened befor
- from the canProcess of this class: there is already a if (!initDone(view)) immediately before there

paolocappelletti (changes requested):

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
```diff
@@ -38,24 +40,65 @@ case class McAddrOwnershipMsgProcessor(params: NetworkParams) extends NativeSmar
...
     view.updateAccountStorage(contractAddress, LinkedListTipKey, LinkedListNullValue)
   }
 
+  override def canProcess(msg: Message, view: BaseAccountStateView, consensusEpochNumber: Int): Boolean = {
+    if (isForkActive(consensusEpochNumber)) {
+      // the gas cost of these calls is not taken into account in this case, we are not tracking gas consumption (and
+      // there is not an account to charge anyway)
+      if (!initDone(view))
+        init(view, consensusEpochNumber)
+      super.canProcess(msg, view, consensusEpochNumber)
```
saratnt, line 93: Maybe `super.canProcess(msg, view, consensusEpochNumber) `can be moved at the beginning of canProcess, so it returns faster in case the transaction shouldn't be handled by this message processor.

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
```diff
@@ -38,24 +40,65 @@ case class McAddrOwnershipMsgProcessor(params: NetworkParams) extends NativeSmar
...
+    } else {
+      // we can not handle anything before fork activation, but just warn if someone is trying to use it
+      if (msg.getTo.isPresent && msg.getTo.get.equals(contractAddress))
```
saratnt, line 96: Maybe it can be substituted with super.canProcess(msg, view, consensusEpochNumber)

sdk/src/test/scala/io/horizen/account/forger/AccountForgeMessageBuilderTest.scala
```diff
@@ -398,15 +398,16 @@ class AccountForgeMessageBuilderTest
...
+          ArgumentMatchers.any[Int]
         )
       )
       .thenReturn(true)
     Mockito
       .when(
         mockMsgProcessor.process(
           ArgumentMatchers.any[Message],
-          ArgumentMatchers.any[BaseAccountStateView],
+          ArgumentMatchers.any[AccountStateView],
```
saratnt, line 410: This should come back to be a BaseAccountStateView.

sdk/src/test/scala/io/horizen/account/utils/AccountMockDataHelper.scala
```diff
@@ -437,10 +437,10 @@ case class AccountMockDataHelper(genesis: Boolean)
   private def setupMockMessageProcessor = {
     val mockMsgProcessor = mock[MessageProcessor]
     Mockito
-      .when(mockMsgProcessor.canProcess(any[Message], any[BaseAccountStateView]))
+      .when(mockMsgProcessor.canProcess(any[Message], any[AccountStateView], any[Int]))
```
saratnt, line 440: This should be a BaseAccountStateView, not an AccountStateView, same below.

saratnt (changes requested):

tools/sidechains-sdk-account_sctools/src/main/java/io/horizen/AppForkConfigurator.java
```diff
@@ -1,11 +1,26 @@
...
 public class AppForkConfigurator extends ForkConfigurator {
     @Override
     public SidechainForkConsensusEpoch fork1activation() {
         return new SidechainForkConsensusEpoch(0, 0, 0);
     }
+
+    @Override
+    public List<Pair<SidechainForkConsensusEpoch, OptionalSidechainFork>> getOptionalSidechainForks() {
+        return List.of(
+                // TODO
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
alsala: I agree, that should not happen. Yet I would prefer having an exception for troubleshooting any possible code change in future.

sdk/src/main/scala/io/horizen/account/state/AccountState.scala
```diff
@@ -449,7 +450,9 @@ class AccountState(
 
   override def isEoaAccount(address: Address): Boolean = using(getView)(_.isEoaAccount(address))
 
-  override def isSmartContractAccount(address: Address): Boolean = using(getView)(_.isSmartContractAccount(address))
+  override def isEvmSmartContractAccount(address: Address): Boolean = using(getView)(_.isEvmSmartContractAccount(address))
```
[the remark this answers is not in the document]
alsala: Agree, the approach would be dangerous. I removed those methods and change back also the original name. The approach now is to let Eoa msg processor to consume any pre-fork transaction to the native smart contract (if any); consider that since fork dependant native smart contracts are not initialized at genesis state, their msg processor must be placed before the Eoa msg processor.

sdk/src/test/scala/io/horizen/account/utils/AccountMockDataHelper.scala
alsala: Done

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
alsala: Ok, I reworked a bit this method

sdk/src/main/scala/io/horizen/account/state/McAddrOwnershipMsgProcessor.scala
alsala: Ok, I reworked a bit this method

sdk/src/main/scala/io/horizen/account/api/rpc/service/EthService.scala
```diff
@@ -360,7 +359,11 @@ class EthService(
       val (success, reverted) = check(highBound)
       if (!success) {
         val error = reverted
-          .map(err => RpcError.fromCode(RpcCode.ExecutionReverted, Numeric.toHexString(err.returnData)))
+          .map(err => {
+              log.warn(s"Execution has been reverted: ${err.getMessage}", err)
```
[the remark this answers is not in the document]
alsala: Done

examples/account/evmapp/src/main/java/io/horizen/examples/AppForkConfigurator.java
```diff
@@ -20,13 +21,18 @@ public List<Pair<SidechainForkConsensusEpoch, OptionalSidechainFork>> getOptiona
...
-                )
+                    new SidechainForkConsensusEpoch(4, 4, 4),
+                    new GasFeeFork(
+                            BigInteger.valueOf(20000000),
+                            BigInteger.valueOf(2),
+                            BigInteger.valueOf(8),
+                            BigInteger.ZERO
+                    )
+            ),
+            // an analogous fork must be included in bootstrapping tool fork configurator
```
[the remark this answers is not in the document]
alsala: Done

sdk/src/main/scala/io/horizen/account/utils/WellKnownAddresses.scala
```diff
@@ -10,4 +10,11 @@ object WellKnownAddresses {
   val CERTIFICATE_KEY_ROTATION_SMART_CONTRACT_ADDRESS: Address = new Address("0x0000000000000000000044444444444444444444")
   val MC_ADDR_OWNERSHIP_SMART_CONTRACT_ADDRESS: Address = new Address("0x0000000000000000000088888888888888888888")
 
+  val listOfNativeSmartContractAddresses = Seq(
```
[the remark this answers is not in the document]
alsala: You are right, btw I removed this container since it is not used anymore

tools/sidechains-sdk-account_sctools/src/main/java/io/horizen/AppForkConfigurator.java
```diff
@@ -1,11 +1,26 @@
...
+                // this object is used by the bootstrapping tool, if the fork is not active at genesis then any
+                // epoch > 0 will do, yet the best thing would be setting the correct fork value for each network type
+                new Pair(new SidechainForkConsensusEpoch(1, 1, 1), new ZenDAOFork(true))
```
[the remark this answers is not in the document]
alsala: Right, btw this piece of code has been removed

saratnt (approved):

paolocappelletti (approved):

PR #877, opened 2023-06-30
St/sdk 826

saratnt:
## Description
See Jira Ticket SDK-826 

## Jira Ticket
SDK-826 (https://horizenlabs.atlassian.net/browse/SDK-826?atlOrigin=eyJpIjoiYWYxNDRlZmMwMWYwNDFiMGFiMDhkOWIzZGVjN2Y3MzciLCJwIjoiaiJ9)

## Changes
- Removed the lock on NodeViewHolder
- Made the number of parallel requests to enclave configurable, in order to avoid pool overflow error
- In case one signature fails, the signatures already created are returned

## Checks
- [ X] Project Builds
- [ X] Project passes tests and checks


qa/sc_cert_submitter_secure_enclave.py
```diff
@@ -18,7 +18,7 @@
     Start 1 MC node and 1 SC node.
     Enable Secure Enclave.
     SC node manages keys 0,1,2.
-    Secure Enclave manages keys 4,5,6
+    Secure Enclave manages keys the remaining keys (from 3 up to MAX_NUMBER_OF_KEYS)
```
i-skrypnyk, line 21: index 3 is intentinally skipped here, so it should be "from 4 up to MAX_NUMBER_OF_KEYS"

sdk/src/main/scala/io/horizen/certificatesubmitter/AbstractCertificateSubmitter.scala
```diff
@@ -269,33 +270,44 @@ abstract class AbstractCertificateSubmitter[
...
-
-    Try(Await.result(signaturesFromEnclaveFuture, timeoutDuration).flatten)
-      .getOrElse(Seq())
+    val keysFromEnclaveFuture = secureEnclaveApiClient.listPublicKeys()
+    val listOfKeys = Try(Await.result(keysFromEnclaveFuture, secureEnclaveRequestTimeout))
+      .map(managedKeys => indexedPublicKeys.filter(key_index => managedKeys.contains(key_index._1))).getOrElse(Seq())
+
+    val steps = listOfKeys.grouped(numOfParallelRequests)
+    val listOfSignatures = for (step <- steps) yield {
+      val listOfSignaturesFutures = step.map(secureEnclaveApiClient.signWithEnclave(messageToSign, _))
```
i-skrypnyk, line 299: :100: 

Let's add a comment to this method, I should have added it before, like this:
```
  /**
   * Get signatures from Secure Enclave for given public keys.
   * Ensures that no more than `numOfParallelRequests` are sent to Enclave at the same time.
   * If a batch fails to produce a response in `secureEnclaveRequestTimeout`, take what we have and continue.
   */
   ```

**Review by i-skrypnyk — APPROVED** [review]
couple tiny comment requests, other than that LGTM

File: sdk/src/main/scala/io/horizen/certificatesubmitter/AbstractCertificateSubmitter.scala [hunk 3]
```diff
@@ -269,33 +270,44 @@ abstract class AbstractCertificateSubmitter[
...
+
+      Try(Await.result(Future.sequence(listOfSignaturesFutures), secureEnclaveRequestTimeout).flatten)
+        .getOrElse {
+          val signaturesOpt = for {
+            sigFutureResult <- listOfSignaturesFutures
+            if sigFutureResult.isCompleted && sigFutureResult.value.get.isSuccess
+          } yield sigFutureResult.value.get.get
```
i-skrypnyk, line 306: if you want, you can do this to avoid if checks and .value.get.get
```
  val signaturesOpt =
    listOfSignaturesFutures
      .map(_.value) // turn Future into Option[Try[Option[CertificateSignatureInfo]]]
      .collect {
        case Some(Success(sigResult)) => sigResult
      }
```

alsala (approved):

PR #878, opened 2023-06-30
SDK-1255 - default gas limit for key rotation tx increased

ddrvar:
## Description
The default gas limit for key rotation transactions increased from 42000 to 500000 after discovering that it was not enough.
Testing shows 276988 gas is used, but after discussion with @MarcoOl94 we put 500000 just to be sure that it's enough to be executed since this type of transaction is rare.
## Jira Ticket
[SDK-1255](https://horizenlabs.atlassian.net/browse/SDK-1255?atlOrigin=eyJpIjoiYmI1M2Q4Mzg5ZmFjNDg2YmEzMjMzMmU3Yjc4MGEyZGIiLCJwIjoiaiJ9)

## Changes
The default limit increased.
sc_evm_cert_key_rotation.py updated to include 1 key rotation transaction with default gas info.

## Breaking Changes
/

## Checks
- [X] Project Builds
 
## Additional information

[SDK-1255]: https://horizenlabs.atlassian.net/browse/SDK-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/qa/SidechainTestFramework/account/httpCalls/transaction/createKeyRotationTransaction.py
+++ b/qa/SidechainTestFramework/account/httpCalls/transaction/createKeyRotationTransaction.py
@@ -4,7 +4,7 @@ import json
 def http_create_key_rotation_transaction_evm(sidechainNode, key_type, key_index, new_key,
                                              signing_key_signature, master_key_signature, new_key_signature,
                                              nonce=None, gas_limit=300000, max_fee_per_gas=900000000,
-                                             max_priority_fee_per_gas=900000000, api_key=None):
+                                             max_priority_fee_per_gas=900000000, api_key=None, include_gas_info=True):
     j = {
         "keyType": key_type,
         "keyIndex": key_index,
@@ -12,13 +12,14 @@ def http_create_key_rotation_transaction_evm(sidechainNode, key_type, key_index,
         "signingKeySignature": signing_key_signature,
         "masterKeySignature": master_key_signature,
         "newKeySignature": new_key_signature,
-        "nonce": nonce,
-        "gasInfo": {
+        "nonce": nonce
+    }
+    if include_gas_info:
+        j["gasInfo"] = {
             "gasLimit": gas_limit,
             "maxFeePerGas": max_fee_per_gas,
             "maxPriorityFeePerGas": max_priority_fee_per_gas
         }
-    }
     request = json.dumps(j)
     if api_key is not None:
         response = sidechainNode.transaction_createKeyRotationTransaction(request, api_key)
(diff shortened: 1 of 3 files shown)


qa/SidechainTestFramework/account/httpCalls/transaction/createKeyRotationTransaction.py
```diff
@@ -4,21 +4,22 @@
 def http_create_key_rotation_transaction_evm(sidechainNode, key_type, key_index, new_key,
                                              signing_key_signature, master_key_signature, new_key_signature,
                                              nonce=None, gas_limit=300000, max_fee_per_gas=900000000,
-                                             max_priority_fee_per_gas=900000000, api_key=None):
+                                             max_priority_fee_per_gas=900000000, api_key=None, include_gas_info=True):
```
MarcoOl94, line 7: Since we already have default values of gas inside the endpoint, I think we can remove the default value here in the python code and avoid to add the new **include_gas_info**

qa/sc_evm_cert_key_rotation.py
```diff
@@ -47,13 +47,17 @@
     - End the WE and verify that the certificates is added to the MC and SC.
```
MarcoOl94, line 47: I think we can avoid to add additional key rotation in the test, we just need to modify an existing call by removing the gasLimit.

MarcoOl94 (changes requested):

qa/sc_evm_cert_key_rotation.py
rushby, line 1: Any changes required to the other tests?

sc_evm_cert_key_rotation_across_epoch.py
sc_evm_cert_key_rotation_multiple_nodes.py
sc_evm_cert_key_rotation_old_circuit.py

ddrvar: No, just ran them successfully, they work as expected without any change

qa/SidechainTestFramework/account/httpCalls/transaction/createKeyRotationTransaction.py
ddrvar: Done
Added a case with gas_limit too low to check if gas limit is used in sdk

qa/sc_evm_cert_key_rotation.py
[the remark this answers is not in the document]
ddrvar: It wouldn't work if all 47 transactions were in one block
But from I see in this epoch 2 of this test, all the transactions are separated in different blocks - https://github.com/HorizenOfficial/Sidechains-SDK/blob/a6219cae35cb3e53432a70e30291da68303d037e/qa/sc_evm_cert_key_rotation.py#L520-L595

MarcoOl94 (approved):

davehorizen (approved):

PR #880, opened 2023-07-03
SDK-960 - avoid nodes sync indefinitely

davehorizen:
## Description
On the current implementation there is a problem in the way we calculate and send `SyncInfo` to the other nodes. The other node height retrieved from the  `SyncInfo` message is not correctly estimated and this lead in some cases to consider another node more updated as younger. In this situation the node consider itself more developed and stop processing other `SyncInfo` messages remaining stuck at certain height.
With the current PR we introduced a new method to correctly calculate the other node height (method `calculateOtherNodeApproxHeight`).
The Equal return statement in case the divergent suffix sequence size is greater than one was  removed because it is currently a fork situation, an existing comment on the code reported exactly that.

## Jira Ticket
details in task [SDK-960](https://horizenlabs.atlassian.net/browse/SDK-960)

## Checks
- [X] Project Builds

[SDK-960]: https://horizenlabs.atlassian.net/browse/SDK-960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
+++ b/sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
@@ -306,7 +306,6 @@ abstract class AbstractHistory[
       return Seq()
 
     var indexes: Seq[Int] = Seq()
-
     var step: Int = 1
     // Start at the top of the chain and work backwards.
     var index: Int = height
@@ -320,6 +319,28 @@ abstract class AbstractHistory[
     // Push the genesis block index.
     indexes :+ 1
   }
+
+  // calculate the other node approximate height knowing that the sequence was created using the method knownBlocksHeightToSync
+  protected def calculateOtherNodeApproxHeight(startingHeight: Int, sequenceSize: Int): Int = {
+    if (startingHeight == 1) {
+      -1 // it is not possible to calculate the other node best block approximate height in this case
+    } else {
+      var currentHeight = startingHeight
+      var currentSequencePosition = sequenceSize - 1
+      var step = 1 // step used in case the input sequence size is > 10
+      while (currentSequencePosition > 0) {
+        if (currentSequencePosition < 10) {
+          currentHeight += 1
+        } else {
+          step *= 2
+          currentHeight += step
+        }
+        currentSequencePosition -= 1
+      }
+      currentHeight
+    }
+  }
+
 
   override def syncInfo: SidechainSyncInfo = {
     // collect control points of block ids like in bitcoin (last 10, then increase step exponentially until genesis block)
@@ -357,7 +378,6 @@ abstract class AbstractHistory[
    */
   override def compare(other: SidechainSyncInfo): History.HistoryComparisonResult = {
     val dSuffix = divergentSuffix(other.knownBlockIds)
-
     dSuffix.size match {
       case 0 =>
         // log.warn(Nonsence situation..._)
@@ -368,20 +388,14 @@ abstract class AbstractHistory[
         else
           Younger
       case _ =>
-        val otherBestKnownBlockIndex = dSuffix.size - 1 - dSuffix.reverse.indexWhere(id => storage.heightOf(id).isDefined)
-        val otherBestKnownBlockHeight = storage.heightOf(dSuffix(otherBestKnownBlockIndex)).get
-        // other node height can be approximatly calculated as height of other KNOWN best block height + size of rest unknown blocks after it.
-        // why approximately? see knownBlocksHeightToSync algorithm: blocks to sync step increasing.
-        // to do: need to discuss
-        val otherBestBlockApproxHeight = otherBestKnownBlockHeight + (dSuffix.size - 1 - otherBestKnownBlockIndex)
-        if (storage.height < otherBestBlockApproxHeight)
+        val otherBestKnownBlockHeight = storage.heightOf(dSuffix.head).get
+        val otherBestBlockApproxHeight = calculateOtherNodeApproxHeight(otherBestKnownBlockHeight, dSuffix.size)
+        if(otherBestBlockApproxHeight == -1)
+          Unknown
+        else if (storage.height < otherBestBlockApproxHeight)
           Older
-        else if (storage.height == otherBestBlockApproxHeight) {
-          if(otherBestBlockApproxHeight == otherBestKnownBlockHeight)
-            Equal // UPDATE: FORK in both cases
-          else
-            Fork
-        }
+        else if (storage.height == otherBestBlockApproxHeight)
+          Fork
         else
           Younger
     }
--- a/sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
+++ b/sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
@@ -323,7 +323,12 @@ abstract class AbstractHistory[
   // calculate the other node approximate height knowing that the sequence was created using the method knownBlocksHeightToSync
   private def calculateOtherNodeApproxHeight(startingHeight: Int, sequenceSize: Int): Int = {
     if (startingHeight == 1) {
-      -1 // it is not possible to calculate the other node best block approximate height in this case
+      // It is not possible to calculate the other node best block approximate height in this case.
+      // If the only common block is the genesis one (height 1) it is not possible to correctly calculate the other node
+      // best block because the method knownBlocksHeightToSync used to create the other node sequence always push the
+      // genesis block index at the end of the list and in this way we don't know how much block separate the genesis
+      // from the second last sequence index height.
+      -1
     } else {
       var currentHeight = startingHeight
       var currentSequencePosition = sequenceSize - 1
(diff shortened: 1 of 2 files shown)


sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -368,20 +388,14 @@ abstract class AbstractHistory[
...
-        else if (storage.height == otherBestBlockApproxHeight) {
-          if(otherBestBlockApproxHeight == otherBestKnownBlockHeight)
-            Equal // UPDATE: FORK in both cases
-          else
-            Fork
-        }
+        else if (storage.height == otherBestBlockApproxHeight)
+          Fork
         else
           Younger
```
i-skrypnyk, line 400: Do you think it is safe to set the status to Younger as default here?
This case means that that other node has fewer blocks, but there is also a Fork, because a number of blocks from the other node history was not found in local history.
So it is both Fork and Younger

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -321,6 +320,28 @@ abstract class AbstractHistory[
     indexes :+ 1
   }
 
+  // calculate the other node approximate height knowing that the sequence was created using the method knownBlocksHeightToSync
+  protected def calculateOtherNodeApproxHeight(startingHeight: Int, sequenceSize: Int): Int = {
```
i-skrypnyk, line 324: Can you please add a test(unit will do) that will fail without the fix and pass with the fix

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -341,6 +341,25 @@ abstract class AbstractHistory[
...
+    while(currentSequencePosition < (dSuffixSize-1)) {
+      if(currentSequencePosition < 10) {
+        heightAtIndex -= 1
+      } else {
+        step *= 2
+        heightAtIndex -= step
+      }
+      currentSequencePosition += 1
+    }
+    if(heightAtIndex < storage.height)
```
paologalligit: This can be simplified by only returning `heightAtIndex < storage.height`

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -321,13 +320,46 @@ abstract class AbstractHistory[
...
+        currentSequencePosition -= 1
+      }
+      currentHeight
+    }
+  }
+
+  // check the presence of a block on the current history at the second divergent suffix sequence height
+  // it is used to check the presence of forks
+  private def checkForkAtPreviousIndexBlock(otherBestKnownBlockHeight: Int, dSuffixSize: Int): Boolean = {
+    var heightAtPreviousIndex = 0
```
paologalligit, line 347: Here we can directly assign
```
val heightAtPreviousIndex = if (dSuffixSize > 10) {
      otherBestKnownBlockHeight + math.pow(2, dSuffixSize-10).toInt
    } else {
      otherBestKnownBlockHeight + dSuffixSize
    } 
```

paologalligit (commented):
I'd like to have a set of tests for the `checkForkAtPreviousIndexBlock` method corner cases

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
davehorizen: Good point, after further analysis I proceed to refactor the method logic.
The current logic in case the divergent suffix sequence size is greater than 1 is the following:
knowing our block height, the best known block height and the calculated best block related to the other node history, we do the following checks:
- if the other node best block height is greater than the current height:
	- and the height of the second index of the divergent suffix sequence is less then the current height we have a `Fork` (this because it means that the block id at a height less than our best block didn't found a match)
	- else the other node history is `Older`
the default condition for the other cases is `Fork`.

No need to manage the `Younger` case because in case the other node is really younger (with no forks) the divergent suffix sequence size will be 1 and we will be outside these cases

The `Unknown` condition is satisfied only if we are not able to calculate the other node best block height, in this case we don't have enough data to return a secure comparison result.

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
davehorizen: Unit test added (`SidechainHistoryTest.synchronizationTestWithHighBlockNumber`) to simulate the task example. The number of blocks involved is less than the one of the task to keep the test runtime low but enough to manage divergent suffix sequence that involved height difference with exponential step.

davehorizen:
> I'd like to have a set of tests for the `checkForkAtPreviousIndexBlock` method corner cases

The unit test `SidechainHistoryTest.checkForkAtPreviousIndexBlock` is already covering this case [here](https://github.com/HorizenOfficial/Sidechains-SDK/blob/db/SDK-960/sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala#L634) and [here](https://github.com/HorizenOfficial/Sidechains-SDK/blob/db/SDK-960/sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala#L666) where we are comparing our history against a longer one with a fork at a certain height (a `Fork` case instead of an `Older` one)

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -321,13 +320,46 @@ abstract class AbstractHistory[
...
+    if (dSuffixSize > 10) {
+      heightAtPreviousIndex = otherBestKnownBlockHeight + math.pow(2, dSuffixSize-10).toInt
+    } else {
+      heightAtPreviousIndex = otherBestKnownBlockHeight + dSuffixSize
+    }
+    if (heightAtPreviousIndex < storage.height)
```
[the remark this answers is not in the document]
davehorizen: agree, more clear! Done

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
davehorizen: done

davehorizen:
Unit test pushed to cover a corner case of the` AbstractHistory.comparison` method that will result in an Older status even if there is a Fork between the two histories.
Details in the[ unit test comment](https://github.com/HorizenOfficial/Sidechains-SDK/blob/db/SDK-960/sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala#L615).

paologalligit (approved):

sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala
```diff
@@ -534,29 +534,173 @@ class SidechainHistoryTest extends JUnitSuite
...
+    // -----------------------------------------------------------------------------------------------------------------
+    // Test 2:
+    // update history2 with 10 more blocks (introduce a fork).
+    // Here we can test a corner case in which the comparison between two histories will result in a Older state even if there is Fork
+    // The node1 (at height 300) will send the node2 a list with the modifiers ID at the following height:
+    // 300 - 299 - 298 - 297 - 296 - 295 - 294 - 293 - 292 - 291 - 289 - 285 - 277 - 261 - 229 - 165 - 37 - 1
+    // The divergent suffix list will be the modifiers at the following height (size 17):
+    // 37 - 165 - 229 - 261 - 277 - 285 - 289 - 291 - 292 - 293 - 294 - 295 - 296 - 297 - 298 - 299 - 300
+    // The internal logic will calculate the second divergent suffix height.
+    // The result (165) is less than the current height (110) and the final result will be an Older status
```
MarcoOl94, line 623:
```suggestion
    // The result (165) is greater than the current height (160) and the final result will be an Older status
```

MarcoOl94 (changes requested):

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -321,13 +320,44 @@ abstract class AbstractHistory[
...
+          step *= 2
+          currentHeight += step
+        }
+        currentSequencePosition -= 1
+      }
+      currentHeight
+    }
+  }
+
+  // check the presence of a block on the current history at the second divergent suffix sequence height
```
paolocappelletti, line 344: let's change the comment and the name of the method to be more clear
checkForkAtSecondToLastDivergentSuffixIndex
or something similar :)

paolocappelletti (approved):

sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala
davehorizen: fixed

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
davehorizen: method name update, the one proposed seems fine and it was used!

sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala
```diff
@@ -618,9 +618,9 @@ class SidechainHistoryTest extends JUnitSuite
     // The node1 (at height 300) will send the node2 a list with the modifiers ID at the following height:
     // 300 - 299 - 298 - 297 - 296 - 295 - 294 - 293 - 292 - 291 - 289 - 285 - 277 - 261 - 229 - 165 - 37 - 1
     // The divergent suffix list will be the modifiers at the following height (size 17):
-    // 37 - 165 - 229 - 261 - 277 - 285 - 289 - 291 - 292 - 293 - 294 - 295 - 296 - 297 - 298 - 299 - 300
+    // 1 - 37 - 165 - 229 - 261 - 277 - 285 - 289 - 291 - 292 - 293 - 294 - 295 - 296 - 297 - 298 - 299 - 300
```
MarcoOl94, line 621: Are you sure of this 1?

MarcoOl94 (approved):

sdk/src/test/scala/io/horizen/utxo/integration/SidechainHistoryTest.scala
davehorizen: removed

PR #905, opened 2023-08-04
Fixed issue with different fee payments.

VolodymyrBabaiev:
## Description
Fixed issue with different fee payments.

## Jira Ticket
SDK-1166 https://horizenlabs.atlassian.net/browse/SDK-1166

## Changes
Updated withdrawal epoch calculation for forger.

## Checks
- [x] Project Builds
- [ ] Project passes tests and checks

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/forger/AccountForgeMessageBuilder.scala
+++ b/sdk/src/main/scala/io/horizen/account/forger/AccountForgeMessageBuilder.scala
@@ -264,7 +264,7 @@ class AccountForgeMessageBuilder(
               )
               require(ommers.isEmpty, "No Ommers allowed for the last block of the withdrawal epoch.")
 
-              val withdrawalEpochNumber: Int = dummyView.getWithdrawalEpochInfo.epoch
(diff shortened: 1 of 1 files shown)


sdk/src/main/scala/io/horizen/account/forger/AccountForgeMessageBuilder.scala
```diff
@@ -264,7 +264,7 @@ class AccountForgeMessageBuilder(
               )
               require(ommers.isEmpty, "No Ommers allowed for the last block of the withdrawal epoch.")
 
-              val withdrawalEpochNumber: Int = dummyView.getWithdrawalEpochInfo.epoch
```
paologalligit, line 267: Why wasn't taking the withdrawal epoch from the view correct?

VolodymyrBabaiev: In case first MC block of the next epoch haven't been added to Sidechain, view consider that it's previous epoch now. When we include all MC blocks from the epoch it causes that fee will be calculated for previous epoch,

i-skrypnyk (approved):

PR #922, opened 2023-09-18
SDK-1430: add test to verify sidechain nodes can still create certificates when mc nodes are offline

paologalligit:
## Description
Add test to verify sidechain nodes can still create certificates when the mainchain nodes their connected to temporarily are offline

## Jira Ticket
[SDK-1430](https://horizenlabs.atlassian.net/browse/SDK-1430)

## Changes
Add test to verify sidechain nodes can still create certificates when the mainchain nodes their connected to temporarily are offline

## Breaking Changes
No breaking changes

## Checks
- [x] Project Builds

[SDK-1430]: https://horizenlabs.atlassian.net/browse/SDK-1430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/qa/run_sc_tests.sh
+++ b/qa/run_sc_tests.sh
@@ -126,6 +126,7 @@ testScriptsEvm=(
     'sc_evm_seedernode.py'
     'sc_evm_consensus_parameters_fork.py'
     'sc_evm_active_slot_coefficient.py'
+    'sc_withdrawal_certificate_after_mainchain_nodes_were_disconnected.py'
 );
 
 testScriptsUtxo=(
--- a/qa/sc_withdrawal_certificate_after_mainchain_nodes_were_disconnected.py
+++ b/qa/sc_withdrawal_certificate_after_mainchain_nodes_were_disconnected.py
@@ -1,0 +1,254 @@
+import logging
+import time
+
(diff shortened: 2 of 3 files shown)


rushby (commented):
I think we should consider how to simulate the MC node deprecation, and the use of old and new versions of ZEND.  It's not something we currently cover in any of the tests

MarcoOl94 (approved):

paologalligit:
> I think we should consider how to simulate the MC node deprecation, and the use of old and new versions of ZEND. It's not something we currently cover in any of the tests

That's interesting, but this can be something we can have in a different test case. We've found an issue where the previous certificate is deleted from the storage and cannot be retrieved anymore, so the zend deprecation should not be a problem. A brief description [here](https://horizenlabs.atlassian.net/browse/SDK-1430)

paologalligit:
Converting current PR into a draft since the fix has not been decided yet

paolocappelletti:
remember to add the new test to run_sc_tests.sh whenever we insert a new one
(I've added it)

PR #925, opened 2023-09-22
SDK-1503 : Jacoco integration

ddrvar:
## Description
Jacoco integration and code coverage report generation

## Jira Ticket
[SDK-1503](https://horizenlabs.atlassian.net/browse/SDK-1503?atlOrigin=eyJpIjoiZDI2ZGQyY2VmNDE3NGVhYWFiZjBjOWIzMjFmNjkxNTgiLCJwIjoiaiJ9)

## Changes
Jacoco tool integrated with SDK. New `jacoco-maven-plugin` in SDK's pom.xml. In order to get the report, it is necessary to run mvn clean install with the test phase included. After that, you can run integration tests which will append their analysis to the previously created report. There is a script `generate_report.sh` which runs unit and integration tests and then generates the report in the end.
For report creation it is necessary to have `org.jacoco.cli-0.8.9-nodeps.jar` which you can download [here](https://repo1.maven.org/maven2/org/jacoco/org.jacoco.cli/0.8.9/)
Lastly, there is a flag `is_jacoco_included` in `scutil.py` for the inclusion of jacoco agent when tests are running.
## Breaking Changes
/

## Checks
- [X] Project Builds
 
## Additional information

[SDK-1503]: https://horizenlabs.atlassian.net/browse/SDK-1503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rushby (changes requested):
Amend generate_report.sh to pass in a jacoco flag that sets the is_jacoco_included to true.

Run and confirm it works for one test, and for many (i'd just comment out all the tests apart from maybe 3 and make sure they are all represented in the report).

coverage-reports/generate_report.sh
```diff
@@ -0,0 +1,32 @@
...
+# Step 3: Execute run_sc_tests.sh
+echo "Executing script for integration tests..."
+cd ../qa || exit 1
+./run_sc_tests.sh -jacoco
+
+# Step 4: Generate the JaCoCo code coverage report
+echo "Generating JaCoCo code coverage report..."
+
+JACOCO_JAR_PATH="$HOME/.m2/repository/org/jacoco/org.jacoco.cli/0.8.9/org.jacoco.cli-0.8.9-nodeps.jar"
+EXEC_PATH="$SIDECHAIN_SDK/coverage-reports/sidechains-sdk-${SNAPSHOT_VERSION_TAG}/sidechains-sdk-${SNAPSHOT_VERSION_TAG}-jacoco-report.exec"
```
paologalligit, line 23: I would add a check to make sure the `SIDECHAIN_SDK` env var is not empty at the beginning of the script

coverage-reports/generate_report.sh
```diff
@@ -0,0 +1,32 @@
...
+cd ../sdk || exit 1
+
+# Step 2: Run 'mvn clean install' in the project directory
+echo "Running 'mvn clean install' in the project directory..."
+mvn clean install
+
+# Step 3: Execute run_sc_tests.sh
+echo "Executing script for integration tests..."
+cd ../qa || exit 1
+./run_sc_tests.sh -jacoco
```
paologalligit, line 17: To run these tests we need to have also BITCOIND and BITCOINCLI env var exported. Let's add a check here and if any of those are missing we can echo a warning message and skip them

paologalligit (changes requested):
Looks all good, only a couple of points to fix in the script. I'd also add some line of instructions on how to generate a jacoco report (`using the generate_report.sh` script), the env var needed, ecc...

coverage-reports/generate_report.sh
ddrvar: done

coverage-reports/generate_report.sh
ddrvar: done

paologalligit (approved):
LGTM, just added a comment about changing a condition logic

paolocappelletti (approved):

PR #926, opened 2023-09-22
Is/modifiers cache memory leak fix

i-skrypnyk:
## Description
the overflow of the modifiers cache is not something we introduced in this release, the logic there allows cache to grow while there's a new block to apply to the current history in the cache

only when the cache does not contain any valid block, it will clear itself

and I believe that on previous version, due to other synchronization problems, we did not encounter this one

but now the sync process is working better and faster, so we hit this bottleneck

## Jira Ticket
https://horizenlabs.atlassian.net/browse/SDK-1513

### DIFF
--- a/sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
+++ b/sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
@@ -175,7 +175,10 @@ abstract class AbstractSidechainNodeViewHolder[
         val bestBlockTimestampPlus24H = history().bestBlock.timestamp + TimeUnit.HOURS.toSeconds(24)
         val (modsToApply, modsToSkip) = mods.partition(m => m.timestamp <= bestBlockTimestampPlus24H)
         modsToApply.foreach(m => modifiersCache.put(m.id, m))
-        if (modsToSkip.nonEmpty) context.system.eventStream.publish(ModifiersProcessingResult(Seq(), modsToSkip))
+        if (modsToSkip.nonEmpty) {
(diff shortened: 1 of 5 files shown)


alsala (approved):
LGTM

sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
```diff
@@ -163,12 +164,21 @@ abstract class AbstractSidechainNodeViewHolder[
...
    */
   override def processRemoteModifiers: Receive = {
     case sparkz.core.NodeViewHolder.ReceivableMessages.ModifiersFromRemote(mods: Seq[PMOD]) =>
-      mods.foreach(m => modifiersCache.put(m.id, m))
+
+      if (modifiersCache.size + mods.size > sparksSettings.network.maxModifiersCacheSize / 2) {
+        val bestBlockTimestampPlus24H = history().bestBlock.timestamp + TimeUnit.HOURS.toSeconds(24)
+        val (modsToApply, modsToSkip) = mods.partition(m => m.timestamp <= bestBlockTimestampPlus24H)
+        modsToApply.foreach(m => modifiersCache.put(m.id, m))
+        if (modsToSkip.nonEmpty) context.system.eventStream.publish(ModifiersProcessingResult(Seq(), modsToSkip))
```
paolocappelletti, line 178: please add a comment on why we are publishing this event as discussed

paolocappelletti (changes requested):

paolocappelletti (approved):

PR #928, opened 2023-09-26
SDK-1527: clean old top certificate only if that certificate was alre…

paologalligit:
…ady used as previous cert

## Description
We do not delete the top certificate if it has not yet been used as a previous certificate.

## Jira Ticket
[1527](https://horizenlabs.atlassian.net/browse/SDK-1527)

## Changes
AccountStateMetadataStorageView.scala: introduced a new method that returns an Option with any certificate to be erased

## Breaking Changes
No breaking changes

## Checks
- [x] Project Builds

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/storage/AccountStateMetadataStorageView.scala
+++ b/sdk/src/main/scala/io/horizen/account/storage/AccountStateMetadataStorageView.scala
@@ -345,8 +345,10 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
           case _ => false
         }
         if (isWithdrawalEpochSwitched) {
-          val certEpochNumberToRemove: Int = epochInfo.epoch - 4
-          removeList.add(getTopQualityCertificateKey(certEpochNumberToRemove))
+          getOldTopCertificatesToBeRemoved(epochInfo) match {
+            case Some(cert) => removeList.add(cert)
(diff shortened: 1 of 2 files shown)


sdk/src/main/scala/io/horizen/account/storage/AccountStateMetadataStorageView.scala
```diff
@@ -388,6 +400,7 @@ class AccountStateMetadataStorageView(storage: Storage) extends AccountStateMeta
   }
 
   private[horizen] def getTopQualityCertificateKey(referencedWithdrawalEpoch: Int): ByteArrayWrapper = {
+    val x = calculateKey(Bytes.concat("topQualityCertificate".getBytes(StandardCharsets.UTF_8), Ints.toByteArray(referencedWithdrawalEpoch)))
```
MarcoOl94, line 403: I think we can remove this

MarcoOl94 (changes requested):

sdk/src/main/scala/io/horizen/account/storage/AccountStateMetadataStorageView.scala
paologalligit: Oooooops

MarcoOl94 (approved):

paolocappelletti (approved):

PR #929, opened 2023-09-27
Storage versioning fix - do not store unused versions

i-skrypnyk:


alsala (commented):
We should update to 0.8.1 also the SNAPSHOT_VERSION_TAG constant in qa/SidechainTestFramework/scutil.py

MarcoOl94 (approved):

paolocappelletti (approved):

PR #941, opened 2023-10-17
SDK-953: wrapping log.debug statements in anonymous functions to lazy…

paologalligit:
… evaluate them only if debug log level is on

## Description
Since log.debug statements are executed only when the debug level is set in the config, we don't want to waste any computational resources to evaluate complex functions if we're not going to log anything. Using the anonymous functions as arguments, we can move a code block inside it and it will only be evaluated if the debug level is on.

## Jira Ticket
[SDK-953](https://horizenlabs.atlassian.net/browse/SDK-953)

## Changes
log.debug statements

## Breaking Changes
/

## Checks
- [x] Project Builds
 
## Additional information

[SDK-953]: https://horizenlabs.atlassian.net/browse/SDK-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
+++ b/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
@@ -77,11 +77,11 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
 
         log.debug(s"history bestBlockId = $historyVersion, stateVersion = $checkedStateVersion")
 
-        log.debug("{}", (() => {
+        log.whenDebugEnabled {
           val height_h = restoredHistory.blockInfoById(restoredHistory.bestBlockId).height
           val height_s = restoredHistory.blockInfoById(checkedStateVersion).height
           s"history height = $height_h, state height = $height_s"
-        }).apply)
+        }
 
         if (historyVersion == checkedStateVersion) {
           log.info("state and history storages are consistent")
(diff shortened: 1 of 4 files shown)


i-skrypnyk:
I think this would not work, and we don't need this. Let me try to explain my thoughts:

1. lambda with .apply gets evaluated immediately, you can test it in the scala worksheet
image

2. we are using scala logging, which already have some complicated solution for lazy evaluation, that works automagically, see here: https://github.com/lightbend-labs/scala-logging
They say that we should not worry about checking the log level or lazy evaluation, because they cover it by using macros for evaluating the calls.

3. But our loggers extend StrictLogging, so the .debug calls are evaluated right away. We should use .whenDebugEnabled, like here:

```
        log.whenDebugEnabled {
          val height_h = restoredHistory.blockInfoById(restoredHistory.bestBlockId).height
          val height_s = restoredHistory.blockInfoById(checkedStateVersion).height
          s"history height = $height_h, state height = $height_s"
        }
```

paologalligit:
> I think this would not work, we should use .whenDebugEnabled. Let me try to explain my thoughts:
> 
> 1. lambda with .apply gets evaluated immediately, you can test it in the scala worksheet
>    image
> 2. we are using scala logging, which already have some complicated solution for lazy evaluation, that works automagically, see here: https://github.com/lightbend-labs/scala-logging
>    They say that we should not worry about checking the log level or lazy evaluation, because they cover it by using macros for evaluating the calls.
> 3. But our loggers extend StrictLogging, so the .debug calls are evaluated right away. We should use .whenDebugEnabled, like here:
> 
> ```
>         log.whenDebugEnabled {
>           val height_h = restoredHistory.blockInfoById(restoredHistory.bestBlockId).height
>           val height_s = restoredHistory.blockInfoById(checkedStateVersion).height
>           s"history height = $height_h, state height = $height_s"
>         }
> ```

I was confused because I didn't find the scala-logging as a direct dependency in the sdk; thanks for the comment

i-skrypnyk (approved):

PR #945, opened 2023-10-26
Pc/eth gas price rework

paolocappelletti:
## Description
reworked suggestTipCap to make it  similar to geth (used in eth_gasPrice)

## Jira Ticket
[SDK-1587](https://horizenlabs.atlassian.net/browse/SDK-1587)

[SDK-1587]: https://horizenlabs.atlassian.net/browse/SDK-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
+++ b/sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
@@ -6,14 +6,15 @@ import io.horizen.account.state.AccountStateView
 import io.horizen.account.transaction.EthereumTransaction
 import io.horizen.account.utils.FeeUtils.INITIAL_BASE_FEE
 import io.horizen.utils.BytesUtils
-import sparkz.util.ModifierId
+import sparkz.util.{ModifierId, SparkzLogging}
 
 import java.math.BigInteger
 
-object Backend {
+object Backend extends SparkzLogging {
 //  private type NV = CurrentView[AccountHistory, AccountState, AccountWallet, AccountMemoryPool]
 
   private val MAX_GAS_PRICE = INITIAL_BASE_FEE.multiply(BigInteger.valueOf(500))
+  private val SUGGEST_TIP_TX_LIMIT = 3 //number of tx to consider per block for the tip estimation algorithm
 
   private var tipCache : Option[TipCache] = Option.empty
 
@@ -68,7 +69,7 @@ object Backend {
     )
 
     // If the latest gasprice is still valid, return it.
-    if (lastHead.isDefined && lastHead == headHash){
+    if (lastHead.isDefined && lastHead.get == headHash){
       return lastPrice
     }
 
@@ -77,7 +78,7 @@ object Backend {
     var prices: Seq[Option[Seq[BigInteger]]] =  Seq() //in go-ethereum this is called result
     var results: Seq[BigInteger] = Seq()
     while (sent < blockCount && number > 0){
-      prices = prices :+ getBlockPrices(history, number, ignorePrice, 3)
+      prices = prices :+ getBlockPrices(history, number, ignorePrice, SUGGEST_TIP_TX_LIMIT)
       sent += 1
       exp += 1
       number -= 1
@@ -87,11 +88,12 @@ object Backend {
       prices = prices.drop(1)
       if (pricesOfABlock.isEmpty) {
         //if we are here we had some errors collecting the blocks
+        log.warn("Error retrieving blocks in history while suggesting tip cap, returning the cached one")
         return lastPrice
       }
       exp -= 1
       var res : Seq[BigInteger] = pricesOfABlock.get
-      if (res.length == 0) {
+      if (res.isEmpty) {
         // Nothing returned. There are two special cases here:
         // - The block is empty
         // - All the transactions included are sent by the miner itself.
@@ -104,7 +106,6 @@ object Backend {
       // is 2*checkBlocks.
       if (res.length == 1 && results.length + 1 + exp < blockCount * 2 && number > 0) {
         prices = prices :+ getBlockPrices(history, number, ignorePrice, 3)
-        sent += 1
         exp += 1
         number -= 1
       }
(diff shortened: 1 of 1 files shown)


sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
+    while (exp > 0){
+      val pricesOfABlock : Option[Seq[BigInteger]] = prices.head
+      prices = prices.drop(1)
+      if (pricesOfABlock.isEmpty) {
+        //if we are here we had some errors collecting the blocks
+        return lastPrice
+      }
+      exp -= 1
+      var res : Seq[BigInteger] = pricesOfABlock.get
+      if (res.length == 0) {
```
saratnt, line 94: Consider replacing  `(res.length == 0)` with `res.isEmpty `because it is more efficient

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
-      .sorted
-      .lift((prices.length - 1) * percentile / 100)
-      .getOrElse(BigInteger.ZERO)
-      .min(maxPrice)
+      // Besides, in order to collect enough data for sampling, if nothing
+      // meaningful returned, try to query more blocks. But the maximum
+      // is 2*checkBlocks.
+      if (res.length == 1 && results.length + 1 + exp < blockCount * 2 && number > 0) {
+        prices = prices :+ getBlockPrices(history, number, ignorePrice, 3)
+        sent += 1
```
saratnt, line 107: Why we still need `sent` here?

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
+        exp += 1
+        number -= 1
+      }
+      results = results ++ res
+    }
+
+    var resultPrice = lastPrice
+    if (results.length > 0) {
```
saratnt, line 115: Consider replace with `results.isNotEmpty`

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
-          .flatMap(history.getStorageBlockById)
-          .get
-        val blockPrices = getBlockPrices(block, ignorePrice, limit)
-        collected += blockPrices.length
-        if (i >= blockCount) moreBlocksNeeded = true
-        blockPrices
+    var number = history.getCurrentHeight
+    val headHash = history.bestBlockId
+    val (lastHead : Option[ModifierId], lastPrice: BigInteger) = tipCache.fold(
+      (Option.empty[ModifierId], BigInteger.ZERO)
```
saratnt, line 65: In geth the default lastPrice value is 1 Gwei, configurable. I think we need to do the same to be sure of having the same results.

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
+    )(cache =>
+      (Some(cache.blockHash), cache.value)
+    )
+
+    // If the latest gasprice is still valid, return it.
+    if (lastHead.isDefined && lastHead == headHash){
```
saratnt, line 71: If I have understood correctly, `lastHead` is an Option[ModifierId] and `headHash` is a ModifierId, shouldn't the comparison always fail?

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
+      }
+      results = results ++ res
+    }
+
+    var resultPrice = lastPrice
+    if (results.length > 0) {
+      resultPrice =
+        results
+          .sorted
+          .lift((prices.length - 1) * percentile / 100)
```
saratnt, line 119: Shouldn't be `results.length` instead of `prices.length`?

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +59,100 @@ object Backend {
...
+      sent += 1
+      exp += 1
+      number -= 1
+    }
+    while (exp > 0){
+      val pricesOfABlock : Option[Seq[BigInteger]] = prices.head
+      prices = prices.drop(1)
+      if (pricesOfABlock.isEmpty) {
+        //if we are here we had some errors collecting the blocks
+        return lastPrice
```
saratnt, line 90: Maybe we could add some logging here, it this happens it's really worrisome

saratnt (changes requested):

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
paolocappelletti: true! thanks

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
paolocappelletti: we will keep 0 since for us the forgers don't have a min fee for collecting tx

saratnt (approved):

sdk/src/main/scala/io/horizen/account/api/rpc/service/Backend.scala
```diff
@@ -57,49 +60,100 @@ object Backend {
...
-    prices
-      .sorted
-      .lift((prices.length - 1) * percentile / 100)
-      .getOrElse(BigInteger.ZERO)
-      .min(maxPrice)
+      // Besides, in order to collect enough data for sampling, if nothing
+      // meaningful returned, try to query more blocks. But the maximum
+      // is 2*checkBlocks.
+      if (res.length == 1 && results.length + 1 + exp < blockCount * 2 && number > 0) {
+        prices = prices :+ getBlockPrices(history, number, ignorePrice, 3)
```
paologalligit, line 108: This 3 should still refer to `SUGGEST_TIP_TX_LIMIT`, right? If yes, we can replace it

paolocappelletti: yes correct, thanks!

PR #958, opened 2023-11-20
Is/node stuck at fork fix

i-skrypnyk:
## Description
If node was not updated to the new version during the hard fork, it will set the status of the block at fork point as invalid as save the status in the persistent storage. Then, after restart, node will take the status from storage and never try to re-evaluate it with the new code. 

This PR adds a config value that resets modifiers status that are retrieved from storage, so that the node is able to try the fork again, and if it succeeds or fail, the new status will be updated both in the in-memory activeChain, and in the storage

## Jira Ticket
https://horizenlabs.atlassian.net/browse/SDK-1613

## Checks
- [x] Project Builds


paolocappelletti (commented):
can you add a description of how (and when) to use the new feature here:
https://github.com/HorizenOfficial/Sidechains-SDK/blob/is/node_stuck_at_fork_fix/doc/release/0.9.0.md#notes-about-newupdated-features

doc/release/0.9.0.md
```diff
@@ -51,7 +51,15 @@ import "./ForgerStakes.sol";
 
 ---
 ## Bug Fixes
-
+- added option to reset modifiers status when comparing forks. This is useful if the node was not updated before the fork
```
saratnt, line 54: Maybe it is worth mentioning that this flag should be used just as long as the node has reached the tip of the chain, then the node should be restarted with resetModifiersStatus = false

sdk/src/main/scala/io/horizen/history/AbstractHistory.scala
```diff
@@ -167,7 +167,7 @@ abstract class AbstractHistory[
     if(newChainSuffix.isEmpty && currentChainSuffix.isEmpty)
       throw new IllegalArgumentException("Cannot retrieve fork changes. Fork length is more than params.maxHistoryRewritingLength")
 
-    val newChainSuffixValidity: Boolean = !newChainSuffix.tail.map(isSemanticallyValid)
+    val newChainSuffixValidity: Boolean = params.resetModifiersStatus || !newChainSuffix.tail.map(isSemanticallyValid)
```
saratnt, line 170: Maybe it is a good idea to put a comment here, just as a reminder of why there is this flag check here

saratnt (approved):

paolocappelletti (approved):

PR #965, opened 2023-12-11
Signature v field serialization

i-skrypnyk:
## Description
Signature v field json representation was wrong - the value should be 0 or 1

## Jira Ticket
https://horizenlabs.atlassian.net/browse/SDK-1627

## Changes
Added a new JsonSerializer specifically for signature.v field - it checks the type of the transaction it is part of, and if the type is legacy - v field is 27 or 28. If the type if EIP1559 - v field is 0 or 1

## Checks
- [x] Project Builds

### DIFF
--- a/sdk/src/main/java/io/horizen/json/serializer/SignatureVFieldSerializer.java
+++ b/sdk/src/main/java/io/horizen/json/serializer/SignatureVFieldSerializer.java
@@ -1,0 +1,22 @@
+package io.horizen.json.serializer;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import io.horizen.account.transaction.EthereumTransaction;
+import org.web3j.crypto.Sign;
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+
+public class SignatureVFieldSerializer extends JsonSerializer<BigInteger> {
+    @Override
+    public void serialize(BigInteger val, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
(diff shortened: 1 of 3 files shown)


sdk/src/main/java/io/horizen/json/serializer/SignatureVFieldSerializer.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.IOException;
+import java.math.BigInteger;
+
+
+public class SignatureVFieldSerializer extends JsonSerializer<BigInteger> {
+    @Override
+    public void serialize(BigInteger val, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException {
+        try {
+            Object pojo = jsonGenerator.getOutputContext().getParent().getCurrentValue();
+            if (pojo instanceof EthereumTransaction && ((EthereumTransaction) pojo).isLegacy())
```
alsala, line 18: Maybe we should test `isEIP1559()` eth tx attribute, after all we have 3 alternatives for the tx type (0, 1, 2), even if 1 is currently not used in EON but it is supported by geth and in future maybe we will also use that type

alsala (changes requested):

sdk/src/main/java/io/horizen/json/serializer/SignatureVFieldSerializer.java
paolocappelletti: agree! 
also I suggest to copy&paste here the same comment in the test
// isEIP1559 txs are defined to use 0 and 1 ..... 
just to remember in the future why we did this

paolocappelletti (changes requested):

paolocappelletti:
please update changelog.md also

paolocappelletti (approved):

PR #966, opened 2023-12-11
Limit number of block to traverse for eth_getLogs

i-skrypnyk:
eth_getLogs with earliest and latest block tags timed out because it tried to inspect all the blocks. Failing EthService actor also failed all the other concurrent calls

## Jira Ticket
https://horizenlabs.atlassian.net/browse/SDK-1643

## Changes
Added a configurable limit of blocks that can be inspected. Limit set to 10000. Change error message to represent new condition.

## Checks
- [x] Project Builds


paolocappelletti (changes requested):
please update CHANGELOG.md also

saratnt (approved):
LGTM. If I'm not wrong, it should be also modified "Notes for Infra" document.

paolocappelletti (approved):

PR #973, opened 2023-12-21
St/shanghai gas sdk

saratnt:
## Description
Support for new intrinsic gas calculation and contract initcode max size limit.

## Jira Ticket
[SDK-1629](https://horizenlabs.atlassian.net/browse/SDK-1629?atlOrigin=eyJpIjoiMjE0YWQ0ZjMwYTJkNGQ4M2JjMDdjMzFmOTcyOTQ2ZWUiLCJwIjoiaiJ9)

## Checks
- [X] Project Builds
- [X] Updated documentation accordingly
- [X] Breaking changes have been correctly tagged and notified
 
## Additional information
Requires libevm PR:
- https://github.com/HorizenOfficial/libevm/pull/27

[SDK-1629]: https://horizenlabs.atlassian.net/browse/SDK-1629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
+++ b/sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
@@ -1,6 +1,7 @@
 package io.horizen.account.transaction;
 
 import com.fasterxml.jackson.annotation.*;
+import io.horizen.account.fork.Version1_3_0Fork;
 import io.horizen.account.proof.SignatureSecp256k1;
 import io.horizen.account.proposition.AddressProposition;
 import io.horizen.account.state.GasUtil;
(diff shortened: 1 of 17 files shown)


sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
```diff
@@ -216,8 +218,14 @@ public TransactionSerializer serializer() {
         return EthereumTransactionSerializer.getSerializer();
     }
 
+    @Deprecated
     @Override
     public void semanticValidity() throws TransactionSemanticValidityException {
+        semanticValidity(0);
```
paolocappelletti, line 224: possible problem: the method semanticValidity() (without the consensus epoch parameter) is still used in SidechainBlockBase, which is called by the validators before appending the block to the history.
With this change we could have a block that pass the history validation but then is marked as invalid in the state apply (where is called semantiValidiry(consensusEpoch))

paolocappelletti (changes requested):

sdk/src/main/java/io/horizen/account/transaction/EthereumTransaction.java
saratnt: You're right. I have moved the transaction validation inside a method in SidechainBlockBase that is overridden by AccountBlock, so the new version of semanticValidity can be called. The old version of semanticValidity now throws an exception , so it cannot be called anymore in account model.

i-skrypnyk (approved):

paolocappelletti (approved):

PR #979, opened 2024-01-11
As/disable forging while syncing

alsala:
Too much forging activity logged if blockhistory has not reached current consensus epoch

JIRA task: https://horizenlabs.atlassian.net/browse/SDK-1638?atlOrigin=eyJpIjoiNTQ4OGEyYjI3OGQwNDExZTgyMmVmNjU4NDU0YTY1ZmQiLCJwIjoiaiJ9


sdk/src/main/scala/io/horizen/forge/AbstractForgeMessageBuilder.scala
```diff
@@ -154,6 +154,10 @@ abstract class AbstractForgeMessageBuilder[
     val currentTipBlockEpochAndSlot: ConsensusEpochAndSlot = TimeToEpochUtils.timestampToEpochAndSlot(params.sidechainGenesisBlockTimestamp, currentTipBlockTimestamp)
     val nextBlockEpochAndSlot: ConsensusEpochAndSlot = ConsensusEpochAndSlot(nextEpochNumber, nextSlotNumber)
 
+    if ((nextEpochNumber - currentTipBlockEpochAndSlot.epochNumber) > 1) {
```
saratnt, line 157: I'm probably just being paranoid but if for any reason for 2 epochs no block is forged, the forging will be stopped forever. I know that it shouldn't be possible but what if there is an attack? Or what if at the beginning of a chain life for some reason we don't have blocks for 2 epochs? 
So, just to err on the safe side, what about creating a configuration param that can disable this check?

sdk/src/main/scala/io/horizen/forge/AbstractForger.scala
```diff
@@ -136,10 +136,36 @@ abstract class AbstractForger[
 
 
   protected def tryToCreateBlockNow(): Unit = {
+    // we are in the timer func, triggered periodically for automatic forging attempt
     val currentTime: Long = timeProvider.time() / 1000
+/* TODO remove after code review
+This could be enabled if we want to skip the forging step entirely while we are syncing, thus saving many cpu cycles. Otherwise the slot is skipped while trying to forge.
+The drawback of this approach is that we call an async func (with a future obj) for getting current tip info every timer expiration
```
saratnt, line 143: I don't think it is worthy to have every time an additional async call for a check that is useful only for a short period of the lifetime of a forger. Maybe it can be done inside tryToForgeNextBlock before calling getBranchPointInfo, so we can still save some cpu cycles without the overhead of the additional async call.

saratnt (approved):
LGTM, I have just a couple of comments.

sdk/src/main/scala/io/horizen/forge/AbstractForgeMessageBuilder.scala
alsala: The suggested approach will be analyzed with the team in a separate context, and in case a new issue will be open for working at it.

saratnt: ok for me

sdk/src/main/scala/io/horizen/forge/AbstractForger.scala
paolocappelletti: agree with Alberto, let's remove the commented code and keep the code simple

paolocappelletti (approved):

PR #1003, opened 2024-03-15
Reward from mainchain (ZenIP 42206) - additional constraint (max cap)

i-skrypnyk:
Add a limit(cap) of funds from mainchain that are available for distribution at the end of withdrawal epoch.
Cap is calculated as 10% of mainchain coinbase reward from the current withdrawal epoch.
Cap is applied during block forging and block application.
Cap application is controlled by a 1.4.0 fork

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
+++ b/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
@@ -161,7 +161,7 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
   override def getFeePaymentsInfo(state: MS, withdrawalEpochNumber: Int) : FPI = {
     val consensusEpochNumber: ConsensusEpochNumber = state.getCurrentConsensusEpochInfo._2.epoch
     val feePayments = state.getFeePaymentsInfo(withdrawalEpochNumber, consensusEpochNumber)
-    AccountFeePaymentsInfo(feePayments)
+    AccountFeePaymentsInfo(feePayments._1)
   }
 
   override def getScanPersistentWallet(modToApply: AccountBlock, stateOp: Option[MS], epochNumber: Int, wallet: VL) : VL = {
--- a/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
+++ b/sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
@@ -160,8 +160,8 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
 
   override def getFeePaymentsInfo(state: MS, withdrawalEpochNumber: Int) : FPI = {
     val consensusEpochNumber: ConsensusEpochNumber = state.getCurrentConsensusEpochInfo._2.epoch
-    val feePayments = state.getFeePaymentsInfo(withdrawalEpochNumber, consensusEpochNumber)
-    AccountFeePaymentsInfo(feePayments._1)
+    val (feePayments, _) = state.getFeePaymentsInfo(withdrawalEpochNumber, consensusEpochNumber)
+    AccountFeePaymentsInfo(feePayments)
   }
 
   override def getScanPersistentWallet(modToApply: AccountBlock, stateOp: Option[MS], epochNumber: Int, wallet: VL) : VL = {
--- a/sdk/src/main/scala/io/horizen/account/state/AccountState.scala
+++ b/sdk/src/main/scala/io/horizen/account/state/AccountState.scala
@@ -210,11 +210,7 @@ class AccountState(
       stateView.updateForgerBlockCounter(mod.forgerPublicKey, consensusEpochNumber)
 
       // If SC block has reached the end of the withdrawal epoch reward the forgers.
-      val distributionCap = if (Version1_4_0Fork.get(consensusEpochNumber).active) {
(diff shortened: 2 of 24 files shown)


sdk/src/main/scala/io/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -57,4 +57,21 @@ object AccountFeePaymentsUtils {
       }
     }
   }
+
+  def getMainchainWithdrawalEpochDistributionCap(epochMaxHeight: Long, epochLength: Long): BigInteger = {
+    val baseReward = 12.5 * 1e8
+    val halvingInterval = 840000
```
alsala, line 63: This `halvingInterval` value would not be valid for regtest (there it is 2000), could it be useful to have a network dependent parameter?

sdk/src/main/scala/io/horizen/account/utils/AccountFeePaymentsUtils.scala
```diff
@@ -57,4 +57,21 @@ object AccountFeePaymentsUtils {
...
+    for (height <- epochMaxHeight - epochLength until epochMaxHeight) {
+      var reward = baseReward.longValue()
+      val halvings = height / halvingInterval
+      for (_ <- 1L to halvings) {
+        reward = reward >> 1
+      }
+      result = result + reward
+    }
+
+    BigInteger.valueOf(result).divide(BigInteger.valueOf(10))
```
alsala, line 75: This percentage value is valid as of now, but in principle it might be modified in future releases. As a consequence we will be required to have this value dependent on a hardfork, therefore it would be better to access this value via an accessor method somehow or at least define a constant.

sdk/src/main/scala/io/horizen/account/state/AccountState.scala
```diff
@@ -210,7 +210,11 @@ class AccountState(
       stateView.updateForgerBlockCounter(mod.forgerPublicKey, consensusEpochNumber)
 
       // If SC block has reached the end of the withdrawal epoch reward the forgers.
-      evalForgersReward(mod, modWithdrawalEpochInfo, consensusEpochNumber, stateView)
+      val distributionCap = if (Version1_4_0Fork.get(consensusEpochNumber).active) {
```
alsala, line 213: This value is in Satoshi, aka Zennies, shouldn't we consider wei value instead?

sdk/src/main/scala/io/horizen/account/state/AccountState.scala
```diff
@@ -210,7 +210,11 @@ class AccountState(
...
+        val mcLastBlockHeight = params.mainchainCreationBlockHeight + ((currentWithdrawalEpochInfo.epoch + 1) * params.withdrawalEpochLength) - 1
+        AccountFeePaymentsUtils.getMainchainWithdrawalEpochDistributionCap(mcLastBlockHeight, params.withdrawalEpochLength)
+      } else BigInteger.valueOf(Long.MaxValue)
+      evalForgersReward(mod, modWithdrawalEpochInfo, consensusEpochNumber, distributionCap, stateView)
```
alsala, line 217: Actually `distributionCap` is required only in the end of withdrawal epoch. Wouldn't be better to calculate it inside the `evalForgersReward` method when this happens?

alsala (changes requested):

sdk/src/main/scala/io/horizen/account/AccountSidechainNodeViewHolder.scala
```diff
@@ -161,7 +161,7 @@ class AccountSidechainNodeViewHolder(sidechainSettings: SidechainSettings,
   override def getFeePaymentsInfo(state: MS, withdrawalEpochNumber: Int) : FPI = {
     val consensusEpochNumber: ConsensusEpochNumber = state.getCurrentConsensusEpochInfo._2.epoch
     val feePayments = state.getFeePaymentsInfo(withdrawalEpochNumber, consensusEpochNumber)
```
alsala, line 163: Cosmetic: maybe `val (feePayments, _) = ...` is better

sdk/src/main/scala/io/horizen/account/state/AccountStateView.scala
```diff
@@ -131,11 +134,11 @@ class AccountStateView(
     metadataStorageView.getForgerBlockCounters
   }
 
-  def resetForgerPoolAndBlockCounters(consensusEpochNumber: ConsensusEpochNumber): Unit = {
+  def subtractForgerPoolBalanceAndResetBlockCounters(consensusEpochNumber: ConsensusEpochNumber, poolBalanceDistributed: BigInteger): Unit = {
     if (Version1_2_0Fork.get(consensusEpochNumber).active) {
       val forgerPoolBalance = getBalance(WellKnownAddresses.FORGER_POOL_RECIPIENT_ADDRESS)
```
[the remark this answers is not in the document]
i-skrypnyk: Yes, makes sense - I added a check for that that throws exception - this will fail fast so to inform about any bad modifications in the future. Thanks!

sdk/src/main/scala/io/horizen/account/state/AccountState.scala
i-skrypnyk: Thanks! Totally missed that

sdk/src/main/scala/io/horizen/account/state/AccountState.scala
i-skrypnyk: Fixed, thanks!

i-skrypnyk:
@alsala thank you for the review - applied/fixed all comments, please take a look!

alsala (changes requested):

alsala (approved):

paolocappelletti (approved):

PR #1004, opened 2024-03-15
St/stake reward

saratnt:
First draft of new forger stake storage model (see https://horizenlabs.atlassian.net/browse/SDK-1749?atlOrigin=eyJpIjoiNGJiYzk1ZmQ5ZmI1NDZmNGJlNjU0YTQ4OTJiNjcxMWIiLCJwIjoiaiJ9)


sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+import sparkz.crypto.hash.Blake2b256
+import sparkz.util.serialization.{Reader, Writer}
+
+import java.math.BigInteger
+import scala.collection.mutable.ListBuffer
+
+
+object ForgerStakeStorageV3 {
+
+  val ACCOUNT = FORGER_STAKE_V2_SMART_CONTRACT_ADDRESS
```
i-skrypnyk, line 22: Why do we call this ACCOUNT and not ADDRESS?

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    if (!forgerData.sameElements(NULL_HEX_STRING_32))
+      Some(ForgerInfoV3Serializer.parseBytes(view.getAccountStorageBytes(ACCOUNT, forgerKey)))
+    else
+      None
+  }
+
+  def apply(view: BaseAccountStateView, forgerKey: Array[Byte]): ForgerInfoV3 = {
+    val forgerData = view.getAccountStorage(ACCOUNT, forgerKey)
+    ForgerInfoV3Serializer.parseBytes(view.getAccountStorageBytes(ACCOUNT, forgerKey))
+  }
```
i-skrypnyk, line 369: forgerData is not used here, I believe we should check the existance as in getForger()

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+
+  def addForger(view: BaseAccountStateView,
+                blockSignProposition: PublicKey25519Proposition,
+                vrfPublicKey: VrfPublicKey,
+                rewardShare: Int,
+                rewardAddress: Address,
+                epochNumber: Int,
+                delegatorPublicKey: Address,
+                stakedAmount: BigInteger): Unit = {
```
i-skrypnyk, line 42: according to the [doc](https://docs.google.com/document/d/1A71qiAENm1SKmclHPczXuTMckM41zClzQiNDRNAFzg0/edit) minimum stake value for creating a forger should be 10 zen - we do not check this condition anywhere in the code, is that ok?

i-skrypnyk: Also at 268 - stakedAmount is checked to be >= 0

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+      None
+  }
+
+  def apply(view: BaseAccountStateView, forgerKey: Array[Byte]): ForgerInfoV3 = {
+    val forgerData = view.getAccountStorage(ACCOUNT, forgerKey)
+    ForgerInfoV3Serializer.parseBytes(view.getAccountStorageBytes(ACCOUNT, forgerKey))
+  }
+
+
+  def getPagedListOfForgers(view: BaseAccountStateView, startPos: Int, pageSize: Int): (Int, Seq[ForgerInfoV3]) = {
```
i-skrypnyk, line 372: Let's add a case class for this response - something like
`case class PagedForgersListResponse(endPosition: Int, forgers: Seq[ForgerInfoV3])
`

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+  case class StakeHistory(uid: Array[Byte])
+    extends StateDbArray(ACCOUNT, Blake2b256.hash(Bytes.concat(uid, "History".getBytes("UTF-8")))) {
+
+    def addCheckpoint(view: BaseAccountStateView, epoch: Int, stakeAmount: BigInteger): Unit = {
+      val checkpoint = StakeCheckpoint(epoch, stakeAmount)
+      append(view, checkpointToPaddedBytes(checkpoint))
+    }
+
+    def updateOrAddCheckpoint(view: BaseAccountStateView, historySize: Int, epoch: Int, op: BigInteger => BigInteger): Unit = {
+      if (historySize > 0) {
```
i-skrypnyk, line 176: In the io.horizen.account.state.ForgerStakeStorageV3#addStake we throw an exception if history size or stake size == 0, so it would never be 0 here

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+      if (newAmount.signum() == -1)
+        throw new ExecutionRevertedException(s"Not enough stake. Claimed $stakedAmount, available $latestStake")
+      newAmount
+    }
+
+    stakeHistory.updateOrAddCheckpoint(view, stakeHistorySize, epochNumber, subtractStake)
+    forgerHistory.updateOrAddCheckpoint(view, forgerHistorySize, epochNumber, subtractStake)
+  }
+
+  def getAllForgerStakes(view: BaseAccountStateView): Seq[(PublicKey25519Proposition, VrfPublicKey, Address, BigInteger)] = {
```
i-skrypnyk, line 145: Let's add case class for this response and it can be used in many other places:
`case class ForgerStake(blockSignProposition: PublicKey25519Proposition, vrfPublicKey: VrfPublicKey, rewardAddress: Address, stakedAmount: BigInteger)
`

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    val listOfForgerKeys = ForgerMap.getKeys(view)
+    listOfForgerKeys.flatMap { forgerKey =>
+      val forger = ForgerMap(view, forgerKey)
+      val delegatorList = DelegatorList(forgerKey)
+      val delegatorSize = delegatorList.getSize(view)
+      val listOfStakes: ListBuffer[(PublicKey25519Proposition, VrfPublicKey, Address, BigInteger)] = ListBuffer()
+      for (idx <- 0 until delegatorSize) {
+        val delegator = delegatorList.getDelegatorAt(view, idx).address()
+        val stakeKey = getStakeKey(forgerKey, delegator)
+        val stakeHistory = StakeHistory(stakeKey)
```
i-skrypnyk, line 155: There's lots of places where we have these 2 lines one after another - why don't we combine it into a single method getStakeHistory that calculates key and returns history right away. Same for forgerHistory

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+
+  }
+
+
+  case class StakeHistory(uid: Array[Byte])
+    extends StateDbArray(ACCOUNT, Blake2b256.hash(Bytes.concat(uid, "History".getBytes("UTF-8")))) {
+
+    def addCheckpoint(view: BaseAccountStateView, epoch: Int, stakeAmount: BigInteger): Unit = {
```
paolocappelletti, line 170: I would add an additional check that the epoch in input is > last epoch we have in the storage.
otherwise throw an exception

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+      if (historySize > 0) {
+        val lastElemIndex = historySize - 1
+        val lastCheckpoint = getCheckpoint(view, lastElemIndex)
+        val newAmount = op(lastCheckpoint.stakedAmount)
+        if (lastCheckpoint.fromEpochNumber == epoch) {
+          val checkpoint = StakeCheckpoint(epoch, newAmount)
+          updateValue(view, lastElemIndex, checkpointToPaddedBytes(checkpoint))
+        }
+        else
+          addCheckpoint(view, epoch, newAmount)
```
paolocappelletti, line 185: also here better to check that lastCheckpoint.fromEpochNumber < epoch
(they must be inserted in order)

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+
+  def getDelegatorKey(delegatorPublicKey: Address): Array[Byte] =
+    Blake2b256.hash(new AddressProposition(delegatorPublicKey).pubKeyBytes())
+
+  def getStakeKey(forgerKey: Array[Byte], delegatorPublicKey: Address): Array[Byte] = {
+    val delegatorKey = getDelegatorKey(delegatorPublicKey)
+    Blake2b256.hash(Bytes.concat(forgerKey, delegatorKey))
+  }
+
+  def addForger(view: BaseAccountStateView,
```
paolocappelletti, line 35: please doulbe check (maybe with a unit test) in all the methods that the behaviour is not influenced by the case in the addresses/keys
(if I change the case of some characters in the address for example I should not end up with 2 different records)

paolocappelletti (changes requested):

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: No particular reason, actually :D 
This is in fact an address, as a type, but the idea was more that this objects lives inside the environment of a specific account and uses its storage. In the end, it's the same, so, if you think it is clearer calling it ADDRESS, I'll change it.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: You're right, I eliminated forgerData. 
As for checking for the existence, this method was created specifically for avoiding the overhead of comparing the result of view.getAccountStorageBytes(ACCOUNT, forgerKey) with NULL_HEX_STRING_32, as in getForger, because it is used in places where I'm expecting the forger exists, as when I'm iterating through the list of all forgers, as in `getAllForgerStakes`. At least for now, if there is a key inside the forger list, it must exist the corresponding forgerData in the storage.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: As Paolo said, this class is just the interface to the data in the statedb, so I tried to avoid putting business logic in there as much as possible. I would add this check (or other similar checks) here only if it helped reducing the gas cost of a transaction calling this method, but for now it doesn't seem the case.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+                epochNumber: Int,
+                delegatorPublicKey: Address,
+                stakedAmount: BigInteger): Unit = {
+
+    val forgerKey = getForgerKey(blockSignProposition, vrfPublicKey)
+
+    if (ForgerMap.existsForger(view, forgerKey))
+      throw new ExecutionRevertedException(s"Forger already registered.")
+
+    ForgerMap.addForger(view, forgerKey, blockSignProposition, vrfPublicKey, rewardShare, rewardAddress)
```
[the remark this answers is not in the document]
saratnt: In this case, I think this check is better to be executed inside the smart contract. It is a check on the input parameters and it can be performed before any additional, maybe more expensive, operations.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: Done

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: In `io.horizen.account.state.ForgerStakeStorageV3#addStake`  the stake size can be 0, in case it is the first time a specific user delegates some stake to this forger. 
But this `updateOrAddCheckpoint` method probably tries to do too many things at once, resulting in code that isn't very clear... I'll try to see if I can refactor it a bit.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: Actually, a class like this exists already (ForgerStakeData). I'll use it, then.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: That's good point. I'll do it for stakeHistory. For forgerHistory, I still prefer like it is, because the forgerKey is used not only for getting the forgerHistory, but also for other operations in the same scope and I prefer not to calculate it more than once.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
+package io.horizen.account.state
```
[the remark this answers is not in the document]
saratnt: Done

sdk/src/test/scala/io/horizen/account/state/ForgerStakeStorageV3Test.scala
```diff
@@ -0,0 +1,735 @@
+package io.horizen.account.state
```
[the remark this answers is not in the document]
saratnt: Done

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+          listOfStakes.append((forger.forgerPublicKeys.blockSignPublicKey, forger.forgerPublicKeys.vrfPublicKey, delegator, amount))
+      }
+      listOfStakes
+    }
+
+
+  }
+
+
+  case class StakeHistory(uid: Array[Byte])
```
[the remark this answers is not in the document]
saratnt: The fact is that the key used for the forger is used not only for getting the forger StakeHistory, but also for other operations inside the same scope. I would prefer not calculate it more than once.
For the stake history of delegator/forger couple, I have created a "factory" method that creates directly its key, as suggested by Ivan.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+      }
+    }
+
+
+    private[horizen] def checkpointToPaddedBytes(checkpoint: StakeCheckpoint): Array[Byte] = {
+      BytesUtils.padRightWithZeroBytes(StakeCheckpointSerializer.toBytes(checkpoint), 32)
+    }
+  }
+
+  case class DelegatorList(fuid: Array[Byte])
```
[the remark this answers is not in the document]
saratnt: Same as previous comment, I would prefer not calculate several times the same forgerkey.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    val stakeAmountLength = r.getInt()
+    val stakeAmount = new BigIntegerUInt256(r.getBytes(stakeAmountLength)).getBigInt
+    StakeCheckpoint(fromEpochNumber, stakeAmount)
+  }
+}
+
+object ForgerMap {
+  val forgerList: StateDbArray = new StateDbArray(ACCOUNT, "Forgers".getBytes("UTF-8"))
+
+  def existsForger(view: BaseAccountStateView, forgerKey: Array[Byte]): Boolean = {
```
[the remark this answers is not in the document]
saratnt: Same as previous comment, I would prefer not calculate several times the same forgerkey.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    val forgerData = view.getAccountStorage(ACCOUNT, forgerKey)
+    // getting a not existing key from state DB using RAW strategy
+    // gives an array of 32 bytes filled with 0, while using CHUNK strategy
+    // gives an empty array instead
+    !forgerData.sameElements(NULL_HEX_STRING_32)
+  }
+
+  def addForger(view: BaseAccountStateView,
+                forgerKey: Array[Byte],
```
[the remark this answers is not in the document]
saratnt: Same as previous comment, I would prefer not calculate several times the same forgerkey.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    val forgerStakeData = ForgerInfoV3(
+      ForgerPublicKeys(blockSignProposition, vrfPublicKey), rewardShare, new AddressProposition(rewardAddress))
+
+    // store the forger stake data
+    view.updateAccountStorageBytes(ACCOUNT, forgerKey,
+      ForgerInfoV3Serializer.toBytes(forgerStakeData))
+  }
+
+  def updateForger(view: BaseAccountStateView,
+                forgerKey: Array[Byte],
```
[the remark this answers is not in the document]
saratnt: Same as previous comment, I would prefer not calculate several times the same forgerkey.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+  def getSize(view: BaseAccountStateView): Int = forgerList.getSize(view)
+
+  def getKeys(view: BaseAccountStateView): Seq[Array[Byte]] = {
+    val listSize = getSize(view)
+    (0 until listSize).map(index => {
+     forgerList.getValue(view, index)
+    })
+  }
+
+  def getForger(view: BaseAccountStateView, forgerKey: Array[Byte]): Option[ForgerInfoV3] = {
```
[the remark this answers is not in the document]
saratnt: Same as before.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
```diff
@@ -0,0 +1,404 @@
...
+    val forgerData = view.getAccountStorage(ACCOUNT, forgerKey)
+    if (!forgerData.sameElements(NULL_HEX_STRING_32))
+      Some(ForgerInfoV3Serializer.parseBytes(view.getAccountStorageBytes(ACCOUNT, forgerKey)))
+    else
+      None
+  }
+
+  def apply(view: BaseAccountStateView, forgerKey: Array[Byte]): ForgerInfoV3 = {
```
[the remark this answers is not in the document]
saratnt: Same as before.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: I have added the check in updateOrAddCheckpoint, because it is the only place where a checkpoint can be added to an existing stake history.

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: Done

sdk/src/main/scala/io/horizen/account/state/ForgerStakeStorageV3.scala
saratnt: This is a very good point and a real issue, thanks for finding it! I have changed the code so the checksum address is always used for the delegator.

paolocappelletti (approved):

i-skrypnyk (approved):

PR #1013, opened 2024-04-09
Metrics endpoint

paolocappelletti:
## Description
First implementation of metrics support.
Added a new /metrics endpoint exposing some metrics in Prometheus format.
The idea is that the infrastructure we have already infra-side to display metrics (Prometheus + Graphana) will ping the endpoint periodically, collect the data and display it on a dashboard.
Added also an endpoint /metrics/help that explains the meaning of data exposed.

Note: docker part could be subject to following changes - waiting for coordination with infra team

## Jira Ticket
[SDK-1549](https://horizenlabs.atlassian.net/browse/SDK-1549)

Output example (tested on pregobi node 1):
image

[SDK-1549]: https://horizenlabs.atlassian.net/browse/SDK-1549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


saratnt (approved):
LGTM, just some minor comments.

sdk/src/main/java/io/horizen/metrics/MetricsManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private MetricsManager(TimeProvider timeProvider) throws IOException {
+        logger.debug("Initializing metrics engine");
+
+        this.timeProvider = timeProvider;
+
+        //JvmMetrics.builder().register(); // initialize the out-of-the-box JVM metrics
+        helps = new ArrayList<>();
+
+        blockApplyTime  =  Gauge.builder().name("block_apply_time").register();
```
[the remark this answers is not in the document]
paolocappelletti: will be evntually calculated on Graphana - here we have to just provide the latest values

sdk/src/main/java/io/horizen/metrics/MetricsManager.java
```diff
@@ -0,0 +1,100 @@
...
+        helps.add(new MetricsHelp(blocksNotApplied.getPrometheusName(), "Number of received blocks not applied (absolute value since start of the node)"));
+
+        mempoolSize =  Gauge.builder().name("mempool_size").register();
+        helps.add(new MetricsHelp(mempoolSize.getPrometheusName(), "Mempool size (number of transactions in this node mempool)"));
+
+        forgeLotteryTime = Gauge.builder().name("forge_lottery_time").register();
+        helps.add(new MetricsHelp(forgeLotteryTime.getPrometheusName(), "Time to execute the lottery (milliseconds)"));
+
+        forgeBlockCreationTime = Gauge.builder().name("forge_blockcreation_time").register();
+        helps.add(new MetricsHelp(forgeBlockCreationTime.getPrometheusName(),  "Time to create a new forged block (calculated from the start timestamp of the slot it belongs to) (milliseconds)"));
```
i-skrypnyk, line 69: Maybe not worth is for a POC, but for next iteration instead of creating MetricHelps, we can do this:

`Gauge.builder().name("forge_blockcreation_time").help("Time to create a new forged block (calculated from the start timestamp of the slot it belongs to) (milliseconds)").register();`

and then something like 

`PrometheusRegistry.defaultRegistry.scrape.iterator().toList.map(ms => new MetricsHelp(ms.getMetadata.getPrometheusName, ms.getMetadata.getHelp))`

This way we do not need to have a Manager class and keep all metrics in a single place, so each actor\class can create and register its own set of metrics

i-skrypnyk (approved):

sdk/src/main/java/io/horizen/metrics/MetricsManager.java
paolocappelletti: I have discarded this option because the default PrometheusTextFormatWriter I use in the endpoint always displays the help if it is set directly in the metric (maybe we can in the future write a custom writer to avoit this - for now postponed)

PR #1015, opened 2024-04-12
St/stake reward get forgers

saratnt:
## Jira Ticket
[SDK-1782](https://horizenlabs.atlassian.net/browse/SDK-1782?atlOrigin=eyJpIjoiNGMxMDM0ZjdkMWI1NGZkYzg5YzAyZWY0ZDA3NGExMGQiLCJwIjoiaiJ9)

[SDK-1782]: https://horizenlabs.atlassian.net/browse/SDK-1782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/java/io/horizen/account/state/ForgerInfoABI.java
+++ b/sdk/src/main/java/io/horizen/account/state/ForgerInfoABI.java
@@ -1,0 +1,42 @@
+package io.horizen.account.state;
+
+import org.web3j.abi.datatypes.StaticStruct;
+import org.web3j.abi.datatypes.generated.Bytes1;
+import org.web3j.abi.datatypes.generated.Bytes32;
+import org.web3j.abi.datatypes.generated.Uint256;
+import org.web3j.abi.datatypes.generated.Uint32;
+
+import java.math.BigInteger;
+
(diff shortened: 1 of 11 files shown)


sdk/src/main/java/io/horizen/account/state/ForgerInfoABI.java
```diff
@@ -0,0 +1,42 @@
+package io.horizen.account.state;
```
paolocappelletti, line 1: just a note: maybe is better to move the abi descriptors here
/src/main/scala/io/horizen/account/state/nativescdata/forgerstakev2/

paolocappelletti (approved):

i-skrypnyk (approved):

sdk/src/main/java/io/horizen/account/state/ForgerInfoABI.java
saratnt: It must be a java class unfortunately, but I have changed its package so it is under nativescdata.forgerstakev2.

PR #1028, opened 2024-05-09
stakeStart() native

i-skrypnyk:
## Description
stakeStart native method

## Jira Ticket
[SDK-1804](https://horizenlabs.atlassian.net/browse/SDK-1804)

[SDK-1804]: https://horizenlabs.atlassian.net/browse/SDK-1804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/qa/sc_evm_native_forger_v2.py
+++ b/qa/sc_evm_native_forger_v2.py
@@ -281,6 +281,19 @@ class SCEvmNativeForgerV2(AccountChainSetup):
             print("Expected exception thrown: {}".format(err))
             assert_true("Forger stake V2 has not been activated yet" in str(err))
 
+        # Check that stakeStart cannot be called before activate
+        method = 'stakeStart(bytes32,bytes32,bytes1,address)'
+        forger_1_vrf_pub_key_to_bytes = hex_str_to_bytes(vrf_pub_key_1)
+
+        try:
+            contract_function_static_call(sc_node_1, forger_v2_native_contract, FORGER_STAKE_V2_SMART_CONTRACT_ADDRESS,
+                                          evm_address_sc_node_1, method, forger_1_sign_key_to_bytes,
+                                          forger_1_vrf_pub_key_to_bytes[0:32], forger_1_vrf_pub_key_to_bytes[32:],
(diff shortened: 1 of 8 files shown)


qa/sc_evm_native_forger_v2.py
```diff
@@ -595,6 +608,22 @@ def run_test(self):
...
+                                                               forger_1_vrf_pub_key_to_bytes[32:],
+                                                               "0x" + evm_address_sc_node_1)
+        result = sc_node_1.rpc_eth_call(
+            {
+                "to": "0x" + FORGER_STAKE_V2_SMART_CONTRACT_ADDRESS,
+                "from": add_0x_prefix(evm_address_sc_node_1),
+                "input": data_input
+            }, "latest"
+        )
+        assert_equal(decode(['uint32'], hex_str_to_bytes(result['result'][2:]))[0], VERSION_1_4_FORK_EPOCH)
```
alsala, line 625: It works here, but actually it should be decoded as an an `int32`. Moreover maybe we could add a test using an address which did not delegate anything and checking that it returns `-1`.

qa/sc_evm_native_forger_v2.py
```diff
@@ -595,6 +608,22 @@ def run_test(self):
...
+
```
alsala, line 626: Another interesting test would be to request the epoch for genesis staking address and see that it returns 1

saratnt (approved):

qa/sc_evm_native_forger_v2.py
i-skrypnyk: Thanks, fixed and added a check

qa/sc_evm_native_forger_v2.py
i-skrypnyk: I will cover this under https://horizenlabs.atlassian.net/browse/EON-1826 as I am adding more fo.rgers with different config there

PR #1037, opened 2024-05-28
Fix for issue EON-1857

alsala:
Fix for issue [EON-1857](https://horizenlabs.atlassian.net/browse/EON-1857)

The fix consists in preventing the `updateForger` operation for the 2 epochs after the 1.4 fork activation, assuming that the V2 activation happens in the same epoch.

This fix has a companion fix in EON repository: https://github.com/HorizenOfficial/eon/pull/110

[EON-1857]: https://horizenlabs.atlassian.net/browse/EON-1857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdk/src/main/scala/io/horizen/account/api/http/route/AccountTransactionApiRoute.scala
+++ b/sdk/src/main/scala/io/horizen/account/api/http/route/AccountTransactionApiRoute.scala
@@ -16,7 +16,7 @@ import io.horizen.account.node.{AccountNodeView, NodeAccountHistory, NodeAccount
 import io.horizen.account.proof.SignatureSecp256k1
 import io.horizen.account.proposition.AddressProposition
 import io.horizen.account.secret.PrivateKeySecp256k1
-import io.horizen.account.state.ForgerStakeV2MsgProcessor.{MAX_REWARD_SHARE, MIN_REGISTER_FORGER_STAKED_AMOUNT_IN_WEI}
+import io.horizen.account.state.ForgerStakeV2MsgProcessor.{MAX_REWARD_SHARE, MIN_REGISTER_FORGER_STAKED_AMOUNT_IN_WEI, NUM_OF_EPOCHS_AFTER_FORK_ACTIVATION_FOR_UPDATE_FORGER}
 import io.horizen.account.state.McAddrOwnershipMsgProcessor._
 import io.horizen.account.state._
 import io.horizen.account.state.nativescdata.forgerstakev2.RegisterOrUpdateForgerCmdInputDecoder.NULL_ADDRESS_WITH_PREFIX_HEX_STRING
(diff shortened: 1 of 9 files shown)


sdk/src/main/scala/io/horizen/account/api/http/route/AccountTransactionApiRoute.scala
```diff
@@ -666,66 +666,72 @@ case class AccountTransactionApiRoute(override val settings: RESTApiSettings,
...
+    if (!fork14.active) {
       ApiResponseUtil.toResponse(GenericTransactionError(s"Fork 1.4 is not active, can not invoke this command",
         JOptional.empty()))
     }
     else if (!accountState.forgerStakesV2IsActive) {
       ApiResponseUtil.toResponse(GenericTransactionError(s"Forger Stake Storage V2 is not active, can not invoke this command",
         JOptional.empty()))
+    }
+    else if (operation == ForgerStakeV2MsgProcessor.UpdateForgerCmd &&
+      epochNumber < (fork14.mainnetActivationEpoch + NUM_OF_EPOCHS_AFTER_FORK_ACTIVATION_FOR_UPDATE_FORGER) ) {
```
saratnt, line 679: Shouldn't be selected the activation epoch for the correct network, i.e for mainnet mainnetActivationEpoch, for regstest regtestActivationEpoch and for testnet testnetActivationEpoch? They are not necessarily the same

sdk/src/main/scala/io/horizen/account/state/ForgerStakeV2MsgProcessor.scala
```diff
@@ -185,10 +186,18 @@ object ForgerStakeV2MsgProcessor extends NativeSmartContractWithFork  with Forge
...
   }
 
-  def doUpdateForger(invocation: Invocation, gasView: BaseAccountStateView): Array[Byte] = {
+  def doUpdateForger(invocation: Invocation, gasView: BaseAccountStateView, context: ExecutionContext): Array[Byte] = {
     requireIsNotPayable(invocation)
     checkForgerStakesV2IsActive(gasView)
 
+    val currentEpoch = context.blockContext.consensusEpochNumber
+    val activationEpoch = Version1_4_0Fork.get(currentEpoch).mainnetActivationEpoch
+    if (currentEpoch < (activationEpoch + NUM_OF_EPOCHS_AFTER_FORK_ACTIVATION_FOR_UPDATE_FORGER) ) {
```
saratnt, line 195: Won't this cause a sync failure in Pregobi? I think we did an updateForger on Pregobi before 2 epochs passed after activation. In addition, shouldn't be used a different activationEpoch depending on the type of network (regtest, testne, mainnet)?

saratnt (changes requested):

sdk/src/main/scala/io/horizen/account/fork/Version1_4_0Fork.scala
```diff
@@ -2,7 +2,11 @@ package io.horizen.account.fork
 
 import io.horizen.fork.{ForkManager, OptionalSidechainFork}
 
-case class Version1_4_0Fork(active: Boolean = false) extends OptionalSidechainFork
+case class Version1_4_0Fork(active: Boolean = false,
+                            regtestActivationEpoch: Int = -1,
```
paolocappelletti, line 6: i don't like re-adding the activation epochs, they should already be stored in the forkmanager. we just need a way to get them.
see for example the method ForkManager.getOptionalSidechainFork (the one used now by the each fork to check is activation): it iterates through a sequence optionalSidechainForks which contains all the forks and their activation heithgs (already configured for the current network)
We may introduce a new method
Forkmanager.getFirstConfiguredActivationHeight([T <: OptionalSidechainFork : Manifest]) which uses the same collection to return the configured consensus epoch of activation

sdk/src/main/scala/io/horizen/account/state/ForgerStakeV2MsgProcessor.scala
paolocappelletti: see my commet in Version1_4_0Fork.scala : with that change, we should have already the correct value for the current running network

paolocappelletti (changes requested):

sdk/src/main/scala/io/horizen/account/api/http/route/AccountTransactionApiRoute.scala
alsala: As discussed, the approach suggested by Paolo allow us to overcome also this issue.

sdk/src/main/scala/io/horizen/account/fork/Version1_4_0Fork.scala
alsala: Ok, I followed the suggested approach

saratnt (approved):

paolocappelletti (approved):

paolocappelletti (approved):

PR #1040, opened 2024-06-12
zen_getFeePayments should display separately the rewards from the mainchain

i-skrypnyk:
## Jira Ticket
https://horizenlabs.atlassian.net/browse/EON-1883

## Checks
- [x] Project Builds

### DIFF
--- a/sdk/src/main/java/io/horizen/account/api/rpc/types/FeePaymentsView.java
+++ b/sdk/src/main/java/io/horizen/account/api/rpc/types/FeePaymentsView.java
@@ -1,5 +1,6 @@
 package io.horizen.account.api.rpc.types;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import io.horizen.account.chain.AccountFeePaymentsInfo;
 import io.horizen.account.utils.AccountPayment;
 import io.horizen.evm.Address;
@@ -7,6 +8,7 @@ import scala.collection.JavaConverters;
 
 import java.math.BigInteger;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class FeePaymentsView {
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/io/horizen/account/api/rpc/types/FeePaymentsView.java
```diff
@@ -15,17 +16,30 @@ public FeePaymentsView(AccountFeePaymentsInfo info) {
...
             this.value = value;
+            this.valueFromMainchain = valueFromMainchain;
+            this.valueFromFees = valueFromFees;
+        }
+
+        public static FeePaymentData fromAccountFeePayment(AccountPayment payment) {
+            return new FeePaymentData(
+                payment.address().address(),
+                payment.value(),
+                payment.valueFromMainchain().getOrElse(() -> BigInteger.ZERO),
```
saratnt, line 40: Is it possible to filter out in the json response `valueFromMainchain` and  `valueFromFees` when they are None? Because putting 0 seems misleading in this case. Or maybe returning -1 instead of 0, just to make more explicit that these 2 fields are not meaningful before the fork.

saratnt (changes requested):

alsala (approved):
I agree with Sara, for the rest lgtm

sdk/src/main/java/io/horizen/account/api/rpc/types/FeePaymentsView.java
i-skrypnyk: Found the Serializer, will try to hide the field completely. If not possible, will change to -1 for missing value

i-skrypnyk: Added a custom serizlized module for java optionals, now empty will be filtered. Thanks!

saratnt (approved):

PR #1042, opened 2024-06-17
St/preimages

saratnt:
## Description
Added preimages configuration for enabling statedb dump.

## Jira Ticket
[EON-1880](https://horizenlabs.atlassian.net/browse/EON-1880)

## Additional information
Requires https://github.com/HorizenOfficial/libevm/pull/32

[EON-1880]: https://horizenlabs.atlassian.net/browse/EON-1880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/qa/resources/template.conf
+++ b/qa/resources/template.conf
@@ -113,6 +113,5 @@ sparkz {
 
   evmStateDump {
       enabled = %(EVM_STATE_DUMP_ENABLED)s
-
   }
 }
--- a/qa/resources/template.conf
+++ b/qa/resources/template.conf
@@ -110,4 +110,9 @@ sparkz {
       txLifetime = %(TX_LIFETIME)d seconds
       allowUnprotectedTxs = %(ALLOW_UNPROTECTED_TXS)s
   }
+
+  evmStateDump {
+      enabled = %(EVM_STATE_DUMP_ENABLED)s
+
+  }
 }
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -351,7 +351,7 @@
     <dependency>
       <groupId>io.horizen</groupId>
       <artifactId>libevm</artifactId>
-      <version>1.1.0</version>
+      <version>1.1.0-snapshot1</version>
     </dependency>
     <dependency>
       <groupId>at.favre.lib</groupId>
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -351,7 +351,7 @@
     <dependency>
       <groupId>io.horizen</groupId>
       <artifactId>libevm</artifactId>
-      <version>1.1.0-snapshot1</version>
+      <version>1.2.0-snapshot</version>
     </dependency>
     <dependency>
       <groupId>at.favre.lib</groupId>
--- a/sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
+++ b/sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
@@ -282,7 +282,8 @@ abstract class AbstractSidechainNodeViewHolder[
     if (!history().contains(pmod.id)) {
       context.system.eventStream.publish(StartingPersistentModifierApplication(pmod))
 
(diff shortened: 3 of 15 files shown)


sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
```diff
@@ -282,7 +282,8 @@ abstract class AbstractSidechainNodeViewHolder[
     if (!history().contains(pmod.id)) {
       context.system.eventStream.publish(StartingPersistentModifierApplication(pmod))
 
-      log.info(s"Apply modifier ${pmod.encodedId} of type ${pmod.modifierTypeId} to nodeViewHolder")
+      log.info(s"Apply modifier ${pmod.encodedId} of type ${pmod.modifierTypeId} to nodeViewHolder, " +
+        s"state dump enabled: ${sidechainSettings.evmStateDump.enabled}")
```
i-skrypnyk, line 286: why log the setting value here?

saratnt: It is a way to check that `preimages` were enabled for each block. For the dump to succeed it is necessary that the node has the `preimages` enabled from the genesis block, without skipping any block. In the original Jira task, it was required that the `nodeInfo` endpoint returned first and last block with `preimages` enabled, in order to check that all blocks had `preimages`, but it was too much work for a feature that will be used just once. With this tracing, that it is a pork-around actually, a `grep` is enough to check that the node synch'ed correctly for the dump.

i-skrypnyk (approved):

alsala (approved):

sdk/pom.xml
```diff
@@ -351,7 +351,7 @@
     <dependency>
       <groupId>io.horizen</groupId>
       <artifactId>libevm</artifactId>
-      <version>1.1.0</version>
+      <version>1.1.0-snapshot1</version>
```
paolocappelletti: 1.2.0-snapshot

paolocappelletti: Also, please upgrade the sdk version in all the pom.xml files of this repo to the next version: 0.13.0-snapshot
And rebase the PR over the latest dev (I see still sparkz version 2.4.0-RC1 here but must be 2.4.0)

qa/resources/template.conf
```diff
@@ -110,4 +110,9 @@ sparkz {
       txLifetime = %(TX_LIFETIME)d seconds
       allowUnprotectedTxs = %(ALLOW_UNPROTECTED_TXS)s
   }
+
+  evmStateDump {
+      enabled = %(EVM_STATE_DUMP_ENABLED)s
+
```
paolocappelletti: remove empty line

sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
paolocappelletti: let's add the string ", state dump enabled:" only if enabled = true
(so that by default we don't add anything more in the logs compared to now)

paolocappelletti (changes requested):
Please add also a new release here:
https://github.com/HorizenOfficial/Sidechains-SDK/tree/master/doc/release
with the description of the new flag and how it is used for

sdk/src/main/scala/io/horizen/AbstractSidechainNodeViewHolder.scala
saratnt: Done

sdk/pom.xml
saratnt: Done except for updating sdk version to 0.13.0-snapshot, because it was done in `st/dump` branch.

qa/resources/template.conf
saratnt: Done

paolocappelletti (approved):

